PR #23, opened 2017-10-11
Development

schristley:
Merge all the recent master changes into development.


schristley:
@BusseChristian Sorry Christian, I should have explained this better. This is just what I call an "auto-merge". There are a bunch of changes on the `master` branch which do not exist on the `development` branch, this situation really shouldn't occur and needs to be fixed, as it will cause conflicts down the road when trying to merge `development` into `master`. How this happened is that changes were committed onto `master` after the `development` branch was created. Generally this should be avoided except in special circumstances.

I left it as pull request for you to review because if you are working on `development`, you will see a bunch of changes come in, and I didn't want to surprise you. So you should accept this request "as-is", this will bring `master` and `development` into sync, then any "clean up" work (e.g. get rid of files, etc.) is done on `development`, and it will avoid conflicts down the road.

BusseChristian:
@schristley Sorry, if I got confused, but on the top it says "... wants to merge 12 commits into `master` from `development`", with is the opposite direction than the one you were describing. What is it that I'm not getting?

schristley:
@BusseChristian Good catch! yes, you are right, the merge is the wrong way. I created a new pull request which should go the right way.

Issue #5, opened 2017-09-27
Harmonize data types with airr-formats


laserson:
I'd propose harmonizing the data types between the two repos.  In the airr-formats group, we specify the types as `string`, `boolean`, `float`, and `integer`.  Any objections to a PR that would make them the same?

BusseChristian:
No, please go ahead.

schristley:
In definitions.yaml I'm using the canonical JSON types as suggested by Uri, plus `enum`.

BusseChristian:
@schristley : Will `enum` allow for `NULL` values?

schristley:
Not exactly, so for NucleicAcidProcessing if the service tried to return

```
{ "library_source": null}
```

then it would throw a schema validation error (I tested this with code). There are two ways to handle this:

1. If the value is null then don't return that field in the response.
2. Define a "null" value with the enum.

Generally services do 1 as clients need to generally check if the field exists first before accessing its value. The issue is if the field is a required field, then you would need to do 2.

Note this issue also applies to boolean, as null isn't a valid value, and maybe some other data types.

BusseChristian:
Well, for items like `library_source` this information will be available and unambiguous. But we should think about medical items like `sex` where a controlled vocabulary would make sense, but the information might not be available and ambiguous.

schristley:
I agree and providing additional enum values like "N/A", "Unknown", "Not specified", and so on can document those cases.

laserson:
Where will the definitions.yaml file most likely be consumed?

In general, I'd say that all fields should be nullable.

schristley:
Right now it is being referenced by the [iReceptor API work](https://github.com/sfu-ireceptor/api/tree/AIRR-compliance) and the common repository WG [example API](https://github.com/airr-community/common-repo-wg/tree/issue-13/specs). I also expect that the formats WG will reference these definitions for tools to access study metadata.

schristley:
@BusseChristian 
Jason makes a good point about GenBank submission, maybe [vdj]_allele needs to be split out into two fields?

https://github.com/airr-community/airr-formats/issues/45

BusseChristian:
See my comment there. I think the best solution for now is to rename the fields to `[vdjc]_call`, although parsing tends to be difficult. If the subcomponents are required by downstream processes, we should discuss the most likely scenarios to see what to include where.

schristley:
[...]

PR #33, opened 2017-10-23
Switch to [vdjc]_call

laserson:
Fixes issue with `_call` in #5.


schristley:
@laserson Can this be redirected to `development` branch? New dev flow is locking master merges except for releases. I changed the base branch to `development`, but it's now showing a ton of commits besides yours?

laserson:
No problem, but would this dev flow be up for discussion?  It appears the dev branch is already very far behind the master branch somehow, and this is one of the disadvantages of that approach.  I'd personally advocate for all development on master, and putting tags on release commits.  If things need to be patched in earlier releases, you branch off of that commit.  (I think this is how most Apache projects are run.)

schristley:
Yeah, things were getting confusing as some people where committing to master and some work is being done on development. We merged master into development so that should bring development forward. This will get all cleaned up with the release this week and we can all be on the same dev flow going forward.

laserson:
Ok, this is off `development` now.

PR #50, opened 2017-11-05
Changed data types in table to OpenAPI types

### LINKED ISSUE #5: Harmonize data types with airr-formats

laserson:
Fixes #5


laserson:
Here are the commands I ran to make these changes:

```python
import numpy as np
import pandas as pd
df = pd.read_csv('AIRR_Minimal_Standard_Data_Elements.tsv', sep='\t', header=0, index_col=None)
df.iloc[:, 2] = (df.iloc[:, 2]
                 .replace(to_replace='String', value='string')
                 .replace(to_replace='String ', value='string')
                 .replace(to_replace='Integer', value='integer')
                 .replace(to_replace='Boolean', value='boolean')
                 .replace(to_replace='Blob', value=np.nan))
df.to_csv('AIRR_Minimal_Standard_Data_Elements.tsv', sep='\t', index=False)
```

Issue #85, opened 2018-02-14
Finalizing Rearrangement fields


javh:
I just finished making all the (relevant) Change-O tools work with the AIRR format.
Here's the set of spec changes I think are necessary after going through that as well as input 
from a few other discussions. Some of those discussions were offline, but some of it's covered in
the issues: #84, #83, #82, #75, #58.

Before making a pull request with changes, I want to hear others' thoughts as some of these things
require naming decisions which I'm undecided on.  For all fields listed, I've proposed a non-exhaustive
list of alternative names. I tend to prefer short names, but there's a good argument for longer,
more explicit names as well.

## Fields to modify

**rearrangement_id (rearrangment)**

+ Remove from `required`.
+ Add the tag `x-crwg: true` to denote the field as required in the CRWG context.
+ Update the definition to clarify that the field serves as a unique identifier for the 
  Rearrangement object and is not necessarily derived from the input sequence or 
  alignment tool result.
	
**rearrangement_set_id (rearrangement_set)**

+ Remove from `required`.
+ Add the tag `x-crwg: true` to denote the field as required in the CRWG context.
+ Update the definition to clarify that the field serves as a unique identifier for an
  arbitrary collection of Rearrangement object and is not necessarily a meaningful
  cross-study, biological classification.
	
**cell_index**

+ Remove from `required`.
+ Change `type` to `string`.

**sequence (query_sequence, sequence_input)**

+ Clarify in the definition that this field contains the unmodified sequence that was input 
  into the tool. Alternatively, this may contain an intermediate "output" sequence in the sense of
  a consensus sequence, unique sequence, assembled mate-pair sequence, or other results of pre-V(D)J 
  alignment processing embedded in the alignment tool. Basically, whatever the `alignment` field (below) 
  was based on that meets the NCBI GenBank submission criteria: "an uncorrected sequence, which 
  may have gone through assembly or consensus, but not indel correction or truncation".

**functional**

+ Rename to `productive`.

## Necessary fields to add

The following fields are missing, but necessary:

**sequence_id (query_id)**

+ String. The identifier for the `sequence`. 
+ This replaces `rearrangement_id` as the row key for the Rearragenment object, 
  within the context of an analysis file (the default context). The default expectation is that this would be 
  equal to the input sequence header, but that is not strictly required, especially in cases where 
  `sequence` corresponds to such an above described intermediate sequence. 
+ The TL;DR of why this is needed is because `rearrangment_id` and `sequence_id` have conflicting 
[...]

PR #89, opened 2018-02-27
First pass at finalizing the Rearrangement fields.

javh:
1. Commented out metadata file generation in the formats python library.
2. Updated Rearrangement fields per discussion in issue #85. There were some conflicts in the opinions, so I just made arbitrary decisions in those cases where preferences weren't universal. I think I got everything? Except `mutated_invariant` - I think it's best to skip that for now, because it seems to be contentious and I don't think I've ever actually used that field for anything anyway.
3. I made some small improvements to the field descriptions for clarity, but we need to take a pass over the text for all of them more thoroughly once finalized.
4. Changed the MiAIRR `JUNCTION` mapping in the consistency check TSV from `junction_nt` to `junction`.
5. Updated the toy data in the python library with new column names.


schristley:
Looks good!

1. This technically changes the MiAIRR spec with TSV data element name change (`junction_nt` -> `junction`). I think we are okay with this, but is the `junction_nt` name being used anywhere? For example, has any data been submitted to GenBank with `junction_nt`?

2. How are we defining the coordinate fields in relationship to the sequence/alignment fields and `rev_comp`? Should tools automatically assume that a coordinate can be used directly on the sequence, or do they need to check `rev_comp` and reverse complement before applying the coordinates?

javh:
1. The official name for submission to GenBank/TLS is `JUNCTION`. `junction_nt` was specific to the Rearrangement format and hasn't been used elsewhere, to my knowledge. (That's part of why I changed it.)
2. Good question. I'm inclined to go with IgBLAST's approach from @jianye00: "all output data (i.e., alignment coordinates, nucleotide sequence, translated protein sequence, etc) are based on the reversed/complemented sequence of the original read." Ie, check `rev_comp` and reverse complement before applying the coordinates.

javh:
Planning to merge this Thursday, if there are no further comments.

BusseChristian:
Fine with me.

PR #103, opened 2018-04-04
IgBLAST and R package updates

javh:
1. Added individual segment alignment fields for IgBLAST support and other local segment alignment tools.
2. Added R package to `definitions.yaml` consistency checks.

The list of official fields is getting pretty long. Which, I'm fine with, but I think we need an organization strategy in the docs so people aren't overloaded. That's not in this PR though.


schristley:
Are these alignment fields just the sequence or is it the actual alignment string with gaps/indels? If its with gaps, how do you do the translation to aa?

javh:
Actual alignment string with gaps/indels.  Incomplete/ambiguous codons translate to `X`, or `---` to `-`.

schristley:
Okay, so if I understand correctly then the aa translation shouldn't be construed as the actual aa's in the receptor. I suppose if you wanted the actual aa's, you would take e.g. `j_sequence_start` and `j_sequence_end`, and pull the bits form `sequence` and translate yourself? Or maybe translate the start/end positions and pull from `sequence_aa`?

javh:
Well, not the actual amino acids in the query sequence. Whether or not missing information due to indels is truth or sequencing error... debatable.  IMGT-gap are in-frame, so you're not really messing anything up there. 

Yeah, you'd have to extract from `sequence` or `sequence_aa` to get what's in the query sequence if indels or IMGT-gaps are present.

PR #106, opened 2018-04-06
Rearrangement definition tweaks.

javh:
1. Removed `c_call`, `c_score` and `c_cigar` from `required` as this is not typical reference aligner output.
2. Renamed `vdj_score`, `vdj_identity`, `vdj_evalue`, and `vdj_cigar` to `score`, `identity`, `evalue`, and `cigar` in keeping with the concept that full length fields should be unqualified (eg, `sequence`, `sequence_alingment`, etc).

Some philosophical changes here. Not necessary, but they seem inline with our overall approach. Thoughts?


schristley:
1. Okay
2. I kinda prefer the vdj prefix. That seems to indicate full sequence to me and follows the style of the other _score, _identity, etc., fields.

javh:
2. The foreseeable problem being if the full alignment includes more than the V(D)J region. The C region in particular. (I think the leader is covered well enough by the `v` in `vdj`)

schristley:
I see. The plain names make me think they are "default" values, with `v_score` being a "specialized" value. Will IgBlast output these? I seem to remember that IgBlast only produced the info for each gene segment. If IgBlast produces values for these, and they would in essence be the main value that users would look at to determine the alignment score, etc., then I'm ok with it.

javh:
No, IgBLAST won't output these, though these are changes inspired by IgBLAST support oddly enough. At the moment, I think only partis is applicable to any of these fields (paging @psathyrella).

I was hoping for a better way to distinguish primary fields. Like using `required` and `recommended` or assigning an `x-category` tag we use for sorting them. Personally, I don't think any of the `_aa` fields should be "main" fields... But, I guess we should sort that out too at some point.

javh:
Oh, and iHMMune-align. Global alignment scores are also relevant to iHMMune-align. Though, I think the likely of that getting AIRR format support is pretty low, we do have a parser for it in MakeDb.

javh:
I'm spamming now, but... global `cigar` and `identity` fields could be also be derived fields from the individual segment fields, making them "main" in sense that they are a "compression" of the corresponding v/d/j/c fields.

schristley:
Prefix them with `sequence_`? `sequence_score`, `sequence_identity`, etc., also makes it sound like its for the full sequence.

> So... does a prefix denote a "subset field" or an "auxiliary field"? I guess that's the question

I don't like using the field name for either of those concepts. I still like the `x-airr-category` idea, and would even suggest that we reserve the `x-airr-` prefix for all specialized attributes. I don't think the names/prefixes/suffixes should be used for categorization or such. Doing any parsing of a field name to understand its "context" ends up being brittle design in the long run, and I'd prefer to avoid it. Especially when we can add as many attributes as we want into the OpenAPI spec.

psathyrella:
I also don't really like the `vdj_` prefix. If nothing else, it doesn't make sense for light chains. And while I agree that it's a bit more clear to have something there indicating that the column corresponds to the entire sequence (maybe `full_` is more clear than `sequence_`?), I dunno, what else are people going to think it corresponds to? It seems it's a tradeoff between clarity and conciseness, and I feel like it's clear enough without the prefix.

Yeah I also don't like requiring the aa fields. It's just soooo easy for even the least computer savvy user to convert, and it'll really increase the file size and complexity to double all the sequence fields.

The way my stuff is set up, I have a defined minimal set of columns that completely specify the rearrangement event, and the first thing that happens on reading in a file is all the cruft gets added back in. I.e. as long as most everyone is reading the file with the approved parsing function, they don't complain that I don't write their favorite column to file, since they only see it after it's been added back in.

```
No, IgBLAST won't output these, though these are changes inspired by IgBLAST support oddly enough. At the moment, I think only partis is applicable to any of these fields (paging @psathyrella).
```
I'm not sure I understand which bits are particularly applicable to partis? As far as scores, I do output a list of the N most likely v/d/j genes along with a score for each (where N is complicated but basically "everybody that's kind of close"). I guess in this case these would get output as separate annotation lines? That seems overly verbose/complicated, but if I don't do that then how will anyone know what the v_score means? Maybe a better way to say it, it's not clear to me how useful it is to specify just one v_score for each sequence.

javh:
@psathyrella,

> I'm not sure I understand which bits are particularly applicable to partis? As far as scores, I do output a list of the N most likely v/d/j genes along with a score for each (where N is complicated but basically "everybody that's kind of close"). I guess in this case these would get output as separate annotation lines? That seems overly verbose/complicated, but if I don't do that then how will anyone know what the v_score means? Maybe a better way to say it, it's not clear to me how useful it is to specify just one v_score for each sequence.

Ah, I must be mistaken then. I thought you had a single score value (probability?) for the entire alignment, but it looks like (from checking the manual), you have individual scores for the v, d and j (`*_per_gene_support`). 

Though, this begs the question of whether these fields are sufficient and clear for your needs:

+ `*_score`
+ `*_identity`
+ `*_evalue`

Do we need to add fields specific for probabilistic aligners, such as `*_probability`, or can we get by with having a fuzzy definition of `*_score`. If the former, we probably need to remove `*_score` from the list of required fields because it could be an either/or situation.

For multiple hits, that'll end up in the `Alignment` (hit table) spec, which we aren't finished with yet:
http://docs.airr-community.org/en/latest/formats/alignments.html


.

psathyrella:
I mean we do have a value for the whole alignment, but I don't know that it's useful to write out -- it's only really useful when comparing to other possible alignments of the same sequence. the `*_per_gene_support` fields are dicts with score/probability-ish numbers for any other possible genes that were deemed somewhat likely, e.g. `IGHD2-8*01:0.999999253081;IGHD2-8*02:7.46918510402e-07`.

So, no, the `*_{score,identity,evalue}` fields don't really seem useful for partis: comparing the v_score for one sequence's best V gene to the score for another sequence's best v gene isn't really meaningful -- what's useful is comparing the score to all the other v genes that were tried for _each_ sequence (i.e. v_per_gene_support column), but it doesn't sound like that information would go in the current format we're talking about.

So I don't strongly object to the score type fields being there, but my feeling is that it's too hard to make it meaningful across all the different aligners, since they're so different, and it'll end up being more confusing than useful.

schristley:
> comparing the v_score for one sequence's best V gene to the score for another sequence's best v gene isn't really meaningful.

This is true for IgBlast as well, not just partis, as in general an alignment score depends upon how long the sequences are and your scoring matrix. So I don't think we have that expectation, and you are correct that the score isn't that useful if you don't know the context and the tool that generated it. But that's fine, we aren't trying to create a universal score, just report the tool output.

> Do we need to add fields specific for probabilistic aligners, such as *_probability, or can we get by with having a fuzzy definition of *_score. If the former, we probably need to remove *_score from the list of required fields because it could be an either/or situation.

> Edit: I guess we could also make *_evalue fuzzy instead by calling it something like *_significance, *_evidence, *_support, etc.

Something like this sounds reasonable. I'm fine with keeping `*_score` but making it fuzzy.

jianye00:
I'd definitely vote for  using generic field names to accommodate various tools as long as people know what they represent.  igblast only provides score for each gene (though we could add up the score if people want).

Providing amino acid sequence translation is very useful for people who want that information although I don't care if its' required or not.  By the way, it is not trivial to do that as you need to know precisely which frame of your nucleotide sequence is the right one (which can be deduced from alignment to a germline sequence where the frame is known).

javh:
Okay, it seems to be that we don't really have a current use case for the full sequence scoring fields (`vdj_score/identity/evalue`).  So... how about we adhere to YAGNI and just delete these from the spec instead of worrying about the naming? We can add them to a later version of the spec if someone advocates for them and can define them better.

javh:
All right, merging this and considering the Rearrangement spec frozen.

PR #112, opened 2018-04-13
Added unittest skeleton to python package.

javh:
Just a base framework. Unlikely to be finished before release.


javh:
Yeah, I'll add a note to the README. It's just:

```
cd lang/python
python3 -m unittest
```

PR #114, opened 2018-04-16
Refactor python lib to simplify

laserson:
WIP: I significantly simplified the the Python package. Let me know what you think.

Main changes:
* Switch to pandas-based loading/parsing
* Remove notion of field order, which we actually don't mandate
* Simplified package dep

Still todo: refactor the CLI accordingly, update docs

The main idea is that we depend on all the tooling that pandas has built, both for performance and also to avoid reimplementing stuff/maintenance.


laserson:
Basically, now there are just `load_airr_data` and `dump_airr_data` that wrap pandas. The former also takes a schema spec to do the validation. There are also "helper" versions for each of the schemas, like `load_rearrangements`.

laserson:
Another note, it's nice to be able to have an iterator interface into the data files. Pandas supports this as well, and we can wrap that use case too...then we still get the pandas parser and it remains easy to do the validation.

javh:
Hrm. Here's my thoughts.

1. We need to keep the iterator interface. There are a lot of use cases for not loading everything into memory. It also breaks changeo if we don't.
2. I'm not sure we should mandate type conversion into numpy types. I'd prefer base types as an option.
3. I think we should keep column ordering, just for the sake of predictable behavior.
4. I don't see forced output of unspecified required fields as empty strings. Is it in there?

schristley:
Why are we rewriting this at the last minute, can't this be put off to the next release?

javh:
> Why are we rewriting this at the last minute, can't this be put off to the next release?

Yeah... That's another thing.  I only budgeted time to finish changeo and the airr python library until last week. This week needs to be the R package and outstanding doc tasks to meet the 4/20 deadline.

But, if we want to merge what's here to replace the current pandas interface, that's trivial, because it won't impact anything else really.

laserson:
Good point @schristley, and sorry for the last min changes, but I haven't been able to look at the code for a while till now. I'm fine pushing this into a later version, as long as we can agree that this code isn't really mature enough to guarantee interface stability...

Regarding @javh's comments...

> We need to keep the iterator interface. There are a lot of use cases for not loading everything into memory. It also breaks changeo if we don't.

Agree fully, and I can put that back in. Are you opposed to delegating the iteration to pandas, though? That way we don't need to implement the parsing twice (or at all, really), and get the performance benefits of the C code. (And also the battle-hardened CSV parsing in pandas)

> I'm not sure we should mandate type conversion into numpy types. I'd prefer base types as an option.

This is doable, though what is the issue with numpy types?

> I think we should keep column ordering, just for the sake of predictable behavior.

I'm not opposed to this either. But do we want to make this required (i.e., fail validation if mis-ordered)?

> I don't see forced output of unspecified required fields as empty strings. Is it in there?

Nope, but easy to add to to the output fn.

javh:
> Agree fully, and I can put that back in. Are you opposed to delegating the iteration to pandas, though? That way we don't need to implement the parsing twice (or at all, really), and get the performance benefits of the C code. (And also the battle-hardened CSV parsing in pandas)

The standard csv library is also in C. I'm extremely skeptical there will be any performance gain in moving from the csv library returning lines as a dict to the pandas library returning single row data frames that are then converted to a dict.

I'm open minded though. :)

schristley:
> Good point @schristley, and sorry for the last min changes, but I haven't been able to look at the code for a while till now. I'm fine pushing this into a later version, as long as we can agree that this code isn't really mature enough to guarantee interface stability...

This interface needs to be stable for this release and beyond; there really is no point doing a release, documenting and publishing a paper if we are gonna change the whole API immediately afterward. Changing the internal functional implementation is fine, but I already spent weeks working with Jason on making sure the library is stable and working properly with our current toolchains.

laserson:
I don't disagree, and I'd prefer not to make interface changes in general. (I guess I'm hrmphing at the original MiAIRR rush to release a v1.0.) Anyway, my preference would be to eliminate interface rather than change it, if possible, so we're not stuck maintaining something that might end up being awkward in the future. One of the advantages of TSV/YAML is that we really should not be parsing anything ourselves. So my preference is to eliminate interface that does that. Really the core of what the libs should do is ensure that the data looks like the proper TSV.

javh:
So `interface.py` really only exists as a place to store top level package imports which were previously in `__init__.py`

So `airr.read = airr.io.RearrangementReader`. (We could actually just replace `airr.read` with `from airr.io import RearrangementReader as read` in `__init__.py`.)

I've no objection to simplifying or removing interface, but I think @schristley had a reason for it's existence?

BTW - PR shortly that integrates the pandas version code as:

1. Added the numpy type map and truth value lists to `schema.Schema`.
2. Some minor changes to `Schema.to_bool` and `Schema.from_bool` to avoid replication for the above.
2. Put read_csv and to_csv code into load/write as comments with TODO.  Can't just replace it right now because of all the column ordering checking stuff.

javh:
Added some, hopefully not in vain, attempts at future proofing in #118.

Oh, and forgot to respond to these from earlier:

> This is doable, though what is the issue with numpy types?

I don't always want to numpy. Though, I imagine if we do decide to do a pandas implementation that creates an iterator which returns a dict for "normal" python use, then we can just convert in that and leave the DataFrame in numpy types.

> I'm not opposed to this either. But do we want to make this required (i.e., fail validation if mis-ordered)?

No, just force ordering of known followed by custom fields on write.

javh:
BTW @laserson, the R package is currently written the same way as your refactor of the python package. I also just plugged in your abstraction for read/write using a schema argument into the R package in #119. I'll add the ellipses pass through to `readr` tomorrow (equivalent to `**kwargs`). I think that'd be good to add too.

Anyway, I'm thinking that refining the structure in the R package might be the best way to tackle this for the future. Given the time constraints.

javh:
Closing this. We should keep the branch around for a while. In case we decide to redo the internals later.

Issue #29, opened 2017-10-17
Auto generate definitions.yaml from TSV?


bcorrie:
Hi All,

I was wondering if it might make sense to auto generate the definitions.yaml file from the MiAIRR TSV file rather than have a consistency check???

There are a couple of reasons for this:

1) They won't ever be out of sync (assuming the generation is done automatically). Probably worth having the Travis job still to ensure this and catch bugs in the conversion 8-)

2) The definitions would be more complete. Currently the definitions.yaml file is very minimal with name and type. The YAML object spec accepts descriptions and examples for fields in the definition. The TSV file has both field descriptions and examples. It would be nice if these could be added to the definitions.yaml file. I was going to do it manually but thought that is a lot of mundane work and will make it even harder to keep the two consistent. The driver for this is that our API uses the definitions.yaml file. But the API description is pretty bare bones as it has no descriptions or examples. It would be nice if the definitions.yaml file had this information.

3) When using the MiAIRR definitions in a swagger API spec like we are, we need definition in multiple forms. For swagger API responses, we need swagger objects. For API parameters, we need a different YAML structure, essentially a list of single parameters and their types, descriptions, and examples. Similar to #2 above, it would be nice if we could automatically generate a parameters.yaml file from the TSV.

I am happy to work on the code snippets to do this... The consistency check that we have now is a good starting point.

Thoughts?

I can create a branch for this development...

Brian

schristley:
If anything I think it should go the other way, generate the MiAIRR TSV file from definitions.yaml. Over time, the yaml file will be more inclusive of all AIRR objects, and MiAIRR can be a generated subset.

However, feel free to add the descriptions and examples to definitions.yaml.

bcorrie:
I think it is  good idea to keep a 1:1 mapping between the TSV and the YAML. One can then say THIS is the MiAIRR TSV file and THAT is the MiAIRR YAML file. If there are other things that are outside of MiAIRR I think they should be defined in a separate YAML file. That way a repository or tool like iReceptor (or VDJServer) can take the MiAIRR YAML definiton and know that they are only getting the MiAIRR definition. If they want the MiAIRR extensions that you are talking about then they can get that by composing those objects from multiple YAML files. This is just like what we are doing now with MiAIRR defining all of the MiAIRR fields and then the Formats rearrangements.yaml defining the Formats extension. 

[...]

PR #115, opened 2018-04-16
Start of autobuild TSV and re-enable consistency checks

javh:
1. Added autobuild of Rearrangements and Alignment TSVs to docs build.
2. Added `specs/miairr.yaml` with MiAIRR content of old TSV.
3. Made formats consistency check function (no refactor - just made it run.)
4. Moved consistency checks to `/tests`

Probably can't get back to this task for a bit (#29), so just want to merge what exists now for cleaner history.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,7 @@ instance/
 
 # Sphinx documentation
 docs/_build/
+docs/_downloads/
 
 # PyBuilder
 target/
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,8 @@ matrix:
       install:
         - pip install pyyaml pandas xlrd deepdiff jinja2 jsondiff
       script:
-        - scripts/check-consistency-miairr.py
+        - tests/check-consistency-miairr.py
+        - tests/check-consistency-formats.py
     - language: python
       python:
         - 3.6
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -235,23 +235,20 @@ html_context = {'airr_schema': airr_schema}
 
 # Write download spec files
 import csv
-fields = ['Name', 'Type', 'Priority', 'Description']
-specs = ['Rearrangement', 'Alignment']
 dl_path = '_downloads'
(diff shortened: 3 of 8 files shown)


schristley:
I'm okay with the PR, but do we need `miairr.yaml`? Any advantage to doing that versus putting custom attributes `x-miairr-set`, etc., in `definitions.yaml`? It just seems like we traded one file for another, just different format.

bussec:
@schristley see our comments [here](https://github.com/airr-community/airr-standards/issues/29#issuecomment-381345810). I don't have any strong opinion about one or files, but would prefer having everything in YAML (instead YAML+TSV).

javh:
So something like this?

**definitions.yaml**

```
Study:
   discriminator: AIRR
   type: object
   properties:
      study_id:
         type: string
         description: Unique ID assigned by study registry
         x-miairr-set: 1
	 x-miairr-subset: study
	 x-miairr-name: Study
	 x-miairr-format: free text
	 x-miairr-example: PRJNA001
```

I was originally thinking more like this:

**miairr.yaml:**

```
Study:
    set: 1, study
    format: free text
    example: PRJNA001   
```

**definitions.yaml:**

```
Study:
   discriminator: AIRR
   type: object
   properties:
      study_id:
         type: string
         description: Unique ID assigned by study registry
         x-miairr: Study
```

So the `miairr.yaml` file is setup with the MiAIRR standard as the top level and the formats spec contains a cross-reference.

The first is easier to maintain, but the second retains MiAIRR independence, which is probably both good and bad. Either way (or an entirely different way) will work to autogenerate the tsv.

schristley:
> So something like this?

Yeah, but now that I've re-read the discussion, I'm good with a separate `miairr.yaml` file. The object schemas in `definitions.yaml` are actually preliminary and weren't approved by MiAIRR, so letting them have their own file retains their independence. I'll note that `miairr.yaml` is not OpenAPI compliant for schema/field definitions though, so we might change it later if that ends up being needed.

javh:
Yeah, we can assess the format of `miairr.yaml` later. Just want something in place for now.

PR #118, opened 2018-04-16
Integration of pandas template code

javh:
Integrate code from #114 to preserve it for potential future work.

No actual changes to the interface except for a small tweak to how `schema.Schema.to_bool` and `schema.Schema.from_bool` work. (Should be faster.)


schristley:
Can we use polymorphism and simple read, load, validate, etc., function names? I realize that airr.read() is ambiguous if we want to support these operations on more than just rearrangements (I see alignments, metadata and workflow provenance), isn't it better to have classes for them for encapsulation. Maybe something like:

```
airr.rearrangement.read()
airr.alignment.read()
airr.metadata.load()
```

and so on? It's more OO-like in my mind though @javh can say if its more pythonic or not.

javh:
Polymorphism is very pythonic. Could also also be separate modules instead of a class. There's also the option of adding a type argument. Eg, `airr.read(file, schema="rearrangement")`.

I've no strong opinion on if/how to implement it - it's all kinda the same to me. Aside from "whatever is the least work this week". :)

For what's it's worth, the R package will have both of these:

```
read_airr(file, schema=RearrangementSchema)
read_rearrangement(file)
```

(There's no reasonable equivalent to airr.rerrangement.read(file) for a filename in R.)

PR #125, opened 2018-04-23
Added R package documentation to main docs site

javh:
First pass. Still really hacky to build, but the script is at `lang/R/docs/build.R`.

1. Added the docs to the main site.
2. Made the citations in the vignette manually entered, because auto-populating from the bibtex file wasn't working.
3. Some tweaks to layout in readme and news.


ssnn-airr:
> Made the citations in the vignette manually entered, because auto-populating from the bibtex file wasn't working

I will add manually in the future.

javh:
Thanks. The way you did the citations is better, but it wasn't working for the rst conversion. It might be possible to get it to work, but I took the low effort approach for now.  Will look into it more at a later date.

PR #128, opened 2018-04-25
First pass at writing logicals as T/F

ssnn-airr:
Probably not the most elegant code


javh:
Eh. Elegance can wait. No rush on that. I'm sure we'll have cleanup to do later.

Though, you could probably simplify by going through a factor. Eg, `factor(data[[log_field]], levels=c(TRUE, FALSE), labels=c("T", "F"))`.

 The [forcats](http://forcats.tidyverse.org) package might have something good for this purpose. Have to check that though.

PR #130, opened 2018-04-28
Added DataRep usage example

javh:
I added another top level folder - `examples. Not in love with the idea, but it didn't (presently) seem appropriate to put it in the docs.


schristley:
I think it belongs in the docs.

javh:
Where should it go in the docs? A new "Examples & Tutorials" section?

schristley:
"Use cases", "example workflows", all sound good. I was thinking of something similar like the workflows section you have for ChangeO.

PR #131, opened 2018-04-30
Update citation links

bussec:
The MiAIRR manuscript is now freely available at PubMed Central.
Add/update the links in the reference section.


javh:
Looks good to me.

PR #134, opened 2018-05-02
remove foreign id fields

### LINKED ISSUE #133: get rid of foreign id fields for MiAIRR object
These aren't specified within MiAIRR but were added into the schema. However, having them in the schema is mandating a particular data model (and relationship between the entities), which (at least in CRWG) we are attempting to avoid and allow the back-end repository to define how they wish. We should remove them.

schristley:
fixes #133 

To be merged after the release.


javh:
Looks like travis is failing because the specs need to be propagated into the lang/python and lang/R locations.

PR #148, opened 2018-08-10
rewrite validation

schristley:


schristley:
I'm not sure why the R:release travis check is failing. It also appears to be failing on the master branch.

javh:
I don't know why R-release is failing either - it's a very unclear error message. I fixed one issue in the `rearrangement_updates` branch, but there's still another one. I'm going to sit down and try it figure it out today - it's some sort of version incompatibility, because the check versions are new but the code didn't change.

BTW - we're going to have merge conflicts, because we both started branches for the same issues...

schristley:
I was gonna use your branch but seeing as I was playing around with the right implementation, I didn't want to mess up your work. How about we merge this into your branch instead of master?

javh:
Yeah, let's try that.  Give me a couple hours to get to the R issue.

ssnn-airr:
Maybe this is related to the R issue. If travis is using https://cran.cnr.berkeley.edu, this could be the problem. I am having errors building images in other project because the site can't be reached.

`
Installing package into '/usr/lib64/R/library'
(as 'lib' is unspecified)
Warning: unable to access index for repository https://cran.cnr.berkeley.edu/src/contrib:
  cannot open URL 'https://cran.cnr.berkeley.edu/src/contrib/PACKAGES'                   
Warning message:                                                                         
package 'devtools' is not available (for R version 3.4.2)                                
Error in loadNamespace(name) : there is no package called 'devtools'                     
Calls: :: ... tryCatch -> tryCatchList -> tryCatchOne -> <Anonymous>                     
Execution halted     
`

javh:
I *should* be using https://cloud.r-project.org (default).

javh:
Hrm. Okay, we're probably having different issues with that.  I just fixed the travis error in the rearrangement_updates branch. I still don't know exactly what caused the error though, so I'm just hoping it's a permanent fix.

ssnn-airr:
@javh If you see all the immcantation builds failing, this above is the issue. I switch to another problem. I have several dramas running in parallel.

javh:
The builds are fine on docker hub. Sometimes the Yale network misbehaves.

PR #150, opened 2018-08-20
info object

### LINKED ISSUE #147: Info object in airr-schema.yaml?
I'm thinking of adding a simple Info object into the yaml file that has a description, version, URL, etc. Essentially stuff that was in the [original schema](https://github.com/airr-community/airr-formats/blob/master/specs/rearrangements.yaml) from long ago.

I realize this becomes another version and etc. to maintain but I think it might be useful for tools that use the schema but don't use a reference library (or can access the underlying git info)

schristley:
fixes #147 

My proposal for an Info object in the schema. Not sure if the list of schema is needed or not, but I added as a way to be more specific versus just relying upon whatever is in the file. There should be an API mechanism to get this object.

### DIFF
--- a/docs/packages/airr-R/overview.rst
+++ b/docs/packages/airr-R/overview.rst
@@ -30,6 +30,7 @@ R commands from the package root ``lang/R``::
     install_deps(dependencies=T)
     document()
     install()
+    test()
 
 .. toctree::
     :maxdepth: 3
--- a/lang/R/R/Schema.R
+++ b/lang/R/R/Schema.R
@@ -8,6 +8,7 @@
 #' @slot    required    \code{character} vector of required fields.
 #' @slot    optional    \code{character} vector of non-required fields.
 #' @slot    properties  \code{list} of field definitions.
+#' @slot    info        schema information.
 #'                  
 #' @details
 #' The following predefined Schema objects are defined:
@@ -103,7 +104,10 @@ load_schema <- function(definition) {
     # Load definition
     definition_list <- spec_list[[definition]]
     definition_list[["discriminator"]] <- definition_list[["type"]] <- NULL
-    
+
+    # schema info
+    info <- spec_list[["Info"]]
+
     # Define member attributes
     fields <- names(definition_list[["properties"]])
     properties <- definition_list[["properties"]]
@@ -120,7 +124,7 @@ load_schema <- function(definition) {
         properties[[f]][["description"]] <- stri_trim(y)
     }
     
-    return(new("Schema", required=required, optional=optional, properties=properties))
+    return(new("Schema", required=required, optional=optional, properties=properties, info=info))
 }
 
 
--- a/lang/R/R/Schema.R
+++ b/lang/R/R/Schema.R
@@ -8,7 +8,7 @@
 #' @slot    required    \code{character} vector of required fields.
 #' @slot    optional    \code{character} vector of non-required fields.
 #' @slot    properties  \code{list} of field definitions.
-#' @slot    info        schema information.
+#' @slot    info        \code{list} schema information.
 #'                  
 #' @details
 #' The following predefined Schema objects are defined:
@@ -27,7 +27,8 @@
 setClass("Schema", 
          slots=c(required="character",
                  optional="character",
-                 properties="list"))
+                 properties="list",
+                 info="list"))
 
 
 #### Methods #####
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1,6 +1,10 @@
 #
 # Schema definitions for AIRR standards objects
 #
+Info:
+    description: Schema definitions for AIRR standards objects
+    version: 1.3
+    url: https://github.com/airr-community/airr-standards/releases/download/v1.3.0/airr-schema.yaml
 
 # The overall study with a globally unique study_id
 Study:
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -2,9 +2,15 @@
 # Schema definitions for AIRR standards objects
 #
 Info:
+    title: AIRR Schema
     description: Schema definitions for AIRR standards objects
     version: 1.3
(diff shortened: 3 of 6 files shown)


laserson:
What's the purpose of this object?

schristley:
The [original schema](https://github.com/airr-community/airr-formats/blob/master/specs/rearrangements.yaml) had some simple info like a description, version, url, etc., that I think is useful.

javh:
Looks fine to me. Except, I think the schema list is unnecessary and a potential cause of errors (by failing to replicate the object set correctly). This information already exists in the schema file as the set of member objects.

laserson:
Ditto. I don't think the schema stuff needs to be there. But otherwise I think it's good.

schristley:
Not sure what you mean by "member objects", but I see by studying the code more that a `Schema` object is for a specific definition, not for the whole schema file. So I think you need to need know the names of the definitions beforehand, unless I'm misunderstanding something.

I removed the schema list, and made some minor code changes to prevent loading "Info" as a definition and make info a variable e.g. `airr.schema.RearrangementSchema.info`

javh:
> Not sure what you mean by "member objects", but I see by studying the code more that a Schema object is for a specific definition, not for the whole schema file. So I think you need to need know the names of the definitions beforehand, unless I'm misunderstanding something.

Sorry, by "member objects" I mean the other top level items in the yaml file (the individual schema). Needing to know the schema name before loading it is just a quirk of how we've implemented it. We can change that, though I don't know that we should. You can get the list of schema from just reading the entire yaml file.

javh:
Oh, and do we need `info.description` for the python API? We could just populate `Schema.version` and `Schema.url`. If `info` is just a blob of unknown data, then it'll be easier to change what it contains, but also less obvious what it does contain... So I don't have a strong opinion. 

We should do the same thing in R though.

schristley:
> Needing to know the schema name before loading it is just a quirk of how we've implemented it. We can change that, though I don't know that we should. You can get the list of schema from just reading the entire yaml file.

That's fine. That schema list was a way to give them, but I don't it's that important, the definitions are gonna be pretty much set in stone after awhile.

schristley:
> We should do the same thing in R though.

I'm working on the R code now. I'm not sure how to test though, I've never try to use the R library yet. Does travis run some tests for R?

schristley:
> Oh, and do we need info.description for the python API? We could just populate Schema.version and Schema.url. If info is just a blob of unknown data, then it'll be easier to change what it contains, but also less obvious what it does contain... So I don't have a strong opinion.

That information needs to be in the schema file itself, not just properties coded in the python Schema object, if that is what you are asking. For the very reason you mention before, people may directly just read the schema file (e.g. that is exactly what I'll be doing for some javascript UI code).

javh:
There are Travis tests. The R package has unit tests (using testthat), which you can run manually using: ``devtools::test()``.

More details here:
http://r-pkgs.had.co.nz/tests.html

javh:
> That information needs to be in the schema file itself, not just properties coded in the python Schema object, if that is what you are asking. For the very reason you mention before, people may directly just read the schema file (e.g. that is exactly what I'll be doing for some javascript UI code).

No, I meant we could instead assign `Schema.version` and `Schema.url` directly from the `Info` section of the yaml file, instead of having `Schema.info.version` and `Schema.info.url`. And probably skip `info.description` because it's a level above and therefore not really correct information. There could be a 'Rearrangement.description' entry in the yaml file though, that we could populate that from, if we think it's necessary.

schristley:
> No, I meant we could instead assign Schema.version and Schema.url directly from the Info section of the yaml file, instead of having Schema.info.version and Schema.info.url.

I see. Yeah that's a good point. The OpenApi has an [Info schema definition](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#infoObject) and I think we should follow it. Though I see now I'm missing the required `title` field...

javh:
> The OpenApi has an Info schema definition and I think we should follow it. Though I see now I'm missing the required title field...

Sounds good to me. Then we can rely on those fields being permanent.

schristley:
@javh I'm hacking my way through the R stuff, I went to the lang/R directory and did these commands:

```
R -e 'install.packages(c("devtools", "roxygen2", "testthat"))'
R -e 'devtools::install_deps(dep=T)'
R -e 'devtools::document()

Updating airr documentation
Loading airr
Error in (function (dep_name, dep_ver = NA, dep_compare = NA)  : 
  Dependency package readr not available.
```

I needed to install `readr` and `yaml` packages before `document()` gave me the real error I'm seeing in the travis build. Should those packages be added as dependencies?

javh:
> I needed to install readr and yaml packages before document() gave me the real error I'm seeing in the travis build. Should those packages be added as dependencies?

No, the are already listed as dependencies. See the `Depends` and `Imports` sections of the `DESCRIPTION` file. Those commands should be fine. (The first two shouldn't even be necessary in the Travis build, but R in Travis isn't behaving well.)

The error you're getting is due to a missing slot spec in the class definition. You need:

```
setClass("Schema", 
         slots=c(required="character",
                 optional="character",
                 properties="list",
                 info="list"))
```

(And to specify the type in the roxygen `@param` field, but that won't cause an error. It's just incomplete docs.)

schristley:
green for go!

schristley:
@javh Did you already review? I'd like to merge if no objections.

javh:
Oh, sure, I thought we were swapping doing the swap from stuff like `Schema.info.version` to `Schema.version`? Doesn't need to be now if you want to merge.

schristley:
> Oh, sure, I thought we were doing the swap from stuff like Schema.info.version to Schema.version? 

We can do that specifically for `version` if you want, but there are a bunch of fields defined by the OpenAPI spec for the Info object, and it also allows [Specification Extensions](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#specificationExtensions), i.e. custom attributes that start with `x-` so it seems easier to just leave it all under `Schema.info`

javh:
Gotcha. Nevermind then. Sorry, I would've pushed approve last week if we hadn't gotten our wires crossed.

PR #151, opened 2018-08-28
Sw tools initial

williamdlees:
Incorporated first version of Software Tools Standard, as approved by the Software WG this month.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -4,17 +4,13 @@ AIRR Standards
 The AIRR Community is developing a set of standards for describing, reporting,
 storing, and sharing adaptive immune receptor repertoire (AIRR) data, such as
 sequences of antibodies and T cell receptors (TCRs). Some specific efforts
-include
+include:
 
-- MiAIRR for describing minimal information about AIRR datasets, including
++ The MiAIRR standard for describing minimal information about AIRR datasets, including
   sample collection and data processing information.
-
-- Data representation (file format) specifications for storing large amounts of annotated AIRR data.
-
-- APIs for exposing a common interface to repositories/databases containing
-  AIRR data.
-  
-- A community standard for software tools, allowing conforming tools to gain community recognition.
++ Data representations (file format) specifications for storing large amounts of annotated AIRR data.
++ APIs for exposing a common interface to repositories/databases containing AIRR data.
++ A community standard for software tools which will allow conforming tools to gain community recognition.
 
 .. toctree::
     :maxdepth: 3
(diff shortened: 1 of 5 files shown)


schristley:
For the compliance checklist, where do tool developers send the info and where is the information recorded, or is that yet to be determined?

williamdlees:
Good questions, Scott.

I have added contact emails to the questionnaire so that submitters know where to send it. Would be good to replace these with some official AIRR emails over time - looking at the AIRR website, the WGs don't seem to list contact points today.

We would like to publish a list of tools on the AIRR website but this may take some time to organise. We can also put a list in this document, in the same way that there is a list of MiAIRR compliant sw. We will definitely put a list into b-t.cr, alongside the other lists that are there. Don't really want to put pointers or blank tables in place until there's at least one tool on the list!

javh:
Looks pretty good. I just pushed a few changes. Mostly just typos and reorganizing the header levels and names in the toc tree.

One potentially contentious change is that I moved the  `Standard Data Sets` topic to an entirely separate `Evaluation Data Sets` page, because it seemed really out of place in the Software Guidance section.

One change I didn't make, which I think is worth considering, is to turn the checklist table into a tsv file and import it via the `csv-table` directive. See `resources/support.rst` for an example. It's in no way necessary, but I find imported table a little easier to maintain than the clunky sphinx syntax for them.

bcorrie:
@williamdlees I can create email aliases in the airr-community.org domain. Let me know what you would like. It probably makes sense to have a contact email for each working group as you suggest, but you might want a separate email for the formal submission.

I should note that there is discussion at the Communications and Exec going on around official emails being @airr-community.org or @antibody-society.org. In the short term I am happy to create the airr-community.org emails.

schristley:
I would be a bit hesitant to create emails for the WGs. Technically, the WGs are meant to be temporary, they do a defined set of tasks then disband. However, there are standing committees, e.g. IARC for inferred allele submissions, which presumably would continue forever, so those make sense for a reserved emails. Presumably the software WG will propose a standing committee to handle software submissions.

Also, as the WGs are public, i.e. anybody can join, I think the google groups we have setup work well for that. Though it would be good to have appropriate links on the web pages if we don't already.

javh:
@schristley, ah, that makes sense. I guess we'd rather user problems come through GitHub issues too.

bcorrie:
@javh I have no control over antibody-society.org, only over airr-community.org. There is discussion/debate at the Comms Working Group going on about which is the right one to use.

bcorrie:
Our google group naming conventions are a bit inconsistent as well 8-)

williamdlees:
@bcorrie could you please create an alias tools-standard@airr-community.org ? If possible this should forward to both william@lees.org.uk and chaim.schramm@nih.gov, if it can only point to one person, then to me, please.

williamdlees:
@all thanks to everyone for input and in particular to Jason for reformatting and typos. Is everyone ok for me to push the merge button now?

javh:
Have at it.

PR #155, opened 2018-09-17
MiAIRR refinement

bussec:
* Fix mapping to SRA library_* term
* Clarify examples for submitter information
* Include controlled vocabularies for some field


bussec:
Hy @graybeal, could you have a quick look whether there is anything that should be included? TNX!

bussec:
Thanks @javh!

PR #156, opened 2018-09-25
common repository API and metadata schema

### LINKED ISSUE #56: Is diagnosis & intervention data really n-to-1 with subject?
In an earlier meeting, it was mentioned that there could be multiple diagnosis&intervention data records associated with a subject. Furthermore, these records are independent from the sample records associated with a subject, which is a n-to-1 relationship. Is it true that diagnosis & intervention data is n-to-1 with subject?

The reason I ask is because it introduces confusion about how to "flatten" records. For example, with NCBI, the MiAIRR objects are de-normalized into a single BioSample record. If a subject has (say) 3 diagnosis&intervention records, and (say) 5 samples, how many BioSample records is that?

### LINKED ISSUE #132: Multiple target loci in a sequencing run
A number of experimental setups (e.g. linkage PCR, single-cell PCR) have more than one target locus. 
However, the keywords `pcr_target_locus`, `forward_pcr_primer_target_location`, `reverse_pcr_primer_target_location`, which are currently expected to hold only a single locus or index. Can we refine these field as holding arrays of strings/integers?

### LINKED ISSUE #144: normalized data model for metadata
The MiAIRR data elements are currently specified in a denormalized way, which introduces a lot of redundancy when thinking about storing it in a file format to be used by analysis tools, as well as providing data entry screens to manipulate it.

My initial attempt at a normalized data model (which is in VDJServer v1.1) follows the 1-to-n data model as originally documented in the yaml spec:

- 1-to-n relationship between a study and its subjects
- 1-to-n relationship between a subject and its diagnoses
- 1-to-n relationship between a subject and its samples
- 1-to-n relationship between a sample and processing of its cells
- 1-to-n relationship between a CellProcessing and processing of its nucleic acid

However, I discovered after using it for awhile that it still has a very large amount of redundancy. The relationship between Study/Subject/Sample is okay, it is really the relationship between Sample, Cell Processing and NucleicAcidProcessing where much of the redundancy occurs.

Let's focus on those three objects and use a simple example to illustrate. Say we have 50 blood samples for TCR sequencing where we do flow sorting to separate CD4 and CD8 cells. At the sequencing output, we have 100 FASTQ files with the raw data.

- 50 Sample records for the blood draws.
- 100 CellProcessing records to document the CD4 and CD8 flow sorting of samples.
- 100 NucleicAcidProcessing records to document the sequencing runs. In this case there are no technical replicates so its a 1-to-1 relationship with CellProcessing.

Notice that we did save redundancy by only having 50 Sample records instead of 100, but a large quantity of data in the CellProcessing and NucleicAcidProcessing records is redundant. The protocol to do the flow sorting is identical for each sample, so why do we need to detail it out 100 times? Likewise, the protocol for the sequencing step is also identical. What we would really want to enter is something like this:

- 50 Sample records for the blood draws.
- 2 CellProcessing records, one for CD4 and one for CD8 flow sorting.
- 1 NucleicAcidProcessing record to document the sequencing protocol.

Now we still need to link this information to the 100 FASTQ files we got from the sequencer, but this would be a simple tuple such as (filename, sample_id, cell_processing_id, nucleic_acid_processing_id) to link the pieces together.

So why not do this? Well, I was working on this very idea (@laserson also independently when thinking about a metadata file format, and CRWG for querying metadata) but ran into a roadblock. Specifically while some data elements are "general" and can be normalized, others are "specific" and thus cannot be normalized. So for the above example, in CellProcessing, `cell_subset` is general, it's "CD4 T-cell" and "CD80 T-cell" for the two records. However, `cell_number` is specific to each physical blob of cells being processed, so there are 100 values that need to be stored.

What's especially annoying is that for a bunch of studies that we've (i.e. VDJServer) made MiAIRR-compliant, we don't have information for most of those "specific" fields, so they are just "NA" or "unknown".

How to handle this? One simple, naive idea is to tag each data element as either "general" or "specific" so applications have a hint about which is which. Another idea is to re-structure the yaml spec to define a specific normalized data model.

### LINKED ISSUE #145: NucleicAcidProcessing object does not properly represent re-sequencing of libraries
Replying to a question from @schristley related to #144, I realized that including information identify specific sequencing runs into the "Processing (Nucl. Acid)" subset (`NucleicAcidProcessing` object in the [AIRR schema](https://github.com/airr-community/airr-standards/blob/f25fcf0d5ebb4d2f9c14b112019d25c75d43f479/specs/airr-schema.yaml#L233)) creates problems as it creates a 1:1 relation between a DNA library and a sequencing run.

The best example for this the re-sequencing of an existing library to obtain more reads: As both the `sequencing_run_id` and `sequencing_run_date` would be different, this would require an additional `NucleicAcidProcessing` record. However, such a record would imply the generation of an additional library (i.e. a technical replicate) instead of the re-sequencing that was actually performed.

The suggested solution for this is to move the following fields into the `RawSequenceData` object, which then will be renamed to `SequencingRun` to reflect this:

`total_reads_passing_qc_filter`
`sequencing_platform`
`read_length`
`sequencing_facility`
`sequencing_run_id`
`sequencing_run_date`
`sequencing_kit`

Note that this will not require any changes to MiAIRR itself, as we consider it to be fully denormalized anyhow. Nevertheless, if someone is aware of any pitfalls regarding this approach please let us know.

schristley:
fixes #145 
fixes #56 
fixes #144 
fixes #132 

This PR has schema implementation for a field with ontological backing, look at `organism` field in the `Subject` schema. [Discussion](https://github.com/airr-community/common-repo-wg/issues/24) about this implementation.


schristley:
> One question - do we want to merge this into master at this time? My thoughts would be that we only want to merge to master when we have a "release" ready to go?

Merges to master have to pass the checks but they are not "releases" exactly, we decided to forgo an "integration/develop" branch for airr-standards. The current check failure is to due to some minor changes that need to be made for #145, and I'm working to resolve.

While this has the API, it also the metadata schema that the DRWG has been discussing, we want that in master so we start thinking about how to support in the python/R libraries as well as how to document. That work should move forward independently of the CRWG-API.

The service code I'm writing to test the API is [here](https://bitbucket.org/vdjserver/api-js-mongodb/src/master/).

bcorrie:
OK, if we aren't using master for more structured releases, then the merge is probably fine.

We are working on our Turnkey and its service JS code at the moment. Once that is working, we will probably look at implementing the AIRR API on our Turnkey implementation as well...

schristley:
@bussec added support for multiple target locus #132 , our consistency script isn't very sophisticated so instead of writing a more complicated script that handles arrays of objects, I just slightly modified the category name in the MiAIRR TSV. Please take a look and verify you are okay with the changes.

bussec:
As discussed, in general I am ok with this. We just need to make sure that the `miairr_subset` terms can still be recognized as subnodes. However, IMO it would be best to first solve the rearrangement->metadata problem, as we might decide to split `processing/sequencing` from `processing/nucl. acid` (to handle library re-sequencing).

javh:
@bussec, was your preferred nomenclature `NucleicAcidProcessing` and `NucleicAcidSequencing`?

bussec:
@javh, the designation is fine with me, just want to point out that the way to get from the `miairr_subset` names to the names of the YAML object is hard to automate:

* `process (cell)` -> `CellProcessing`
* `process (nucleic acid)` -> `NucleicAcidProcessing`
* `process (sequencing)` -> `NucleicAcidSequencing`

However, I agree that `SequencingProcessing` would likely be confusing as it sounds like it describes something that is actually described in `Software Processing`.

Thus +1 for `NucleicAcidSequencing`.

schristley:
> Thus +1 for NucleicAcidSequencing

right now the entity is called `SequencingRun` but it can be easily changed to `NucleicAcidSequencing` if that is preferable.

bussec:
@schristley, I just saw that there is a conflict in `AIRR_Minimal_Standard_Data_Elements.tsv` due to some changes that I brought into `master` via #155. What are your plans for this PR, do you want to merge rather soon (i.e. I fix this now)?

schristley:
@bussec I'm still waiting for approval from Data Rep WG, the next WG meeting is Dec. 10 and I'm hoping can merge fairly soon after that. You can fix the conflict now if you want or I will do it later as it needs to be done before the merge.

javh:
My idea from the call was just to try to flatten the Repertoire a little, if possible, and alter property names to be non-redundant with MiAIRR object.  Maybe:

```
Repertoire:
    discriminator: AIRR
    type: object
    properties:
        repertoire_id:
            type: string
            description: Identifier for the repertoire object.
        specimen:
            $ref: '#/Study'
            $ref: '#/Subject'
            diagnosis:
                type: array
                items:
                    $ref: '#/Diagnosis'
        library:
            type: array
            items:
                allOf:
                    - $ref: '#/Sample'
                    - $ref: '#/CellProcessing'
                    - $ref: '#/NucleicAcidProcessing'
                    - $ref: '#/SequencingRun'
        analysis:
            type: array
            items:
                allOf:
                    - $ref: '#/SoftwareProcessing'
                    - $ref: '#/RearrangementSet'
```

Not sure about the structure of `analysis`.

schristley:
> My idea from the call

That's an interesting idea. I kind of see what you are doing and it doesn't seem to increase/decrease the complexity of the structure really, just a different organization. I tend to like the name `sample` over `library` as `sample` is more intuitive to me, and `specimen` seems like a synonym for `subject` yet it contains study data. So, I'm not really sure of what are the advantages to this structure in comparison to other?

It does look like it would be challenging to pull out just the subject data, for example if you wanted to normalize the subject data. You would have to know what fields to pull out of `specimen` and I don't see how you would know whether any custom fields belonged to subject or the study.

javh:
The reasoning behind `specimen` and `library` was to avoid ambiguity with the existing MiAIRR `Sample` object. So you don't run the risk of messing up a conversion from a flat to nested representation. 

Though, `specimen` might be unnecessary in favor of just having the refs to `Study`, `Sample` and `Diagnosis` at the top level.

schristley:
Okay, I see, study and subject are both the exact MiAIRR objects, but sample isn't. I guess we could move the sub-pieces of sample to the top level (I've thought about this structure before too, which is why I was waffling on whether we needed more identifiers fields e.g. nucleic_acid_processing_id).

```
Repertoire:
    discriminator: AIRR
    type: object
    properties:
        repertoire_id:
            type: string
            description: Identifier for the repertoire object.
        study:
            $ref: '#/Study'
        subject:
            $ref: '#/Subject'
        diagnosis:
            type: array
            items:
                $ref: '#/Diagnosis'
        sample:
            type: array
            items:
                    - $ref: '#/Sample'
        cell_processing:
            type: array
            items:
                    - $ref: '#/CellProcessing'
        nucleic_acid_processing:
            type: array
            items:
                    - $ref: '#/NucleicAcidProcessing'
        sequencing_run:
            type: array
            items:
                    - $ref: '#/SequencingRun'
        sequence_annotation:
            type: array
            items:
                $ref: '#/RearrangementSet'
```

The one thing I don't like about this structure is that it breaks up the array of sample's data, i.e. how do you match entries together across the arrays?

javh:
Yeah, that's by big concern - making sure we don't give two different objects the same name.

Hrm. Would something inbetween work?

```
Repertoire:
    discriminator: AIRR
    type: object
    properties:
        repertoire_id:
            type: string
            description: Identifier for the repertoire object.
        study:
            $ref: '#/Study'
        subject:
            $ref: '#/Subject'
        diagnosis:
            type: array
            items:
               $ref: '#/Diagnosis'
        library:
            type: array
            items:
                allOf:
                    - $ref: '#/Sample'
                    - $ref: '#/CellProcessing'
                    - $ref: '#/NucleicAcidProcessing'
                    - $ref: '#/SequencingRun'
        analysis:
            type: array
            items:
                allOf:
                    - $ref: '#/SoftwareProcessing'
                    - $ref: '#/RearrangementSet'
```

schristley:
Or this which keeps the sample pieces together but explicitly names each of the sub-pieces so that they match the MiAIRR objects.

```
Repertoire:
    discriminator: AIRR
    type: object
    properties:
        repertoire_id:
            type: string
            description: Identifier for the repertoire object.
        study:
            $ref: '#/Study'
        subject:
            $ref: '#/Subject'
        diagnosis:
            type: array
            items:
               $ref: '#/Diagnosis'
        sample:
            type: array
            items:
                allOf:
                    - $ref: '#/Sample'
                    - cell_processing:
                        $ref: '#/CellProcessing'
                    - nucleic_acid_processing:
                        $ref: '#/NucleicAcidProcessing'
                    - sequencing_run:
                        $ref: '#/SequencingRun'
        analysis:
            type: array
            items:
                allOf:
                    - $ref: '#/SoftwareProcessing'
                    - $ref: '#/RearrangementSet'
```

javh:
Seems like you could also put the `SoftwareProcessing` and `RearrangementSet` refs under `sample` with that structure.

schristley:
It depends if you are thinking about `SoftwareProcessing` for each specific entry in the sample array, or whether it refers to all of the entries in the sample array. I'm thinking the latter, i.e. the `SoftwareProcessing` object would be identical for each entry in the sample array, so putting it under sample explodes the size of the data (duplicate entries).

Another way to think about it is that all of the sample entries should be considered merged together as a whole and processed together. Generally there will only be one sample entry in the array, corresponding to a single sequencing run, but you might technical replicates and/or multiple sequencing runs (to increase coverage), and thus want to consider all of those files together as part of the same repertoire.

schristley:
Another criteria I am trying to meet is not to have the object be too deep. The more deep it goes then the longer the field names have to be for doing a query. In PL structure, doing a query on the cell_subset would use field name `sample.cell_subset`. With the above structure, that field name becomes `sample.cell_processing.cell_subset`, which isn't too onerous but any more levels and it starts to become excessive.

bcorrie:
Hi All,
I wanted to add another use case for multiple software_processing runs that might apply to a single repertoire that we have been talking about internally in iReceptor recently. This is pretty far outside my domain of expertise, so let me see if I can explain it in my non-expert terms... I hope I don't totally destroy the terminology here. 

This is about which germline set you use when running an annotation tool. My understanding is that although relatively uncommon today, it is possible to use different germlines with the annotation tools, and annotation can be more accurate if you use more specific germline sets for an individual. Ultimately, you could determine an individual's germline set and use that for annotation, which would give you the best rearrangement annotation...

This is related to issue: 
- https://github.com/airr-community/airr-standards/issues/157 (our germline DB colleagues) (#157)
- https://github.com/airr-community/airr-standards/issues/144 (#144) 

The germline is currently a feature of an individual rearrangement, when I think it makes more sense to have this as a feature of a repertoire, with it possible to have different rearrangement_sets using different germlines from a single repertoire.

This has two impacts:
- We probably need a MiAIRR germline field at the repertoire level? And it might not make sense to have this at the rearrangement level any longer. In fact, there is probably a fair bit of value in being able to store (or at least link to an external) germline set for two different cases. One is the germline set that is used for annotation, but the other might be the germline set that is know for the individual. If I am not mistaken, one is a biological thing (this individual has this germline set - which you may or may not know) and the other is a software processing thing (this is the germline set that was used to generate this set of rearrangements).
- This is another use case for having multiple different rearrangement_sets from a single repertoire.

I hope I got that at least partially correct 8-)

javh:
Yeah, I can see how depth could be an issue.  I'm not sure what to do about, because this structure is setup to have `Sample` as an array, and each `Sample` can have a different chain of (`CellProcessing`, `NucleicAcidProcessing`, `SequencingRun`, `SoftwareProcessing`), so it's inherently deep.

Hrm. I dunno. Maybe you could move the refs to (`Sample`, `CellProcessing`, `NucleicAcidProcessing`, `SequencingRun`, `SoftwareProcessing`) into the `RearrangementSet` object and then have `Repertoire` contain only (`Study`, `Subject`, `Diagnosis`) and an array of refs to each `RearrangementSet`? Then maintain some sort of flat table that links lower level stuff like `CellProcessing` to each `RearrangementSet`? Not sure that gets you any performance.  

As long as the ambiguous naming issue is resolved, my only real concern is addressed. Which is to make sure that a mostly-flat representation in a yaml file (a) fits MiAIRR and (b) can be restructured easily.  I think after that it's largely CR-WG territory to come up with a structure that works with your implementations. I don't think I know enough about that to provide good input, so whatever works for you. :)

schristley:
> My inclination would be to put the germline database link/description in RearrangementSet or SoftwareProcessing

I agree

laserson:
Shouldn't the `diagnosis` objects be inside the `subject` object?

schristley:
> Shouldn't the diagnosis objects be inside the subject object?

yes

laserson:
I thought we were considering a `repertoire_id` to correspond to a tube in a freezer. Why is `sample` an array type?

schristley:
> I thought we were considering a repertoire_id to correspond to a tube in a freezer. Why is sample an array type?

To handle the situations like technical replicates or deep sequencing, i.e. essentially separate sequencing runs that you want to merge together into the same repertoire.

bussec:
IIRC, the two points that substantially influenced the decision of putting the GL reference information into the individual `rearrangement` were:
1. We felt that the information should not be in Set 5 (i.e. `SoftwareProcessing`) but in Set 6 (processed sequences), as we considered it metadata for the segments calls. However, as MiAIRR does not have a structure that would be equivalent to `RearrangementSet`, the information went into the individual `Rearrangements`.
2. Having the information attached to the individual `rearrangement` made NCBI implementation straight-forward as it did not require the creation of custom keywords. It was also influenced by requests from the community to have things findable with a normal search engine, which requires the information to be in the same HMTL (i.e. Genbank record).

Having said this, I agree that this decision creates a huge redundancy and `RearrangementSet` would be the more appropriate place for this information (as using multiple GL reference withing one `RearrangementSet` would be pointless.

laserson:
>To handle the situations like technical replicates or deep sequencing, i.e. essentially separate sequencing runs that you want to merge together into the same repertoire.

@schristley is it possible to isolate the replicate info from the rest of the metadata? If you take literally the same library and dump it into a sequencer multiple times for several different runs, all the metadata should essentially be identical, except for the fact that there is a replicate.

laserson:
>I have a study where they performed additional blood draws to get more sequence data, so its gone through a separate processing to get a different tube of stuff.

This seems to me like something that should probably live in user-land, rather than build the structures necessary to model this into the general use case. I would argue that different tubes of blood are different repertoires, and a user may choose to combine those into a single "repertoire" bc it makes sense for their study.

schristley:
> This seems to me like something that should probably live in user-land, rather than build the structures necessary to model this into the general use case. I would argue that different tubes of blood are different repertoires, and a user may choose to combine those into a single "repertoire" bc it makes sense for their study.

A repertoire is defined by the study designer, we are purposefully avoiding defining what a repertoire must be and then requiring users to conform to it. Instead, we are creating a structure that support as many different ways as we can think of for how a repertoire might be defined.

laserson:
Our goal is to very specifically define one version of "repertoire" that is broadly useful (and ideally easy to use). My general default is to make the data structure simpler if possible.

In your example, presumably you'll forever propagate that some rearrangements came from one blood draw while another came from a different blood draw, which is essentially saying you have two different blobs of metadata with different global IDs that have some special relationship to each other. Why not generate two different repertoire objects that replicate all the metadata? Or if the two datasets are truly mergeable, just eliminate the fact that they come from different blood draws entirely for downstream analyses and use one repertoire object?

javh:
> Why not generate two different repertoire objects that replicate all the metadata? Or if the two datasets are truly mergeable, just eliminate the fact that they come from different blood draws entirely for downstream analyses and use one repertoire object?

Correct me if I'm wrong, but this sounds pretty similar to having `Repertoire` be a set of refs to `RearrangementSet` objects and each `RearrangementSet` containing a single (`Sample`, `CellProcessing`, `NucleicAcidProcessing`, `SequencingRun`, `SoftwareProcessing`) tuple.

schristley:
> Correct me if I'm wrong, but this sounds pretty similar to having Repertoire contain an array of refs to RearrangementSet objects and each RearrangementSet containing a single (Sample, CellProcessing, NucleicAcidProcessing, SequencingRun, SoftwareProcessing) tuple.

That is one way to structure the data, though it changes the semantics of RearrangementSet to be something else (not sure exactly what). The initial semantics I had for RearrangementSet is meant to cover the process of taking the raw sequence data for a `repertoire_id` and annotating them to generate rearrangements, so it's a link from the SequencingRun objects to the Rearrangement objects, with SoftwareProcessing as annotation on that "link".

nishmm:
> Correct me if I'm wrong, but this sounds pretty similar to having `Repertoire` contain an array of refs to `RearrangementSet` objects and each `RearrangementSet` containing a single (`Sample`, `CellProcessing`, `NucleicAcidProcessing`, `SequencingRun`, `SoftwareProcessing`) tuple.

This makes more sense. +1. Either go from sample metadata to many rearrangement sets (due to differences in software processing, runs ,replicates etc.) or as described above go backwards i.e. each rearrangement set as a unit that has the metadata associated with it.


> That is one way to structure the data, though it changes the semantics of RearrangementSet to be something else (not sure exactly what). The initial semantics I had for RearrangementSet is meant to cover the process of taking the raw sequence data for a `repertoire_id` and annotating them to generate rearrangements, so it's a link from the SequencingRun objects to the Rearrangement objects, with SoftwareProcessing as annotation on that "link", and thus doesn't have Sample, CellProcessing, or NucleicAcidProcessing info within it.

I am as convinced as during the call that 'repertoire_id' is still not defined properly as to what it is being referred to. Is it the biomaterial in the test tube or the initial "raw" data that comes out of the sequencing run? With respect to referring to the test tube. Not sure if it is being referred to the final sample aliquot that gets loaded into a sequencing run or a sample at a higher level (which could be split into many ways, technical replicates or new child samples with different processing).

schristley:
> I am as convinced as during the call that 'repertoire_id' is still not defined properly as to what it is being referred to. Is it the biomaterial in the test tube or the initial "raw" data that comes out of the sequencing run?

It's neither. A "repertoire" is an abstract organizational unit of analysis that is defined by the researcher and consists of a set of rearrangements, study metadata, subject metadata, sample metadata, cell processing metadata, nucleic acid processing metadata, sequencing run metadata, a set of raw sequence files, and data processing metadata. A "repertoire" gathers all of this information together into a composite object. We desire a "repertoire" to be FAIR (findable, accessible, interoperable and reusable)

* findable: by giving each repertoire a globally unique `repertoire_id`
* accessible: by providing a CRWG API where repertoires can be retrieved
* interoperable: by defining a OpenAPI schema for the repertoire object
* reusable: by linking all of the metadata and rearrangements together into a standard format

The main constraint is that a repertoire is specific to a single subject otherwise it can consist of any number of samples (which can be processed in different ways), any number of raw sequence files, and any number of rearrangements. It can also consist of any number of data processing metadata objects that describe the processing of raw sequence files into rearrangements.

Note that another researcher can take the same raw sequencing data and associated metadata and create their own "repertoire" that is different from the original researcher's. A common example is to define a repertoire that is a subset such as "productive rearrangements for IGHV4" whereas the original researcher defined a more generic "B cell repertoire". This new "repertoire" would have much of the same metadata as the original "repertoire", except associated with a different study, and with additional information in the data processing metadata that describes how the rearrangements were filtered down to just the "productive rearrangements for IGHV4".

The reason a "repertoire" is defined as an abstraction instead of referring to a single well-defined physical entity is because not everybody agrees on that single physical entity.

laserson:
> A "repertoire" is an abstract organizational unit of analysis that is defined by the researcher. [...] The reason a "repertoire" is defined as an abstraction instead of referring to a single well-defined physical entity is because not everybody agrees on that single physical entity.

This makes me think it should be private to the researcher then, no? This concept doesn't exist in NCBI for example. Put another way, perhaps a "repertoire" should basically just be a list of identifiers. Each of those identifiers points to an object (maybe this is the "RearrangementSet") that has a single instantiation of all the contained objects and is the "rigorous" object that we really specify.

schristley:
> This makes me think it should be private to the researcher then, no?

Not exactly, as for scientific reproducibility, you want to know how the researcher defined them so you can repeat the experiment.

> This concept doesn't exist in NCBI for example.

That's correct. BioSample is close.

> Put another way, perhaps a "repertoire" should basically just be a list of identifiers. Each of those identifiers points to an object (maybe this is the "RearrangementSet") that has a single instantiation of all the contained objects and is the "rigorous" object that we really specify.

That's essentially what the "normal" [example](https://github.com/airr-community/airr-standards/blob/CRWG-API/lang/python/examples/florian.normal.yaml#L1282) and "denormalized" [example](https://github.com/airr-community/airr-standards/blob/CRWG-API/lang/python/examples/florian.airr.yaml) are, they are a list of identifiers, with different choices on how to structure/format the data associated with those identifiers.

schristley:
> Yeah, I can see how depth could be an issue. I'm not sure what to do about, because this structure is setup to have Sample as an array, and each Sample can have a different chain of (CellProcessing, NucleicAcidProcessing, SequencingRun, SoftwareProcessing), so it's inherently deep.

@javh I'm fine with this level of depth...

```
Repertoire:
    discriminator: AIRR
    type: object
    properties:
        repertoire_id:
            type: string
            description: Identifier for the repertoire object.
        study:
            $ref: '#/Study'
        subject:
            $ref: '#/Subject'
        diagnosis:
            type: array
            items:
               $ref: '#/Diagnosis'
        sample:
            type: array
            items:
                allOf:
                    - $ref: '#/Sample'
                    - cell_processing:
                        $ref: '#/CellProcessing'
                    - nucleic_acid_processing:
                        $ref: '#/NucleicAcidProcessing'
                    - sequencing_run:
                        $ref: '#/SequencingRun'
```

laserson:
I'm ok to merge as is and perform and necessary fixup in followup (smaller) commits/PRs.

PR #167, opened 2018-11-30
R tests

ssnn-airr:
Ppdated the coordinates in the example files from 0 to 1


javh:
Do we also need to change the tests themselves in `test/testthat/test-Interface.R`? They are currently set to load the data with `base="0"`.

ssnn-airr:
Yes, thanks, for consistency. It doesn't really make a difference, because we don't actually test if the file is 0 or 1 based. I changed to 1 because I noticed the files for python had been changed to use 1. It would be good if we used the exact same files for testing.

Issue #96, opened 2018-03-28
NULL-able MiAIRR fields


bussec:
A student in our group has recently started to work out MiAIRR-compliant data exports from our sciReptor databases, which exclusively hold single-cell data. During this process we came across a couple of points, which I will address in this an two additional tickets (#97, #98).

* Are we fine with just a NULL value or would we like to also have a more explict "not applicable" or "not determined" value?
* We already discussed that some fields must not be NULL. How do we store this information in the YAML?
* Should we set all fields to NOT NULL first and then do opt-outs once it becomes clear that the data for this field might not be available for legitimate reasons?

schristley:
Listing a field as `required` is meant to prevent it from having a NULL value, however for strings, the value can still be blank `""`. Note, it is up to any tools and/or GUI to enforce that rule.

I don't think a "not applicable" value offers much info but others like "not collected" or "missing" offer a slight nuance, i.e. we collected this information for some samples but couldn't get it for all of them. If all samples in a study are marked "not collected" then that's not helpful, better to leave the field blank to save space. Personally I don't like using those values because when doing a query on a repository you will need to specify multiple combinations to get the data you want.

javh:
If a field is not required, and is `null`, then I think it should be excluded from the file.  

I think `null` is most appropriate for required missing data. The meaning is pretty universal. `null` in yaml, `NULL` in R, and `None` in Python all mean undefined (no value).

BusseChristian:
Speaking with my MiAIRR hat on: MiAIRR fields are always `required` but some can be `NULL` (data that is simply not available in certain setups but crucial in others). Otherwise we would need to find a new terms for these levels, but IMO `required` transports the message "you have to actively report this field, even if your dog ate the USB stick with the data on it" pretty well.
So could we agree on the following three levels?
* required / not NULL (=> reporting required)
* required / NULL-able (=> reporting required)
* not required / not NULL (=> reporting *not* required, do not report if NULL)

bcorrie:
WRT not collected/missing, the problem here is that this will be reported in various ways. If we wanted to support that, then the spec should probably specify what the possible values are and what they mean. We don't want "not collected" and "missing" to be in two files from two tools and meaning the same thing. As Scott says, things can get messy quickly, especially when trying to make queries or comparisons that capture these edge cases.

[...]

Issue #157, opened 2018-09-25
data model and format for germline annotation sets


schristley:
Based upon the schema William and others initially devised, formalize the schema with the OpenAPI spec and integrate with the other schema and reference libraries. The actual germline annotation sets with data for various species are stored in another AIRR [repository](https://github.com/airr-community/germline-sets)

- [x] formalize schema
- [x] formalize file format (current yaml/json).
- [ ] link repertoire/rearrangement to germline. The link is currently at the rearrangement level but might want to reconsider this, maybe at the repertoire level with software processing and annotation.
- [ ] provide API in reference library to load germline annotation sets from file and perform basic operations like gene lookup and retrieve gene annotations.
- [ ] provide tool to generate IgBlast internal tables with appropriate annotations (this is if IgBlast doesn't want to read the germline annotation file directly).
- [ ] provide tool that performs a "diff" between two germline annotation sets.
- [ ] Should we provide a tool to generate germline sets from IMGT libraries?
- [ ] documentation, examples, etc.

A nice little example use case would be to take an existing AIRR TSV file, specify a different germline annotations set (e.g. one with inferred alleles in it), rerun IgBlast to generate new AIRR TSV with the different germline, then perform a simple analysis comparison, like V gene allele usage, between the before/after data.

williamdlees:
Looks good, Scott.

Should we provide a tool to generate germline sets from IMGT libraries?

schristley:
> Should we provide a tool to generate germline sets from IMGT libraries?

That's a good question. There are two points to that. First are the gene sequences and annotations, and the second is combining a set of those genes into a germline set. Right now my belief (which could be incorrect) is that IMGT is just a set of unstructured sequences, mouse is a good example, the genes for all the different strains are mixed together, so IMGT doesn't really define germline sets as I'm thinking about them. Thus my initial thought is no, and the germline sets will likely be hand-curated for the moment.

I see these annotations sets as living side-by-side with the IMGT sequences. That is, people will still need to download the IMGT gapped (and ungapped) FASTA sequences for tools that need access to the sequence, while AIRR annotations are provided which are not easily extracted from LIGM-DB or other sources.

javh:
> Right now my belief (which could be incorrect) is that IMGT is just a set of unstructured sequences, mouse is a good example, the genes for all the different strains are mixed together, so IMGT doesn't really define germline sets as I'm thinking about them. 

[...]

PR #182, opened 2019-03-06
MiAIRR update round one / Merge MiAIRR keys to AIRR schema

bussec:
Addressing #29 and #96. There are more changes coming but as there are already a couple of questions on my list I already wanted to create this PR now so that I can address them on-the-fly.

1. @schristley: The `x-miairr-format` key introduces so redundancy with both controlled vocabularies (implemented as `enum`, e.g. for `sex`) and ontologies (as you implemented it for `organism`). Ultimately we would like to have a single structure, what is your take on this?
1. All MiAIRR keys have an `x-miairr-example` field. However, I saw that some keys also have examples in their `description`. Should `x-miairr-example` become a general `example` holding this information?
1. The AIRR schema does not have a `raw_data` key, which holds information required to populated the fields for the "Raw data" column in an autogenerated TSV. I put this information in as a comment (line 819--829). Any ideas where this could go?
1. Why are there two `germline_database` keys (one in `Alignment` and one in `Rearrangement`) in the schema?
1. The `x-miairr-set` and `x-miairr-subset` keys are rather redundant at the current level of the hierarchy. Would it be possible to move them up one level (so that e.g. `Subject` would have them as direct child nodes)?


schristley:
Thanks @bussec , I was looking at the OpenAPI spec and extension properties can be any type, meaning they could be basic types like strings and number, but they can also be objects or arrays. I'm wondering if the schema would look less cluttered if we make these MiAIRR attributes into an object. It still uses the same amount of space, but encapsulates those attributes like this:

```
x-miairr:
    required: true
    nullable: true
    set: 1
    subset: study
    name: Study
    format: free text
    example: PRJNA001
```

What do you think? This would also allow us to define the "schema" for this object like with others:

```
MiAIRR_Attributes:
    discriminator: AIRR
    type: object
    properties:
        required:
            type: boolean
        nullable:
            type: boolean
```

which would be useful for doing validation, or at least be able to describe what these fields are with a data type, description and etc.

bcorrie:
@bussec , there was discussion around germline_database here:
https://github.com/airr-community/airr-standards/issues/157#issuecomment-451604385
https://github.com/airr-community/airr-standards/pull/156#issuecomment-447919836
I think the consensus was this probably doesn't belong at the rearrangement level because it would be identical for many rearrangements and that it belonged at the metadata level somewhere. 

But I don't think we came to a resolution on this... Which is why it probably exists in both places.

bussec:
@bcorrie, thanks for the references. IMO `germline_database` should be located in the `Repertoire` and/or `Rearrangement_set` object. For this PR, I will leave this as they are, i.e. keep both `germline_database` keys in their current place and their content in sync. However, I have opened a new issue (#183) to address this issue.

schristley:
> 2. All MiAIRR keys have an x-miairr-example field. However, I saw that some keys also have examples in their description. Should x-miairr-example become a general example holding this information?

The OpenAPI spec has a formal [examples object](url
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#example-object) which can be used to give a more detailed examples. We should probably use that.

> 4. Why are there two germline_database keys (one in Alignment and one in Rearrangement) in the schema?

The Alignment schema is experimental. It was meant to be a hit table output for IgBlast, however IgBlast has indicated it is difficult for them to output a hit table and instead they are directly producing the AIRR TSV output. I think we can deprecate the Alignment schema.

schristley:
> 1. The x-miairr-format key introduces so redundancy with both controlled vocabularies (implemented as enum, e.g. for sex) and ontologies (as you implemented it for organism). Ultimately we would like to have a single structure, what is your take on this?

`x-miairr-format` seems redundant in a number of ways, e.g. if the field is boolean type then there is no need to specific TRUE/FALSE, also if the field is string type with no ontology then that implies "free text", and so on.

I think we should define either an ontology or an enum for a field, but not both. If we have these properly defined then there probably isn't a need for `x-miairr-format`

schristley:
@bussec 

> 3. The AIRR schema does not have a raw_data key, which holds information required to populated the fields for the "Raw data" column in an autogenerated TSV. I put this information in as a comment (line 819--829). Any ideas where this could go?

When we split out the SequencingRun object, I added some non-MiAIRR fields to try to capture the raw sequencing data. Note that `raw_data` is somewhat ambiguous as it suggests the field holds the "raw reads", which isn't really true, instead that field holds a filename for the FASTQ file which contains the raw reads. However, one field isn't really sufficient to cover all of the possibilities. I also wanted to handle the case where FASTQ files might not be available but FASTA is, so I took the fields that we use in VDJServer and added them in.

Should we deprecate `raw_data` for these more detailed fields, or keep `raw_data` and have it store "summary" information? It's important that programs can access this information programmatically because they might want to read and process these files.

```
        file_type:
            type: string
            enum:
                - fasta
                - fastq
        filename:
            type: string
        read_direction:
            type: string
            enum:
                - forward
                - reverse
                - mixed
        paired_filename:
            type: string
        paired_read_direction
            type: string
            enum:
                - forward
                - reverse
                - mixed
```

schristley:
@bussec @bcorrie 
I've pushed a commit to the branch which adds a `MiAIRR_Attributes` and an `Ontology` schema definition. I've also modified `study_description` and `organism` as examples.

I'm not sure if a standard `Ontology` schema is needed, as it is just two fields, or whether we should code the properties directly like how `organism` was before?

schristley:
One disadvantage to defining an `Ontology` schema is that some standard parsers, like pyyaml, do not automatically follow `$ref` to fill out the schema. We will need to rewrite code in various places to support this. This situation already applies to `Repertoire` which references all of its sub-objects using `$ref`.

schristley:
@bussec Now I'm thinking we should rename `x-miairr` to just `x-airr` as it will likely contain attributes beyond just those specific to MiAIRR. We can keep a `miairr` boolean within the `x-airr` object that signifies whether the field is a MiAIRR field.

schristley:
@bussec I've restructured, hopefully I didn't make a mistake with all the cut/pasting, please review.

The autogeneration of TSV and integration with docs will need to wait for the next PR because the docs are undergoing major changes at the moment in another branch.

schristley:
@javh look okay?

javh:
I'll make an effort to review before Monday.  There are also several issues I haven't read through yet. I've been out this week, so I haven't had a chance to catch up yet. I will Soon(TM).

bussec:
Nice work, @schristley! Trying to address some of your points below:

1. `Ontology` object: Yes, I think it is reasonable to introduce this object at this point, as the number of ontology-controlled fields is expected to grow.
2. "raw data" field: I am ok with replacing this field with a set of more specific keywords as long as it is obvious that they belong to MiAIRR set 4.
3. `x-miairr` to `x-airr`: Fine with me.

bussec:
@schristley, some more questions:
1. What should the `url` within the `Ontology` object point to? Currently these refer to a human-usable web interfaces, would an API endpoint be preferable?
2. Is the `physical quantity` format predefined in OpenAPI or do we still need to specific this?
3. Is there anything left from your side before we merge?

schristley:
> "raw data" field: I am ok with replacing this field with a set of more specific keywords as long as it is obvious that they belong to MiAIRR set 4.

Okay, I think I will go back to a slightly earlier version of the schema for this when I had the "raw files" separated out as its own object, that will make it more obvious that they belong to MiAIRR set 4. Right now I have it together with SequencingRun.

schristley:
> 1. What should the url within the Ontology object point to? Currently these refer to a human-usable web interfaces, would an API endpoint be preferable?

Probably because it will likely be used more by computers than humans.

> 2. Is the physical quantity format predefined in OpenAPI or do we still need to specific this?

Unfortunately not, the [formats](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-typesl) in OpenAPI deal more with lower level representation, like 32 vs 64 bits, instead of higher level formats. They specify a date format but that's about it.

schristley:
Separating the raw sequencing files allows us to be a little more flexible in specifying those files.

> 3. Is there anything left from your side before we merge?

I think we are good!

javh:
This all looks good to me except for two things:

1. `AIRR_Attributes` uses a different style than all other top level items which are in CamelCase. 'AIRRAttributes'? or just 'Attributes'?
2. Schema consistency checks fail. Looks like it just needs propagation from the top level into the python library.

PR #200, opened 2019-04-17
Introduce "keywords_*" fields

### LINKED ISSUE #169: Heavy and light chain at the metadata level?
Hello All,
I believe we have a mechanism for pairing heavy and light chain at the rearrangement level, this is the cell_id field. If two rearrangements with a repertoire are a paired heavy and light chain they should have the same cell_id (correct me if I am wrong).

My questions is more at the metadata/repertoire level. We have use cases where people are only interested in studies that have paired heavy/light chains (or alpha/beta). Is there a place in MiAIRR where it is noted whether or not a repertoire has this feature? It appears that the closest thing I can find is:

3 / process (nucl. acid)	Physical linkage of different loci	string	Controlled vocabulary	Describes the mode of linkage if a method was used which physically links nucleic acids derived from distinct loci in a single-cell context	IGH-IGK/IGL-head/head	physical_linkage

If this is the case, what would one expect to find in this value for the cases where:
1) There are no paired heavy/light chain data
2) There is paired data in the repertoire and it is paired (I am not sure I understand the meaning of the example given "IGH-IGK/IGL-head/head"

### LINKED ISSUE #174: Is there a MiAIRR field that differentiates T-cell and B-cell?
Hi All,
A quick question. Is there a single field in MiAIRR that would differentiate a repertoire as to whether it was a B-cell or T-cell specific repertoire? There are many ways to infer it, but is there a field that states this explicitly. Nothing is jumping out at me.

Brian

### LINKED ISSUE #185: Which terms do we need as MiAIRR keywords
As discussed in #174, introduction of one of more `keyword` field in MiAIRR will increase the findability of data sets and reduce abuse of fields intended for experimental annotation. As discussed in today's MiniStd call:
* The `keyword` fields will hold a comma separated list from a predefined vocabulary, free-text will not be allowed.
* We will start with a small and flat list of terms. Over time this is expected to evolve and include some hierarchies.
* As it is unlikely that there is a single ontology to provide us with all terms, we can potentially cross-link/map our terms to other ontologies.

For a start, we are now collecting suggestion for 10-15 top-level terms that would be frequently used (please focus on content for now):

* `contains_ig`
* `contains_tcr`
* `contains_single_cell`
* `contains_paired_chain`

Pinging @bcorrie @emilyvcbarr @lgcowell @schristley

bussec:
closes #185
closes #174
closes #169

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -229,8 +229,8 @@ Study:
                     - contains_paired_chain
             description: Keywords describing properties of one or more data sets in a study
             example:
-	      - contains_ig
-	      - contains_paired_chain
+              - contains_ig
+              - contains_paired_chain
             x-airr:
                 miairr: true
                 required: true
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -218,6 +218,24 @@ Study:
                 set: 1
                 subset: study
                 name: Relevant publications
+        keywords_study:
+            type: array
+            items:
+                type: string
+                enum:
+                    - contains_ig
+                    - contains_tcr
+                    - contains_single_cell
+                    - contains_paired_chain
+            description: Keywords describing properties of one or more data sets in a study
+            example: "[ 'contains_ig', 'contains_paired_chain' ]"
+            x-airr:
+                miairr: true
+                required: true
+                nullable: true
+                set: 1
+                subset: study
+                name: Keywords for study
 
 # 1-to-n relationship between a study and its subjects
 # subject_id is unique within a study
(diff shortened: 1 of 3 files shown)


schristley:
#174 wants to determine status for a repertoire, while these keywords are at the study level. @bcorrie is this sufficient for your requirements? Likewise for #169 where I think we also need to add `pair_id` at the rearrangement level.

bussec:
@schristley I agree that we need additional `keyword` fields on lower levels and guideline how the terms "trickle down" (or "float up") between these levels. Is it already clear that we need one the repertoire level (if that is still the correct term)?

bussec:
@schristley is it ok to proceed and merge? Or is there anything that need to be clarified?

schristley:
@bussec looks good!

PR #211, opened 2019-06-28
MiAIRR single-cell extension (scXT)

### LINKED ISSUE #206: Single-cell expression data annotation/linkage
One of the goals of MiniStd WG is to extended the current MiAIRR specificiation to provide mean for more detailed annotation on the single-cell level. We already discussed chain-pairing in #169 (now in PR #200). In addition to this, we also aim to provide marker expression and reactity data. However, the corresponding raw data sets usually contain data from a lot of events and partitioning can be complex. Therefore in analogy to the MiAIRR Set 6 annotation, two levels will be provided: First, a link to the raw data set along with a pointer to the record(s) in question. Second, the most relevant parts of the processed data to allow for fast search and compare operations.

This issue will only deals with marker expression, for single-cell reactivity data see #207.

The proposal is to add the following information to the cell-level of the schema:
* single-cell expression data:
  * keyword describing the methology used to assess expression. Controlled Vocabulary. Currently: {"flow cytometry", "single-cell transcriptome"}.
  * Reference to raw data
    * DOI of raw data set containing the current event
    * Index addressing the current event within the raw data set. Note that the index field is not necessarily numeric but can also hold a DNA barcode.
  * _n_ processed data records:
    * `value`: transformed and normalized expression level
    * `marker`: standardized designation of the transcript or epitope

Potential issues of the approach:
* `value`: Transformation is a not standardized procedure, especially for FC data. Also note that no gating information is provided as this often goes FUBAR after transformation.

* `marker`: To be interoperatable, this should come from an ontology. While this is simple for gene expression, flow cytometry markers are more difficult as there is BTMK no ontology for them. Also note that gene name != protein name != FC marker.

### LINKED ISSUE #207: Single-cell reactivity data annotation/linkage
For general background, see #206. The aim is to provide reactivity information for individual Ig/TCRs. Although this will often occur in a single-cell context, attaching the information to the `paired_chain_id` (and not to `cell_id`) is important as individual cells can express multiple receptors, e.g. due to simultaneous expression of two TCR alpha chains. Referencing to the reactivity record via `paired_chain_id` resolves this issue and at the same time allow the annotation of stochastically inferred chain pairs, which inherently lack a `cell_id`.

The proposal is to add the following record structure to the AIRR schema, so that it can be referenced by the paired-chain identifier.

* single-cell reactivity data:
  * keyword describing the methodology used for measurement. Note that only methods that return an absolute and quantitative measurement can be used (affinity, binding energy, etc.). Examples: "SPR", "ITC".
  * UID of the receptor at IEDB, which can hold additional and richer data records.
  * DOI of raw data measurements
  * _n_ records of reactivity measurement
    * `antigen`: The antigen against which the receptor was tested. For TCRs this is the combination of MHC and peptide.
    * `value`: The absolute (processed) value of the measurement.
    * `unit`: The physical unit of the measurement, e.g. M or kJ mol-1

Potential issues:
* `antigen`: To allow meaningful comparisons across studies, this field must be populated from a controlled vocabulary, IEDB would be the obvious candidate for this, but need to check whether there are any potential restrictions in terms of molecule classes.

bussec:
closes #206 
closes #207

### DIFF
--- a/docs/miairr/miairr_single_cell_extension.rst
+++ b/docs/miairr/miairr_single_cell_extension.rst
@@ -33,15 +33,17 @@ Therefore the following proposed extension to the MiAIRR data standard
 have been brough forward to the AIRR Community General Assembly in
 May 2019 and were ratified as provisional standard:
 
--  Add the following record structure to the AIRR schema, so that it can
-   be referenced via ``pair_id``.
-     -  keyword describing the methodology used for measurement. Note that only methods that return an absolute and quantitative measurement can be used (affinity, binding energy, etc.). Examples: "SPR", "ITC".
-     -  UID of the receptor at IEDB, which can hold additional and richer data records.
-     -  DOI of raw data measurements
-     -  n records of reactivity measurement
-        -  ``antigen``: The antigen against which the receptor was tested. For TCRs this is the combination of MHC and peptide.
-        -  ``reactivity_value``: The absolute (processed) value of the measurement.
-        -  ``reactivity_unit``: The physical unit of the measurement, e.g. M or kJ mol-1
+- Add the following record structure to the AIRR schema, so that it can
+  be referenced via ``pair_id``.
+     - keyword describing the methodology used for measurement. Note that only methods that return an absolute and quantitative measurement can be used (affinity, binding energy, etc.). Examples: "SPR", "ITC".
+     - UID of the receptor at IEDB, which can hold additional and richer data records.
+     - DOI of raw data measurements
+     - n records of reactivity measurement
+          - ``antigen``: The antigen against which the receptor was tested. For TCRs this is the combination of MHC and peptide.
+          - ``reactivity_value``: The absolute (processed) value of the measurement.
+          - ``reactivity_unit``: The physical unit of the measurement, e.g. M or kJ mol-1
+          
+
 *  Add the following record structure to the AIRR schema, so that it can
    be referenced via ``cell_id``.
    *  keyword describing the methology used to assess expression.
--- a/docs/miairr/miairr_single_cell_extension.rst
+++ b/docs/miairr/miairr_single_cell_extension.rst
@@ -35,13 +35,13 @@ May 2019 and were ratified as provisional standard:
 
 *  Add the following record structure to the AIRR schema, so that it can
    be referenced via ``pair_id``.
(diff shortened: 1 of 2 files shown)


laserson:
This is supposed to go into the rearrangement file somehow?

bussec:
The two keys that we need in the general AIRR schema (`cell_id` and `pair_id`) are already in the `Rearrangement` object there, so no changes required there. We have not specified where the additional information will be stored, as I assume that this will be implementation dependent. Does this answer your question?

laserson:
I guess I'm not certain what kind of data you're talking about. If, for example, you're thinking of connecting this with 10x scRNA-seq, I'd rather think of how to incorporate our rearrangements data into the dataset structure being used in scRNA-seq, e.g., Loom format.

schristley:
Hi @bussec , the documentation you wrote looks fine to me. Moving forward it sounds like you want to go the full standards route, that is create schema objects, devise file formats, integrate that into the python, R and the ADC API software, etc. I think I'm comfortable with that for the reactivity measurements as that's clearly immune receptor related. I'm less comfortable with the expression measurements as this brings in the whole world of single cell RNA-seq (if I'm understanding correctly? and flow sorting) and trying to collapse all that complexity in protocols and methods down into a list of values and markers seems problematic. Maybe it's better if we just worry about how to perform cross-referencing between cells in rearrangements data and external data sources/formats?

It's also reasonable that `Rearrangement` is not the appropriate object level. In particular because there can be multiple rearrangements for a cell. Maybe `Cell` should be a schema object, and thus one record per cell, and links at the cell level would reside there and not duplicated throughout the rearrangement data.

There could be a similar discussion about that for clones, early discussions of the API thought about querying on clones which is a simpler abstraction versus querying all the rearrangement data and pull out the clones.

schristley:
@bussec @javh Can we merge this, it looks like only documentation that discusses a proposed extension, or is it obsolete?

javh:
Isn't this for MiAIRR v2? If so, we don't want to merge it into master until we are ready to do the v2.0 release.

bussec:
@javh @schristley Yes, this is v2.0 and we should resolve it along with #320 to avoid confusion. I have assigned it to the respective milestone.

javh:
Current API testing. Will be revisited after testing is finished.

bussec:
The original issues are/will be addressed in #574  and #404. As there have been a bunch of conceptual changes, the documentation needs to be rewritten, will open a separate issue for this.

PR #218, opened 2019-07-30
library support for repertoire metadata

### LINKED ISSUE #176: what capabilities to add to reference library for study metadata?
While we have been mainly discussing the structure and format of study metadata, we should also decide what capabilities we want to add to the reference library. Post ideas and we can discuss implementation. Here are some ideas I have:

- read/write metadata file
- merge metadata files, e.g. when merging rearrangement files
- validate metadata, i.e. basic schema checks like with rearrangements
- "attach" metadata to rearrangements? One idea is to pass a parameter to `read_rearrangement` that specifies the metadata file. Or don't worry about "attaching", users manage them separately. However, "attaching" should allow us to perform additional validation checks and functionality.
- How does code interact with a repertoire object? Unlike rearrangements which is just a list of fields together in a dictionary, the metadata is a nested structure. Does code traverse that structure directly, or do we provide an abstraction layer to hide it?

Here, for example, is one implementation issue. The `repertoire_ids` are not indexed in the metadata file, they are just a list, so looking up a `repertoire_id` would naively requiring iterating through the list to find it, which is expensive if doing that for every rearrangement record. Alternatively, the list is converted into a dictionary keyed by the `repertoire_id` to allow for quick lookup. We could consider the same thing for `study_id`, `subject_id`, etc.

schristley:
fixes #176 

Not ready to be merged yet but PR serves as discussion point.

Right now the read/write/validate routines are file-oriented. However, one of the use cases is to retrieve data from a data repository through the ADC API. Depending upon how the web request is performed, the data either comes back as JSON or as a stringified JSON. I’d like to be able to use RepertoireReader and RepertoireWriter (which isn’t written yet) to read the data, be able to validate it, and/or write it out to a file. Now load_repertoire() is defined to take a file name, and RepertoireReader is defined to take a handle, so what is the pythonic way to get the data from an existing JSON object?

You can look at the example code I wrote here:

https://github.com/airr-community/airr-standards/blob/metadata-docs/docs/examples/api/retrieve_data.py

That example just uses the returned JSON from API directly, but I'd want to change it to use the reference library to perform a validation of the returned repertoires as well as write out to a file.

I can imagine adding another parameter to load_repertoire() where the JSON object is passed in. Or maybe there is a load_repertoire_from_json() function as a separate interface, but RepertoireReader can take either a handle or a JSON object?
 
Now the repertoire metadata is a little different from the rearrangements in that with rearrangements, we stream the data in. Because the repertoire data is not expected to be crazy large, the load routine reads in the whole file. That means the write operation is a little different too, unlike rearrangements where you write one at a time, RepertoireWriter would want all of the repertoires so it can write them out to a file all at once.


schristley:
@bcorrie I've added the validation routine for repertoire metadata, and you can also use the `airr-tools` command line program.

schristley:
@javh Hey Jason, can you review this? The update to the python library docs is actually on the metadata-docs branch.

javh:
Yes, I can. Sorry for the slow.  I'm going to try to partition off a chunk of time on Friday to take care of some AIRR stuff, including this.

lang/python/airr/interface.py
```diff
@@ -196,3 +200,148 @@ def validate_rearrangement(filename, debug=False):
...
+      dictionary of repertoire objects.
+    """
+
+    # Because the repertoires are read in completely, we do not bother
+    # with a reader class.
+    md = None
+
+    # determine file type from extension and use appropriate loader
+    ext = filename.split('.')[-1]
+    if ext == 'yaml' or ext == 'yml':
```
javh: if ext in ('yaml', 'yml'):

lang/python/airr/interface.py
```diff
@@ -6,10 +6,14 @@
 import pandas as pd
 from collections import OrderedDict
 from itertools import chain
+from pkg_resources import resource_filename
+import json
+import yaml
+import yamlordereddictloader
```
javh: I can't recall why right now, but I had issues with `yamlordereddictloader` in recent PyYAML versions (needed to solve a warning).  What I did instead was:

```
# Set YAML loader to OrderedDict
from collections import OrderedDict
def dict_representer(dumper, data):  return dumper.represent_dict(data.iteritems())
def dict_constructor(loader, node):  return OrderedDict(loader.construct_pairs(node))
yaml.add_representer(OrderedDict, dict_representer)
yaml.add_constructor(yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG, dict_constructor)
```

Then  you can load the yaml with the FullLoader and it'll be ordered:

```
yaml.load(handle, Loader=yaml.FullLoader)
```

lang/python/airr/interface.py
```diff
@@ -196,3 +200,148 @@ def validate_rearrangement(filename, debug=False):
...
+
+    # Because the repertoires are read in completely, we do not bother
+    # with a reader class.
+    md = None
+
+    # determine file type from extension and use appropriate loader
+    ext = filename.split('.')[-1]
+    if ext == 'yaml' or ext == 'yml':
+        with open(filename, 'r', encoding='utf-8') as handle:
+            md = yaml.load(handle, Loader=yamlordereddictloader.Loader)
```
javh: Could be replaced with:

``` 
md = yaml.load(handle, Loader=yaml.FullLoader)
```

If above is done instead of importing `yamlordereddictloader`.

lang/python/airr/interface.py
```diff
@@ -196,3 +200,148 @@ def validate_rearrangement(filename, debug=False):
...
+                sys.stderr.write('%s has repertoire at array position %i with validation error: %s\n' % (filename, i, e))
+
+    return valid
+
+def write_repertoire(filename, repertoires, info=None, debug=False):
+    """
+    Write an AIRR repertoire metadata file
+
+    Arguments:
+      file (str): path to the output file.
```
javh: filename (str):

lang/python/airr/interface.py
```diff
@@ -196,3 +200,148 @@ def validate_rearrangement(filename, debug=False):
...
+      repertoires (list): array of repertoire objects.
+      info (object): info object to write. Will write current AIRR Schema info if not specified.
+      debug (bool): debug flag. If True print debugging information to standard error.
+
+    Returns:
+      bool: True if the file is written without error.
+    """
+    if not isinstance(repertoires, list):
```
javh: No objection to requiring it be a list, but does it need to be? Could it be any iterable instead (which would also include a list)? Eg,

```
try:
    iter(repertoires)
except TypeError:
    if debug:  sys.stderr.write('Repertoires parameter is not iterable\n')
```

javh:
Looks really good to me. I haven't pulled and tested the code yet, but I skimmed over it and made a few minor comments.

(If I can get to testing today I will, but I thought it better to approve sooner in case I get delayed.

schristley:
> Looks really good to me. I haven't pulled and tested the code yet, but I skimmed over it and made a few minor comments.

Great, I'll work through those. I was just thinking that I should add few tests too...

lang/python/airr/interface.py
schristley: changed

lang/python/airr/interface.py
schristley: I copied the code from `schema.py`. I'm using PyYaml 5.1.2 and it seems to be working just fine?

lang/python/airr/interface.py
schristley: fixed

lang/python/airr/interface.py
schristley: I'm using `yaml.dump` and `json.dump` to output the whole object. I suppose if it let's me assign the iterator to the metadata `md` object and it outputs to the file then I'm okay. If I actually have to iterate over the object to pull out the data, then it's less useful.

```
    md['Repertoire'] = repertoires
```

javh: Ah, I think I missed that the list was a single object (ie, `yaml.dump` instead of `yaml.dump_all`). Nevermind.

lang/python/airr/interface.py
javh: Okay, then let's not worry about it? If the problem (you know, that one I can't remember) pops up we can deal with it then.

PR #247, opened 2019-09-13
Fix link to MiAIRR-to-NCBI Submission Manual

marcosmro:
I've replaced .rst by .html to make the link work

### DIFF
--- a/docs/cairr/overview.rst
+++ b/docs/cairr/overview.rst
@@ -40,9 +40,8 @@ the metadata quality of AIRR-seq studies.
 Submission Steps
 ----------------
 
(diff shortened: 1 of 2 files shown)


schristley:
A better to technique is to use sphinx cross references versus directly linking to the output html.

In the `manual_miairr_ncbi.rst` file, put a label right before the section you want to link

```
.. _MiAIRR_NCBI_Submission_Manual_Option_1:
```

then in the `cairr.rst` file you can reference it with

```
:ref:`MiAIRR_NCBI_Submission_Manual_Option_1`
```

marcosmro:
Thanks @schristley. Done. Please, check that the link works.

schristley:
thanks @marcosmro

PR #250, opened 2019-09-17
AIRR schema for Olmsted tool

eharkins:
As discussed during the AIRR call on 9/16/19, we are seeking feedback on the schema we have developed for [Olmsted](https://github.com/matsengrp/olmsted), a B-cell repertoire and clonal family tree explorer tool developed by me and @metasoarous in @matsen's group. 

As promised, [here is where we have example data](https://github.com/matsengrp/olmsted/tree/master/example_data) both broken down into Olmsted parsable files (output from our script) and in the full schema json (input to our script).

**Some notes from the meeting** (numbered for reference, not by importance):
1. The AIRR rearrangment schema requires a `germline_alignment` field while our app relies on the `naive_seq` field in our schema, though we don't require it in our schema. Perhaps we can add `germline_alignment` to our schema if it would allow the schema to flex beyond our context, though the app would of course need to change to reflect whether a naive sequence or germline alignment was specified. 
2. Could we use an ICA or MRCA vs a naive in order to be more flexible? I think we want to stick with naive or some form of it for the purposes of this app.
4. Trees from other pipelines may exceed 100 seqs and blow the scale up. This is something to keep in mind but we are not at a point where we should be too concerned with it yet.
5. Having the sampled sequences in each tree representation is unnecessary duplication; we could move those sequences up a level to be at the clonal family level. Because inferred ancestral sequences will all be generally different across trees, this duplication is less explicit, though it is important to note as a potential step in helping us scale up.
6. Edge annotations are limited to the newick and to the node objects. Further edge annotations could go in the node objects so long as they only have one ancestral branch.
7. The AIRR schema has `clone_id`, whereas we have clonal family ids. Should our clonal family ids be unique across samples or used to track clones across samples?
8. We may want to separate clonal family objects into separate files or have them be on one line to save parsing when we scale.
9. Something I realized after the meeting: partis cdr3 fields refer to the cdr3 region in the naive sequence (@psathyrella correct me if I'm wrong), where as the cdr3 fields in the AIRR rearrangement schema seem to have to do (at least in some cases) with an individual query sequence. If this is accurate, we should make it clear in our schema that our fields refer only to the cdr3 region of the naive rearrangement, unless it seems like this will not be a point of confusion.

In general I have left field names in our schema and made notes in the `description` (search for `AIRR:` in the schema) about the potential analogous field in the AIRR schema where I was not positive of an identity between the two. For example, both schemas have fields denoting the start and end positions of the germline genes, but the indexing is different (see notes for those fields in the `olmsted-airr-schema.yaml`).

### DIFF
--- a/specs/olmsted-airr-schema.yaml
+++ b/specs/olmsted-airr-schema.yaml
@@ -1,0 +1,196 @@
+$id: https://olmstedviz.org/input.schema.json
+$schema: https://json-schema.org/draft-07/schema#
+description: 'Olmsted dataset input file. AIRR: see Study'
+properties:
+  build:
+    description: 'Information about how a dataset was built. AIRR: see DataProcessing'
+    properties:
+      commit: {description: Commit sha of whatever build system you used to process
+          the data, type: string}
+      time: {description: Time at which build was initiated, type: string}
+    required: [commit]
+    title: Build info
+    type: object
+  clonal_families:
+    description: Information about each of the clonal families
+    items:
+      description: Clonal family of sequences deriving from a particular reassortment
+        event
+      properties:
+        cdr3_start: {description: 'Start of the CDR3 region. From partis "zero-indexed
+            indel-reversed-sequence positions of the conserved cyst and tryp/phen
+            codons". AIRR: see cdr3_start, which excludes the conserved residue and
+            uses a "1-based closed interval"', minimum: 0, type: integer}
+        d_call: {description: D gene used in rearrangement., type: string}
+        d_end: {description: 'Position in d gene at which rearrangement ends. AIRR:
+            see d_germline_end, which uses 1-based closed interval as opposed to 0-based
+            python slice convention intervals used by partis.', minimum: 0, type: integer}
+        d_start: {description: 'Position in d gene at which rearrangement starts.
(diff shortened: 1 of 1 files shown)


schristley:
thanks @eharkins , to organize the work, I've created a `clones-and-trees` branch. This isn't quite ready to commit to `master` yet. If you could do me a favor and change the pull request to merge into that `clones-and-trees` branch, then we can merge and continue work on that branch.

schristley:
oh duh, I realized I could change that myself, if I only knew what buttons to click ;-D anyways @eharkins got there first, thanks.

psathyrella:
Yes, that's correct: partis views cdr3 length as a property of the rearrangement event, not of the individual mature sequences. So in cases with shm indels in the cdr3, the cdr3 length in the mature sequence would probably need to be calculated from the shm indel info (although I haven't thought through this recently, so I could be remembering something wrong). Also note that partis cdr3 length [includes](https://github.com/psathyrella/partis/blob/master/docs/output-formats.md) codons on both sides, which is different from the imgt definition, and I know that's horrible, but I swear it was because I was going off something else that had switched cdr3 and junction lengths, and there's just no way to change it now...

PR #254, opened 2019-09-25
Introduce age ranges 

### LINKED ISSUE #192: separate magnitude and unit for age?
There was earlier discussion in #27 about the age fields in MiAIRR. The actual `age` field is a physical quantity, which means it has both the magnitude and unit combined together into a text field. This makes it problematic to perform range queries on the age which is desirable for the CRWG API. Should we consider separating this field into two fields: `age` and `age_unit` where `age` is number and `age_unit` is an enum (yr, wk, day)?

franasa:
fixes #192 

First commit requires further review


schristley:
thanks @franasa , you will want to leave the `age` field in the schema but change `x-airr` to have `deprecated: true`

@bussec Any objection to using the full word, weeks, months, years, for `age_unit` instead of using abbreviations?

schristley:
@bcorrie @bussec think this is final.

schristley:
ontology it is, changes made...

Issue #255, opened 2019-09-26
Not clear how "Content format" is stored in the yaml file


franasa:
Hi all, as my first contribution I made this simple script to create a .tsv file from the airr-schema.yaml. It is not clear to me how the "Content format" column from AIRR_Minimal_Standard_Data_Elements.tsv was created, can anyone give me a hint?
I am also looking forward to any feedback (I hope I'm not breaking anything)

bussec:
Thanks @franasa ! There was a glitch in the position of the commit in the tree, so I just rebased it to `master` (instead of `issue-192`). Please fetch again from `origin`.

schristley:
Within `x-airr` there is a `format` field which has a controlled vocabulary defined [here](https://github.com/airr-community/airr-standards/blob/master/specs/airr-schema.yaml#L60). If the `format` field is not there, or is null then the full range of the data type is assumed, e.g. "free text" or "any number"

If it's `ontology` then there should be an `ontology` field in `x-airr` for the ontology. If it's `controlled_vocabulary` then the field should have an `enum` with the list of allowed values, and lastly `physical quantity` has a [physical quantity](https://en.wikipedia.org/wiki/Physical_quantity)

schristley:
thanks @franasa ! Likely what we will do is integrate your code into the docs, specifically `conf.py` which can generate the TSV when the docs are generated, which I think the design laid out in #29

javh:
Great, thanks @franasa.

You can see how the other tsv files are currently being generated from the yaml spec starting on [line 246](https://github.com/airr-community/airr-standards/blob/metadata-docs/docs/conf.py#L246) of `docs/conf.py` in `metadata-docs` branch. 

Just plug your code into that block after the other entries and add link to the new tsv in an appropriate place in the MiAIRR docs.

franasa:
thanks for your comments!, I just integrated some lines at the bottom of `docs/conf.py` including your comments. @schristley, in #192 data `type` is number but in the master it is integer, will it be changed to number, should i include this in the future? 
@javh the file is saved as `AIRR_Minimal_Standard_Data_Elements.tsv`  in the same `_downloads` folder where the previous files are generated, is this fine? I was not sure what you exactly meant with _add link to the new tsv in an appropriate place in the MiAIRR docs_. could you please elaborate on that?  thanks a lot!

schristley:
> in #192 data type is number but in the master it is integer, will it be changed to number, should i include this in the future?

[...]

PR #257, opened 2019-10-03
Resolve Issue 217

bcorrie:
Suggestions for resolving the description of v_call, d_call, and j_call as per Issue 217.

Use this pull request to discuss and finalize a solution...


javh:
Looks good to me. I just pushed some minor wording/format changes and added the additional text to `c_call`. I didn't update the TSV, in anticipation of #255 being merge in soon (but we could).

bcorrie:
Hi All,

I had added an extended description for d_call as including the extended naming as per #258. I had only added this to d_call as I wasn't sure we should have this much documentation.

Comments on the change below.

"6 / data (proc. seq.)	D gene	string	Free text	D segment gene and allele. If referring to a know gene in a reference database such as the IMGT/Gene DB, the relevant gene/allele nomenclature should be followed (e.g. IGHD2-2*01). If the gene/allele is not known, the name should include the locus (e.g. IGH), the domain (e.g. D), the gene subgroup (e.g. 2), and the gene subgroup designations (e.g. 2 delimeted with a -) and the allele (e.g. 01, delimted with a *) to construct the gene name  (e.g. IGHD2-2*01).	IGHD2-2*01	d_call"

https://github.com/airr-community/airr-standards/pull/257/commits/46e5e9f754d2f31c0e915764972a9baf595a9ceb

schristley:
> Should this be added to the other v/j/c call values? Or is this too much?

I don't agree with the additional text. It is suggesting/enforcing a nomenclature which we really aren't in the position of enforcing (i.e. IUIS controls this). Furthermore, for less documented species, the correct organization of the genes into groups and subgroups might not even been known. I think we want to avoid documenting any nomenclature ourselves and just reference the appropriate authorities for that.

Issue #139, opened 2018-05-16
Potential additions to the Rearrangement spec


javh:
From feedback on the manuscript, here are some additional fields to discuss for the next release of the Rearrangement schema:

- [ ] 1. A way to record the species of the V, J , D segments.
- [ ] 2. Clarity on if they using a global vs local alignments.
- [ ] 3. A place to enter the numbering scheme you are using. (It seems to assume IMGT, but doesn’t require it.)
- [ ] 4. The Deletion count in nucleotides from the V 3’, 5’ and 3’ D, and 5’J. (Although this could be in the Cigar string if you did a global alignment rather than a local alignment.)
- [ ] 5. An annotation for the Leader Sequence.
- [ ] 6. A way to annotate V-gene Conversion
- [x] 7. A way to annotate D-D recombination.

bussec:
@javh, my two cents on this:

- (1) Is this related to #137 or do you just need a `species` field that is independent of the MiAIRR record?
- (7) We currently have neither a MiAIRR nor an GenBank specification on for this. I assume this should be included there as well. Will this be restricted to any of the three possible loci (_IGH_, _TRB_, _TRD_)?

javh:
Sorry, these were copy-paste from an email conversation to make sure we discuss them. I neglected some details and my personal opinions.

1. Yes, I think this relates directly to #137. I think for the vast majority of use cases the existing MiAIRR `species` tag in the metadata should suffice, combined with the flexibility in gene call nomenclature. Ie, just not restricting people from using gene names like "Homo sapiens IGHV1-69*01".
2. I don't think we need to mandate this. Most aligners do semi-global, and some of the coordinate fields don't make a lot of sense outside that context, but they are optional. This is something we can just add to the technical metadata, such as the numbering scheme, software versions, alignment tool, etc.
3. I'm inclined to make it a *required* metadata field.
4. These can already be derived from `*_germline_start`, `*_germline_end` or the CIGAR strings if they include the tail `N` operators. However, that's all optional and this information is biological, so I think some optional fields similar to the N/P fields would be good additions. Eg, `v3_deletion`, etc.
5. This is encompassed be the V segment data, but I think there's good reasons to add optional leader fields to analyze them separately. We'll just need some more fields that start with `L` or `U` - I prefer `U` just because lowercase `L` is hard to distinguish from other characters.
6. I don't have enough experience with such data to know now to implement this, but I think we should do it. We need some example use cases.
[...]

PR #260, opened 2019-10-10
Add species information to `cell_subset` and `locus` fields

### LINKED ISSUE #137: How to indicate xenografted or transgenic animals
MiAIRR currently assumes that all entries in a record pertain to the same `species`, which is defined in data set 1. As `species` will be `required` (#96) and restricted by a species taxonomy (most likely NCBITAXON, #97) we need a mechanism to describe subjects that cannot be clearly assigned to a single species. Two examples of such a subject would be:

* Animals containing complete human immunoglobulin loci as transgenes. Note that it is not possible to unambiguously infer the species from `target_locus` as the gene symbol does often not contain this information. Therefore the sequence information could be wrongly interpreted as host sequences.

* Chimeric animals that have been grafted with human bone marrow (and potentially other tissues). Here also additional fields like `cell_subset` would be affected.

One solution would be to introduce a "species prefix" for the fields involved. However, as this changes MiAIRR, it a slightly bigger decision.

bussec:
fixes #137 

Also addresses #139 point 1


bussec:
Yes. I think this should be a general principle that the same field on lower levels overwrites the field on higher levels. The questions that come into mind:
1. Where should this be documented
2. Would it be smart to use high level field name as suffix (probably: yes)

schristley:
> Where should this be documented?

Adding to the `description` should be sufficient.

bussec:
`description` added, are we good to go?

schristley:
@bussec looks good, so we are sticking with `subject.organism` and not changing it?

bussec:
See my post in #266, we are changing the MiAIRR name, but it won't happen before early next year, as this is a major version step. Would you like to combine those changes or merge this one here for now and see how the decision turns out for the other one?

schristley:
Okay then, let's merge this now and leave the changes considered MiAIRR V2 for later.

bussec:
The `locus_species` part has now been moved into #614.

PR #261, opened 2019-10-10
autogenerate MiAIRR TSV from yaml schema

### LINKED ISSUE #29: Auto generate definitions.yaml from TSV?

### LINKED ISSUE #255: Not clear how "Content format" is stored in the yaml file

schristley:
fixes #29 
fixes #255

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/python3
 # -*- coding: utf-8 -*-
 #
 # airr-standards documentation build configuration file, created by
@@ -21,7 +21,16 @@
 import os
 import sys
 import yaml
-import yamlordereddictloader
+
+
+try:
+    import yamlordereddictloader
+except ImportError:
+    import pip
+    pip.main(["install", "--user", "yamlordereddictloader"])
+    import yamlordereddictloader
+
+# import yamlordereddictloader
 from unittest.mock import MagicMock
 sys.path.append(os.path.abspath('.'))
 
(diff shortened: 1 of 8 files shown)


schristley:
nm, gonna redo on new branch to avoid all the conflicts

javh:
Cool. I wouldn't put the `yaml_schema_to_tsv.py` at the root either. Maybe somewhere like `docs/_scripts`?

franasa:
> nm, gonna redo on new branch to avoid all the conflicts

thanks Scott, also for the clean up in ontologies. 

> Maybe somewhere like docs/_scripts?

great,  I was not sure where to put the `yaml_schema_to_tsv.py` or wheather it should be deleted since it's already in de `docs.py`.

 and also sorry for the commits/branch mess.

Issue #163, opened 2018-10-26
Question around "Study Type" and study_description


bcorrie:
Again, some feedback from the iReceptor folk (Emily and myself)...

One comment is that the tag study_description doesn't align too well with the actual use of the field, as it is really a "study_type" or "study_design". This is a bit confusing...

From Emily:

"Currently MiAIRR has a semi-controlled vocabulary for this field, as do we. We treat this field as a simplistic "Cancer study" or "Healthy baseline study" entry point to explicitly state the general study type. [Others treat] it as a study description and detailed their entry further. Do we want a more detailed field on top of study_description? More detail holds merit."

When we say "others" above, when we are importing data from other research groups and ask them for the MiAIRR information, when they respond based on the tag name (which we provide in a spreadsheet for computational purposes) leads them to entering a lengthy description. This points to the confusion around the naming and also points out the benefit of having both.

Wondering if it might make sense to have a "study_design" field basically defined as "study_description" is defined now and then add a "study_description" field that is free text and is a more detailed description of the study from the research group???

@schristley I think VDJServer has a long text description of the study in "study_description", as this is what the MiAIRR YAML spec (v0.1.0) had when we froze this version of the API. This is also what we have received back on occasion in our MiAIRR metadata spreadsheet that we send out to people to fill in...

Brian

schristley:
@bcorrie The `study_description` that we have in VDJServer is a long text description, generally the abstract for the publication. I think that field is VDJServer specific though, MiAIRR actually has a `study_type` field which is a controlled vocabulary term similar to what you mention but MiAIRR doesn't have a field for the long text description. I agree that `study_description` is useful but I'm not sure if it's "mandatory" from a minimal information perspective. That's something I suppose the MSWG can discuss.

schristley:
@bcorrie my mistake, you are right that the MiAIRR field is called `study_description`, I got confused with the name which is "Study type". So I agree with you, `study_description` is probably not the best field name, `study_type` would be better, which leads to your question about whether a long text description for the study should be added to MiAIRR.

bcorrie:
Doing an old issue check... 

Summary for this: MiAIRR "Study Type" (study_description) is a controlled vocabulary term that says "{"ontology": "NCIT", "top_node": "Study Design", "draft": true} with a description of "Generic study design" and an example of "Case-Control Study"

I think the question @schristley raised above is whether MiAIRR should:

1. Have a better field name (should study_description be study_type)?
[...]

PR #262, opened 2019-10-10
autogenerate MiAIRR TSV from yaml schema

### LINKED ISSUE #29: Auto generate definitions.yaml from TSV?

### LINKED ISSUE #255: Not clear how "Content format" is stored in the yaml file

schristley:
fixes #29
fixes #255


schristley:
@javh @bussec Are we okay with the format of the TSV changing slightly?

schristley:
Also, while the file is autogenerated, do we also want to actually commit it to the repository? Just in case any outside sources are pointing to it?  We could probably add a Travis consistency check to see if the files (committed version versus the one generated by the docs) are different. I suppose the bad thing about that is that is may force devs to build the docs to get a new version to commit.

javh:
I don't think we should commit it. People can get persistent link by linking to the docs site.

javh:
Yeah, I don't think we need the check anymore.

Do we want to render the TSV as a table in the docs somewhere, like we did for the schema fields? Eg: http://docs.airr-community.org/en/latest/datarep/rearrangements.html#fields

schristley:
> render the TSV as a table in the docs

Yeah that would be nice actually, I like being able to preview it without importing into excel

javh:
Okay, here's something not-very-good just to get us started:
http://docs.airr-community.org/en/issue-255_2/miairr/data_elements.html#miairr-elements

The page needs more words and better formatting. We should probably rename the columns too.

Pinging @franasa and @bussec.

javh:
Do we need both the `MiAIRR field designation` and `AIRR Schema field name` columns?

bcorrie:
> Do we need both the `MiAIRR field designation` and `AIRR Schema field name` columns?

I think the AIRR Schema field name is probably important, as it gives a precise mapping to the spec.

I would lean toward having the field designation as well, as a "short name" for the field???

javh:
Yeah, I agree we should have the schema field name.  

`MiAIRR field designation` contains the designations used in the MiAIRR manuscript. It now seems redundant to me, but maybe we need to keep it for clarity? Not sure if it fosters or hinders clarity to have two "fields names" in the table.

If we did drop it, then I think we would also have to remake the figure here:
http://docs.airr-community.org/en/issue-255_2/miairr/introduction_miairr.html

schristley:
> Do we need both the MiAIRR field designation and AIRR Schema field name columns?

Yes, they serve different purposes, one is the human-readable form and the other is the computable form. The first will be useful for showing in UIs and such.

It's worthwhile to change `MiAIRR field designation` to something else...

schristley:
@javh is it possible to change the table into a multi-row table, that is, each record uses multiple rows instead of on a single row. The `MiAIRR content definition` and `Field value example` could be put on their own rows, this would provide them with more space for the long content, and maybe allow the whole table to fit better width-wise.

javh:
> @javh is it possible to change the table into a multi-row table, that is, each record uses multiple rows instead of on a single row.

I don't know. I'll look into it. Not as part of the `csv-table` directive, but we there _might_ be a way to set multirow columns with the `tabularcolumns` directive.

bussec:
What is the motivation behind changing `MiAIRR field desgination`? I am not generally opposed to it, just would like to understand the issue. Is there anything else that needs to be done from the MiniStd side?

javh:
> What is the motivation behind changing `MiAIRR field designation`? I am not generally opposed to it, just would like to understand the issue. Is there anything else that needs to be done from the MiniStd side?

It's just a cosmetic concern. It makes the table wider and it contains information that is already in `MiAIRR content definition` and `AIRR Schema field name` (I assume the MiAIRR field names predates the full AIRR schema field names). So the table looks kind of cluttered on the docs site. But if we need it, we will figure out how to smoosh it.

No, I don't think there's anything needed from MiAIRR.

schristley:
@javh There is something called [Grid Tables](https://sphinx-rtd-theme.readthedocs.io/en/stable/demo/lists_tables.html#grid-tables) which supports column/row spanning. Though I'm not sure if that's available in our theme?

bussec:
@javh I agree that the `MiAIRR field designation` is redundant to the combination of the two other fields. However, these are the designations used in the MiAIRR1.0 manuscript, hence we have to kepp them for now. We can discuss deprecation for MiAIRR2.0.

javh:
Sounds good. In the interest of getting this merged in, I pushed a commit with shorter names for the columns.  The table is still a bit ugly, but I'm not sure how to clean it up at the moment. Maybe we can address that separately?

(@schristley, I don't think we can use grid tables, not because of the theme, but because we are loading the table from a tsv.)

bcorrie:
@javh I notice that there is a "study_description" field in the spec and there is no study_description in the documentation table... I also notice in the spec that that there are no x-airr fields for study_description (is this why it is not generated). I am pretty sure study_description should be included in the MiAIRR spec, no??? @bussec is that correct?

bussec:
@bcorrie this field (`study_description`) does not exist in MiAIRR, which only know `study_type`. See #163 for more background. Currently I do not consider the free text description to be minimal, but plese let me know if you think otherwise.

bcorrie:
@bussec no problem, was just a bit confused as I had forgotten that decision... One question - study_description as a field used to be in MiAIRR v1.2.1. It is now not in MiAIRR but it is still in the spec. Is there some way we can denote this so tools can track this compatibility change? It isn't a deprecated field in the spec sense (since it is still in the spec) but it is deprecated in the MiAIRR sense as it is no longer in MiAIRR...

javh:
> One question - study_description as a field used to be in MiAIRR v1.2.1.

Do you mean in the sense that it had `x-miairr: true` in the AIRR schema file at v1.2.1 of the schema?

MiAIRR "owns" the major version number, so technically I think everything in MiAIRR is supposed to be locked in to what was in v1.0.0 until we bump the repo release to v2.0.0. If my understanding is correct, then the discrepancy between the `x-miairr` tags between v1.2.1 and the tip is just a mistake.

schristley:
> If my understanding is correct, then the discrepancy between the x-miairr tags between v1.2.1 and the tip is just a mistake in v1.2.1.

It's a little more convoluted than that. The semantics of the field changed. `study_description` was MiAIRR but it described a "type" versus being a "description", to fix that semantic, the field was renamed to `study_type`, that's the MiAIRR field. Then `study_description` was re-introduced to be a "description" field but it isn't MiAIRR.

Technically, MiAIRR standard didn't change, it always required a "study type" but the field name in the schema used to store that information was changed. I suppose this should go into the Changelog...

javh:
Ah, okay. Thanks, @schristley.

PR #263, opened 2019-10-11
add sample_processing_id

### LINKED ISSUE #249: sample_processing_id in Rearrangements and Repertoire spec
In https://github.com/airr-community/airr-standards/issues/181#issuecomment-478657199

We discussed having three _id fields at the rearrangement level to disambiguate mapping rearrangements back to different combinations of repertoires, samples, and data processing.

> 
> At the AIRR TSV level, although sequence_id is necessary, I don't think it is sufficient in that:
> 
> 1. If a specific sequence_id is processed by two different SoftwareProcessing pipelines (e.g. igblast and mixcr) then we need to be able to differentiate them in an AIRR TSV file with a SoftwareProcessingID.
> 2. If we want to be able to identify which SampleProcessing regime was applied to get to a specific SequencingRun we need the SampleProcessingID. Although it is possible to go from a sequence_id in an AIRR TSV file to specific SequencingRun entity that came from a specific sample processing regime, without the SampleProcessingID in the AIRR TSV the only way to determine that link is to search the entire set of fasta/fastq files for the sequence_id (correct me if I am wrong).
> 3. If we want to provide a mechanism to tell which Repertoire (the abstract organizational unit of analysis that is defined by the researcher) a single line in an AIRR TSV file comes from, then we need a RepertoireID.
> 
> I think if we don't have all three and given our current metadata structure we lose the ability to differentiate at one of these levels. At the same time, it appears there are arguments that both the current structure is desirable and the there is a need for identifying things from the AIRR TSV file at each of these levels.
> 
> Given the level of complexity of the metadata structure that we are capturing and the relatively small amount of added information at the rearrangement level (three identifiers, RepertoireID, SampleProcessingID, and SoftwareProcessingID), it seems like having those three fields in a rearrangement would seem to make sense to me. It certainly gives us the most flexibility and power moving forward, and the lack of any one of them seems to restrict things substantially???

This seemed to have general consensus in #181 that this was reasonable, but we currently only have repertoire_id (RepertoireID from above) and data_processing_id (SoftwareProcessingID). SampleProcessingID was not added to the spec.

Reopening to revisit...

schristley:
fixes #249


bcorrie:
I think this can be merged...

PR #270, opened 2019-10-25
Fix file sync problem between spec and MiAIRR TSV

bcorrie:


bcorrie:
@bussec I noticed that the current airr-schema.yaml does not have an ontology listed for tissue, and yet I thought this was one of the ontologies that we selected? Is that correct? Does this need to be updated in the spec (and then in the MiAIRR TSV file).

bcorrie:
> @bussec I noticed that the current airr-schema.yaml does not have an ontology listed for tissue, and yet I thought this was one of the ontologies that we selected? Is that correct? Does this need to be updated in the spec (and then in the MiAIRR TSV file).

Hmm, the same is true for disease_diagnosis. I would add these, but I am not sure what the root node in the ontology should be...

bussec:
Yes, both field have and ontology, as per OntoVoc Sprint Report 11/2018:
* Diagnosis (`disease_diagnosis`)
  * Ontology: DOID   https://bioportal.bioontology.org/ontologies/DOID
  * License: CC-BY
  * notes: Features ICD cross-reference
* Tissue (`tissue`)
  * Ontology: Uberon   https://bioportal.bioontology.org/ontologies/UBERON
  * License: CC-BY

bussec:
I will check on the root nodes... just realized that they are not in the OntoVoc Document.

bcorrie:
Added the ontologies for disease_diagnosis and tissue to both the spec and the MiAIRR TSV file.

schristley:
@bcorrie in case you aren't aware, #262 will be deleting the TSV file so I don't think you need to manually make these changes. You probably just want to verify that #262 is generating the TSV with the proper info.

bcorrie:
> @bcorrie in case you aren't aware, #262 will be deleting the TSV file so I don't think you need to manually make these changes. You probably just want to verify that #262 is generating the TSV with the proper info.

Yep, but until that was done I thought it valuable to keep the two up to date... And I noticed that they were different because I actually use the TSV file. I think we should be careful - if you provide something as a TSV file (even on just a docs site) you can assume that people will use it as more than just a human readable file (as @bussec suggests in #262). We all know TSVs are handy to compute on. 

So if we have a TSV, we should make sure it is correct so if people use it they won't get in to trouble. If we want it to live as a human readable table, I would suggest generating the table in markdown and take away the TSV completely.  8-)

schristley:
okay, yeah, I guess meant deleting the TSV from the github but a TSV will still be downloadable from the docs.

javh:
#262 also contains a doc update to render the TSV:
http://docs.airr-community.org/en/issue-255_2/miairr/data_elements.html#miairr-elements

So it'll build the TSV from the yaml spec and then that TSV will be rendered by sphinx.

Best thing might actually be to just clean up the column names and rendering so we can merge #262 into master... All the functionality is there, yes?

schristley:
@bcorrie with the TSV being autogenerated now, can you double check the resultant table in the docs and the TSV. If everything looks good, you can close this PR.

bcorrie:
OK, looks like the auto generation of the TSV is taking care of this, closing...

bcorrie:
Changes WRT the TSV file not required, ontology changes are update in #343

PR #271, opened 2019-10-27
Suggested changes based on discussion

scharch:
- Added `CloneSet` as a top-level object. Right now just links to the `DataProcessing` and a list of the `Repertoire`s.
- Moved `CloneRepertoire` to a sub-object of `Clone`. Added a CloneRepertoireList field to `Clone` to link. Also added some sample fields to `CloneRepertoire`.
- Added a TreeList field to `Clone` and started filling out `Tree` and `Node` objects.

Please check my syntax carefully, as I am dumbly copy-paste-ing without really knowing what I am doing.

### DIFF
--- a/specs/airr-schema.yaml
+++ b/specs/airr-schema.yaml
@@ -1847,9 +1847,27 @@ Rearrangement:
                 subset: data (processed sequence)
                 name: V(D)J germline reference database
 
+# 1-to-n relationship between a CloneSet and the Repertoires it is derived from
+# Question: Do we want to include a direct link to `Subject` in this object?
+CloneSet:
+    discriminator: AIRR
+    type: object
+    properties:
+        clone_set_id:
+            type: string
+            description: Identifier for the clone set object.
+        repertoire_list: #did I do this correctly?
+            type: array
+            description: List of repertoires that were processed to produce clone set
+            items:
+                - $ref: '#/Repertoire'
+        data_processing_id:
+            type: string
+            description: Indentifier for the data processing object
+            $ref: '#/DataProcessing'
+
+
 # A unique clone object that has been constructed from a single clone processing.
-# The information provided here is common among or aggregated across all repertoires
-# in the clone processing.
 Clone:
     discriminator: AIRR
     type: object
@@ -1857,9 +1875,10 @@ Clone:
         clone_id:
             type: string
             description: Identifier for the clone.
-        data_processing_id:
+        clone_set_id:
             type: string
             description: Identifier for the data processing object.
+            $ref: '#/DataProcessing' #is this right?
         v_call:
             type: string
             description: V gene with allele. For example, IGHV4-59*01.
@@ -1954,8 +1973,19 @@ Clone:
         deletions:
             type:  string
             description: Clone deletion mutations
+        CloneRepertoireList:
+            type: array
+            description: Properties of the clone members in each Repertoire
+            items:
+                - $ref: '#/CloneRepertoire'
+        TreeList:
+            type: array
+            description: Trees describing the phylogenetic structure of the clone.
+            items:
+                - $ref: '#/Tree'
 
-# Clone information for a specific repertoire within a single clone processing.
+# Clone information within a single repertoire
+# 1-to-n relatonship between Clone and Clone Repertoire
 CloneRepertoire:
     discriminator: AIRR
     type: object
@@ -1963,15 +1993,40 @@ CloneRepertoire:
         clone_id:
             type: string
             description: Identifier for the clone.
-        data_processing_id:
-            type: string
-            description: Identifier for the data processing object.
+            $ref: '#/Clone'
         repertoire_id:
             type: string
             description: Identifier for the repertoire object.
+        rearrangements:
+            type: array
+            description: List of rearrangments that are members of this Clone in this Repertoire
+            items:
+                allOf:
+                    - $ref: '#/RearrangementSet' #I think we don't need this yet...
+                    - $ref: '#/Rearrangement'
+        #some examples copied from the immunedb schema
+        instances:
+           type: integer
+           description: Clone overall instance count 
+        avg_v_identity:
+            type: string
+            description: The average V-gene NT identity across all the sequences in a clone.
+        top_copy_seq_sequence:
+            type: string
+            description: The full aligned sequence with the highest copies in the clone
+        top_copy_seq_copies:
+            type: integer
+            description: The number of copies for top_copy_seq_sequence
+        top_instance_seq_sequnence:
+            type: string
+            description: The sequence for instance with the highest copies
+        top_instance_seq_copies:
+            type: integer
+            description: The number of copies for top_instance_seq_sequence
 
 # 1-to-n relationship for a clone to its trees.
-CloneTree:
+# Is there a reason to call this 'CloneTree' instead of just 'Tree'?
+Tree:
     discriminator: AIRR
     type: object
     properties:
@@ -1981,3 +2036,34 @@ CloneTree:
         clone_id:
             type: string
             description: Identifier for the clone.
+        newick_string:
+            type: string
+            description: Newick string of the tree
+        node_list:
+            type: array
+            description: Nodes in the tree
+            items:
+                - $ref: '#/Node'
+                
+# 1-to-n relationship between Tree and Node
+Node:
+    discriminator: AIRR
+    type: object
+    properties:
+        node_id:
+            type: string
+            description: Identifier for this node, matches id in newick string
+        tree_id:
+            type: string
+            description: Identifier for the tree.
+            $ref: '#/Tree'
+        clone_id: #maybe not strictly necessary
+            type: string
+            description: Identifier for the clone.
+            $ref: '#/Clone'
+        sequence:
+            type: string
+            description: DNA sequence of this node
+        meta:
+            type: object #????
+            description: A place for users to add custom info


specs/airr-schema.yaml
```diff
@@ -1847,19 +1847,38 @@ Rearrangement:
...
     type: object
     properties:
         clone_id:
             type: string
             description: Identifier for the clone.
-        data_processing_id:
+        clone_set_id:
             type: string
             description: Identifier for the data processing object.
+            $ref: '#/DataProcessing' #is this right?
```
eharkins: @scharch I think the `$ref: '#/<thing>'` is yaml's way of deduplicating something that would be literally copied there. So I think we would not want to use that in cases like this since we want to just keep an ID rather than a copy of the object referred to by the ID, if that makes sense.

specs/airr-schema.yaml
```diff
@@ -1954,24 +1973,60 @@ Clone:
         deletions:
             type:  string
             description: Clone deletion mutations
+        CloneRepertoireList:
```
eharkins: is using `List` the standard in AIRR schemas or would it be acceptable to just pluralize it into `CloneRepertoires`?

specs/airr-schema.yaml
```diff
@@ -1954,24 +1973,60 @@ Clone:
...
     discriminator: AIRR
     type: object
     properties:
         clone_id:
             type: string
             description: Identifier for the clone.
-        data_processing_id:
-            type: string
-            description: Identifier for the data processing object.
+            $ref: '#/Clone'
```
eharkins: Same thing as above with `$ref`

specs/airr-schema.yaml
```diff
@@ -1954,24 +1973,60 @@ Clone:
...
+        top_instance_seq_sequnence:
+            type: string
+            description: The sequence for instance with the highest copies
+        top_instance_seq_copies:
+            type: integer
+            description: The number of copies for top_instance_seq_sequence
 
 # 1-to-n relationship for a clone to its trees.
-CloneTree:
+# Is there a reason to call this 'CloneTree' instead of just 'Tree'?
```
eharkins: Tree seems good from an Olmsted perspective and in general since it is a "clones and trees (of clones)" schema

specs/airr-schema.yaml
```diff
@@ -1981,3 +2036,34 @@ CloneTree:
...
+            type: string
+            description: Newick string of the tree
+        node_list:
+            type: array
+            description: Nodes in the tree
+            items:
+                - $ref: '#/Node'
+                
+# 1-to-n relationship between Tree and Node
+Node:
```
eharkins: would want to remove the `$ref`s from `_id` reference fields here as well I think

specs/airr-schema.yaml
```diff
@@ -1981,3 +2036,34 @@ CloneTree:
...
+            $ref: '#/Tree'
+        clone_id: #maybe not strictly necessary
+            type: string
+            description: Identifier for the clone.
+            $ref: '#/Clone'
+        sequence:
+            type: string
+            description: DNA sequence of this node
+        meta:
+            type: object #????
```
eharkins: @scharch Is your question whether `type: object` is valid (it is) or just putting question marks to indicate this still being up in the air as far as format for node metadata?

specs/airr-schema.yaml
scharch: dunno - @schristley and @javh ?

specs/airr-schema.yaml
scharch: @eharkins the question is whether an `object` is the best/most appropriate type. My assumption is a list of arbitrary and undefined properties isn't valid so they have to get put inside an object, but maybe there's a simpler way to handle it.

specs/airr-schema.yaml
scharch: I am following the `Repertoire` schema here, to be consistent:
https://github.com/airr-community/airr-standards/blob/2ea341e121e7e6e8c2fa886ee6976a94ea82cd82/specs/airr-schema.yaml#L1189
@schristley?

eharkins: Whoops, [looks like this actually comes from Swagger](https://swagger.io/docs/specification/using-ref/). In any case, I think 
> we would not want to use that in cases like this since we want to just keep an ID rather than a copy of the object referred to by the ID, if that makes sense.

still applies.

specs/airr-schema.yaml
eharkins: Object makes sense to me for now unless others have suggestions for an improvement.

specs/airr-schema.yaml
```diff
@@ -1847,19 +1847,38 @@ Rearrangement:
...
-# in the clone processing.
 Clone:
     discriminator: AIRR
     type: object
     properties:
         clone_id:
             type: string
             description: Identifier for the clone.
-        data_processing_id:
+        clone_set_id:
```
eharkins: Is there any benefit to having this relationship referenced the other direction (directly as opposed to through `repertoire_list`). That is, in addition to (or instead of) having a clone_set_id in a Clone, we would have a list of clone_id in a CloneSet.

Or will a CloneSet always be composed of sets of Repertoires and so `repertoire_list` is enough to describe the relationship between CloneSet and Clone?

scharch: The way I have it set up, you need the `clone_set_id` in `Clone` in order to be able to get back to the `DataProcessing`, but I think that you're right, `CloneSet` should probably have a list of `clone_id`s in it.

schristley:
@scharch sorry, I'm finally catching up...

The `CloneSet` construct is interesting. I was wondering if from your earlier comments that you were suggesting such a thing or not. I'd wondered myself if something like `CloneSet` was needed or not, and in an early design I had something like it. I see it as analogous to `Repertoire` which groups a set of rearrangements together, so `CloneSet` groups a set of clones together.

Is a CloneSet for all repertoires and subjects that were processed together, or is a clone set for a specific subject and the set of repertoires for that specific subject? So if a study has 10 subjects in it, will there be 10 `CloneSet` objects for the study?

The current schema can already give you that list of repertoires in `CloneSet`, it just isn't immediate, you have to do some searching:

* Let's say you start with a study or repertoire id, as they are both pretty similar.
* From there, you can get a `data_processing_id` for the "clone processing" that was done.
* Now walk through the set of repertoires, and pull out each one that has that `data_processing_id`, that gives you the list of repertoires.

Now that list is for all subjects. So if you want to break it down further, you would iterate through that list of repertoires and split them out by subject id.

schristley:
@scharch In general, when representing 1-to-n relationships, it's better to put the relationship link in the "n" object versus the "1" object. I say in "in general" because it tends to be okay if the number of "n" objects will be small. But if the "n" is already large enough to be stored separately, then it's more efficient for them to "point back" to the "1" object. So constructs like this:

```
        CloneRepertoireList:
            type: array
            description: Properties of the clone members in each Repertoire
            items:
                - $ref: '#/CloneRepertoire'
```

should be removed and instead put the id of the "1" object, in this case `clone_id` into the "n" object like so.

```
CloneRepertoire:
    discriminator: AIRR
    type: object
    properties:
        clone_id:
            type: string
            description: Identifier for the clone.
```

Here is another example, this should be removed and the "n" object (rearrangement in this case) should point back to the clone object (using `clone_id`).

```
        rearrangements:
            type: array
            description: List of rearrangments that are members of this Clone in this Repertoire
            items:
                allOf:
                    - $ref: '#/RearrangementSet' #I think we don't need this yet...
                    - $ref: '#/Rearrangement'
```

Now conceptually these are correct, but from a data storage standpoint, they are inefficient. Essentially, what they are saying is to store the "n" objects within the "1" object.

Now actually, in the schema, those relationship links (`clone_id`) already exist, so adding those "arrays" into the "1" object schema is redundant.

schristley:
@scharch @eharkins The `Tree` and `Node` schemas confuse me. According to @matsen [here](https://github.com/airr-community/airr-standards/pull/251#issuecomment-545068699), the newick format offers internal node labels, and everything can be stored in the newick string, so all we should need is this:

```
Tree:
    discriminator: AIRR
    type: object
    properties:
        tree_id:
            type: string
            description: Identifier for the tree.
        clone_id:
            type: string
            description: Identifier for the clone.
        newick_string:
            type: string
            description: Newick string of the tree
```

we just need to give each tree an identifier and link it to a clone.

Unless you are creating a `Node` schema in an attempt to formalize the names of the attributes? The problem is the schema is for describing a JSON format, not a NEWICK format. So if we want to formalize those names within the NEWICK format, we need to do it in a different way...

Even though @matsen says

> We are not creating a custom tree format

it still looks like a custom tree format to me. According to the schema, if the data is In a file, it will have the JSON format

{
 "tree_id": 123,
 "clone_id": 456,
 "newick_string": "( ... )"
}

So can I just pass this file to (say) PYHLIP, as is, without changes, and it will work properly?

eharkins:
@schristley Sorry for the confusion! If by 
> everything can be stored in the newick string

, you mean including node attributes, then that is not the case for our schema.  

> we want to formalize [the names of the node attributes] within the NEWICK format 

We do not want to do this. The newick format remains a string in our schema, so we can 
> just pass [the newick string] to (say) PYHLIP, as is, without changes, and it will work properly 

We are simply using the node identifiers from the newick to reference `Node` objects (separate from `Tree` objects) which allow for storing attributes for each node outside of the newick string. 

Does that make sense?

scharch:
> So if a study has 10 subjects in it, will there be 10 `CloneSet` objects for the study?

Yes, because `Clone`s cannot span multiple people

> The current schema can already give you that list of repertoires in `CloneSet`, it just isn't immediate, you have to do some searching:
> * Let's say you start with a study or repertoire id, as they are both pretty similar.
> * From there, you can get a `data_processing_id` for the "clone processing" that was done.
> * Now walk through the set of repertoires, and pull out each one that has that `data_processing_id`, that gives you the list of repertoires.

Not really... A `Repertoire` contains a `DataProcessing` that describes how the `Rearrangement`s were derived from the raw sequencing data; it does not have a (formal) slot for a different `DataProcessing` ("CloneProcessing") that describes how those `Rearrangement`s were grouped into `Clone`s. Moreover, I don't think that would be the desired solution, as a researcher might have a standard workflow, and hence a single `DataProcessing` that is linked by multiple `CloneSet`s and hence finding all `Repertoire`s with that `DataProcessing` won't give the correct set.

scharch:
> Even though @matsen says
> 
> > We are not creating a custom tree format
> 
> it still looks like a custom tree format to me.

@eharkins beat me to this, but I would just add that the term "tree format" here may be what's causing the confusion. When we say "tree format" we very specifically mean the *phylogenetic relationships*, ie which nodes are descended from which other nodes. That is what is captured by the Newick string and what we don't want to reinvent. The sequence and other properties of a node are not part of the "tree" (ie they do not express phylogenetic relationships); those are what get stored in the proposed `Node` objects.

schristley:
> the term "tree format" here may be what's causing the confusion. When we say "tree format" we very specifically mean the phylogenetic relationships, ie which nodes are descended from which other nodes. That is what is captured by the Newick string and what we don't want to reinvent. The sequence and other properties of a node are not part of the "tree" (ie they do not express phylogenetic relationships); those are what get stored in the proposed Node objects.

@scharch @eharkins 
Okay, that's what I suspected. Thank you for explicitly stating that. I was definitely using the more inclusive "tree" to mean the nodes, edges, connections, attributes, identifiers, all of it. So going forward, to avoid confusion, are we sticking with "tree" to just be the Newick string? And if so, what do we name this larger construct with all the other stuff?

schristley:
@scharch @eharkins @matsen 

Part of what I was trying to convey with this [comment](https://github.com/airr-community/airr-standards/pull/251#issuecomment-541227335) is that the "format" isn't the driving question, it's the data model that we want to focus upon.

So let me pose this question, do you consider these two data models to be essentially the same?

```
Tree:
  type: object
  properties:
    tree_id:
        type: string
        description: Identifier for the tree.
    clone_id:
        type: string
        description: Identifier for the clone.
    nodes:
      type: array
      items:
        $ref: '#/Node'
    edges:
      type: array
      items:
        $ref: '#/Edge'
```

versus

```
Tree:
    type: object
    properties:
        tree_id:
            type: string
            description: Identifier for the tree.
        clone_id:
            type: string
            description: Identifier for the clone.
        newick_string:
            type: string
            description: Newick string of the tree
        node_list:
            type: array
            description: Nodes in the tree
            items:
                - $ref: '#/Node'
```

The answer should be yes, right? Because the only real difference is in the first, the edges are stored in a list, while in the second, the edges are encoded inside a string (which happens to be a newick format).

Now let me take it one step forward. These are both data models, and we can have one or more "formats" that store this data (in a file, database, etc.). The most obvious one is the JSON format. 

```
{
"tree_id": 123,
"clone_id": 456,
"edges": [],
"nodes": []
}

{
"tree_id": 123,
"clone_id": 456,
"newick_string": "( ... )".
"nodes": []
}
```

But it's also perfectly reasonable to store in Nexml format:

```
<!-- nested inside /nexml/trees, preceded by an otus element -->
<!-- A tree with float edges. -->
<tree id="tree1" xsi:type="nex:FloatTree" label="tree1">
  <node id="n1" label="n1" root="true"/>  <!-- this is the root node -->
  <node id="n2" label="n2" otu="t1"/>  <!-- this is probably a tip -->
  <node id="n3" label="n3"/>
  <node id="n4" label="n4"/>
  <node id="n5" label="n5" otu="t3"/>
  <node id="n6" label="n6" otu="t2"/>
  <node id="n7" label="n7"/>
  <node id="n8" label="n8" otu="t5"/>
  <node id="n9" label="n9" otu="t4"/>
  <!-- optional root edge, for coalescent trees.
       note: only has target  -->
  <rootedge target="n1" id="re1" length="0.34765" />
  <edge source="n1" target="n3" id="e1" length="0.34534"/>
  <edge source="n1" target="n2" id="e2" length="0.4353"/>
  <edge source="n3" target="n4" id="e3" length="0.324"/>
  <edge source="n3" target="n7" id="e4" length="0.3247"/>
  <edge source="n4" target="n5" id="e5" length="0.234"/>
  <edge source="n4" target="n6" id="e6" length="0.3243"/>
  <edge source="n7" target="n8" id="e7" length="0.32443"/>
  <edge source="n7" target="n9" id="e8" length="0.2342"/>
</tree>
```

You can imagine a YAML format, a TSV format or something else.

Now would you agree that regardless of the "format", they are all essentially storing the same information and represent essentially the same data model?

scharch:
> Now would you agree that regardless of the "format", they are all essentially storing the same information and represent essentially the same data model?

Yes, but the data model that stores the tree as a list of edges is not useful because there's also "libraries" issue. If you want to manipulate the tree (eg reroot it, as Olmstead currently does) or display it, you either need it in Newick format or you need to develop a whole new set of tools from scratch. Moreover, it would be simple to go from a Newick string to a list of edges, but the reverse transformation again requires bespoke code. As has been mentioned, NeXML is a partial exception, but the point stands.

Also, I can't think of a use case where it would relevant/interesting to query on edges, anyway...

schristley:
> If you want to manipulate the tree (eg reroot it, as Olmstead currently does) or display it, you either need it in Newick format or you need to develop a whole new set of tools from scratch. Moreover, it would be simple to go from a Newick string to a list of edges, but the reverse transformation again requires bespoke code. As has been mentioned, NeXML is a partial exception, but the point stands.

Okay, yeah, that was one of the issues I wanted to talk through to make sure we're okay with it. Changing edge attribute and/or the tree structure requires parsing the newick string, make the appropriate changes, then generating a new newick string to be stored.

What I understand everybody to be saying is that pretty much all the tools accept newick strings anyways, so storing the edges in a different data model and generating a newick string on-the-fly is really just adding unnecessary overhead (plus having to write/support code to do that), therefore it's better to just store the newick string directly as it can be passed with change to many tools?

schristley:
> Not really... A Repertoire contains a DataProcessing that describes how the Rearrangements were derived from the raw sequencing data; it does not have a (formal) slot for a different DataProcessing ("CloneProcessing") that describes how those Rearrangements were grouped into Clones. Moreover, I don't think that would be the desired solution, as a researcher might have a standard workflow, and hence a single DataProcessing that is linked by multiple CloneSets and hence finding all Repertoires with that DataProcessing won't give the correct set.

@scharch It sounds to me that you are advocating for a [CloneProcessing](https://github.com/airr-community/airr-standards/blob/f7fbf9bcaa1f8e839a51a091ec34b61636b925fb/specs/airr-schema.yaml#L1845) object that I originally had in the schema but eliminated. That CloneProcessing had an id, a list of repertoire_ids, plus some info on how clones where generated.

In your `CloneSet`, you don't have the info about the clones where generated, I gather this information is still in `DataProcessing` as `CloneSet` holds a `data_processing_id`

matsen:
> What I understand everybody to be saying is that pretty much all the tools accept newick strings anyways, so storing the edges in a different data model and generating a newick string on-the-fly is really just adding unnecessary overhead (plus having to write/support code to do that), therefore it's better to just store the newick string directly as it can be passed without change to many tools?

Yes, exactly!

scharch:
> It sounds to me that you are advocating for a [CloneProcessing](https://github.com/airr-community/airr-standards/blob/f7fbf9bcaa1f8e839a51a091ec34b61636b925fb/specs/airr-schema.yaml#L1845) object that I originally had in the schema but eliminated.

We eliminated `CloneProcessing` as a distinct *class* of object, but there still needs to be a separate instance `DataProcessing` to hold the information you were proposing to put into `CloneProcessing`.

> In your `CloneSet`, you don't have the info about the clones where generated, I gather this information is still in `DataProcessing` as `CloneSet` holds a `data_processing_id`

Yes, just not the same `DataProcessing` instance as linked by the `Repertoire`, since generating `Rearrangement`s and `Clone`s are very different types of analyses.

schristley:
> We eliminated CloneProcessing as a distinct class of object, but there still needs to be a separate instance DataProcessing to hold the information you were proposing to put into CloneProcessing.

> Yes, just not the same DataProcessing instance as linked by the Repertoire, since generating Rearrangements and Clones are very different types of analyses.

I'm not sure I agree with that. I think MiAIRR perceived DataProcessing to encompass all of the data processing in the study, not just only for rearrangements, so clone processing info might be in the same data record as for the rearrangements versus a separate data record. I also seem to remember @javh saying (though I cannot find the comment) that some tools don't do the rearrangements/clones in separate steps, they may go from raw sequences to clones in one process, so that would be represented with just a single DataProcessing record.

schristley:
>> So if a study has 10 subjects in it, will there be 10 CloneSet objects for the study?

>Yes, because Clones cannot span multiple people

Regardless of the data processing, I think this is the more critical point. The `CloneSet` most directly represents the set of repertoires that were processed together for a single subject to define a set of clones. The question in my mind, is there a simpler way to represent information that doesn't require a separate object and data records to hold it (and presumably file format, etc)? Can this information be incorporated in a simple way into the existing `Repertoire` object and/or in the `Clone` and `CloneRepertoire` object?

Furthermore, in what might be the common use case where clones are defined on just a single `Repertoire`, the `CloneSet` is rather wasteful as `Clone` could just have a `repertoire_id` field to point to that single repertoire.

scharch:
> I think MiAIRR perceived DataProcessing to encompass all of the data processing in the study, not just only for rearrangements, so clone processing info might be in the same data record as for the rearrangements versus a separate data record. I also seem to remember @javh saying (though I cannot find the comment) that some tools don't do the rearrangements/clones in separate steps, they may go from raw sequences to clones in one process, so that would be represented with just a single DataProcessing record.

Ok, I could see this. It would mean that [sotfware_versions](https://github.com/airr-community/airr-standards/blob/c653e77d5f4a9f2fbd585796f5059a0fcc0693a7/specs/airr-schema.yaml#L1130) would have to allow multiple entries for cases in which the tools are not the same (eg IgBlast for annotations followed by Clonify for determining clones) but that seems easy enough. Might pay to do the same for data_processing_protocols, though not strictly necessary.

ETA: This theoretically results in a single clone being linked to multiple `DataProcessing` objects in a case where multiple `Repertoire`s are processed with different annotators, but then the resulting `Rearrangement`s are pooled together and processed with the same clonal analysis tool. I ...think that's ok? but wanted to point it out.

scharch:
> The question in my mind, is there a simpler way to represent information that doesn't require a separate object and data records to hold it (and presumably file format, etc)?

I mean, that was why I was advocating limiting `Clone`s to a single `Repertoire` in the first place. I suppose we could define some kind of `domain_repertoire_list` within `Clone` that spells out the `Repertoire`s used in the analysis (separate from the `clone_repertoire_list` which tells you the subset of those in which this particular `Clone` was found). YMMV on whether copying that info into each and every clone is better or worse than having a topline `CloneSet` that is unnecessary pass-through in most cases.

bcorrie:
> Ok, I could see this. It would mean that [sotfware_versions](https://github.com/airr-community/airr-standards/blob/c653e77d5f4a9f2fbd585796f5059a0fcc0693a7/specs/airr-schema.yaml#L1130) would have to allow multiple entries for cases in which the tools are not the same (eg IgBlast for annotations followed by Clonify for determining clones) but that seems easy enough. Might pay to do the same for data_processing_protocols, though not strictly necessary.

This is basically how we record (and therefore report) these:
```
"software_versions": "SRA Toolkit 2.8.2-1, cutadapt 1.14, fastqc 0.11.5, pear 0.9.10, biopython 2.7.13, MiXCR v2.1.10"
"data_processing_protocols": "Primers removed, quality filtered (cutadapt), paired (PEAR), Annotated (MiXCR)"
```
See https://ipa3.ireceptor.org/v2/samples?sample_id=p2255_d180_part1

javh:
Now that we have merged in #251, would we prefer to:

1. Leave this issue open as a starting point for future discussion on Repertoire grouping and clone  sets?
2. Close this and start fresh?

scharch:
I lean toward starting fresh. If @schristley and @eharkins agree I can collect some of the discussion and start a new issue/PR

schristley:
yeah, I would close this and start fresh.

Issue #258, opened 2019-10-03
Should we add fields for gene family and gene on rearrangements?


schristley:
The rearrangement schema has locus and allele fields but it does not have fields for the gene family and the gene. For the ADC queries, that means there is no simple mechanism to query on them. The current "hack" is to do a substring search, but this is inefficient, and worse is dependent upon the nomenclature for the name.

bcorrie:
I would support this... We actually have this already, as we already generate these fields internally in our repositories for exactly this purpose. For us to optimize searches, we take v_call and if it is a substring at the gene or family level we do an exact match on our internal ir_gene_family and ir_gene_gene fields 8-)

As discussed, this is error prone as the logic for  mapping a v_call string to a gene or family can be problematic...

javh:
I lean in favor of these being custom fields, as they are gene nomenclature specific concepts.  Having reserved fields for gene and allele might be fine though. Family is more sketchy.

schristley:
> I lean in favor of these being custom fields, as they are gene nomenclature specific concepts.

Yep, I'm mainly thinking about reserving the field names. I consider these to be annotations in the germline_database but that's not going to get resolved anytime quickly.

In the OGRDB [schema](https://github.com/airr-community/ogrdb/blob/master/schema/inferred_gene_submission_schema.yaml#L877) these are the `gene_subgroup` and `subgroup_designation` fields, so instead of specifically saying "family", they are gene groups.

We could just make this an informal spec, or we can have them in the schema.

javh:
`v_allele`, `v_gene`, and `v_subgroup` then?

bcorrie:
Is v_allele == v_call? Or are we saying that this is a shortened field that for IGHV1-1*01 that v_allele == "01"

javh:
I was thinking:

```
v_call: IGHV1-1*01
v_subgroup: IGHV1
v_gene: IGHV1-1
v_allele: 01
```

Which is not perfectly semantically consistent, but matches with the GenBank structure:

```
     V_segment       154..447
                     /gene="IGHV3-73"
                     /inference="COORDINATES: alignment:IgBLAST:1.6.1"
                     /allele="02"
```

schristley:
@javh sounds good, maybe bring up in DataRep call to see if anybody has any comments? What about d, j and c?

javh:
> What about d, j and c?

Same structure, I presume. Though, subgroup and gene will be redundant for the IMGT nomenclature.

[...]

PR #282, opened 2019-11-25
Add locus x-airr extensions

### LINKED ISSUE #272: Addition of locus field in set 6
From a single cell perspectives - v_call, d_call,c_call, j_call and cdr3 annotations are suitable for both heavy (H) and light chains (K, L). It is notable from the names of the genes ('VH10.1.86') if they are heavy or light but this will require user to hard-code the names in the query. So, I’m wondering if it might make sense to add a locus field that is a controlled vocabulary (H, K or L) in set 6 to distinguish between the V,D,J, CDR3 regions of heavy and light chains .

For instance for a single cell, information would look like
```
[
  {
    "Processed sequence with annotations": [
      {
        "c_call": "Igkc", 
        "d_call": NaN, 
        "j_call": "JK4", 
        "junction_aa": "LQSDNLPFT", 
        "locus": "K", 
        "v_call": "bt20"
      }, 
      {
        "c_call": "Ighm", 
        "d_call": NaN, 
        "j_call": "JH4", 
        "junction_aa": "ARSGSYYAMDY", 
        "locus": "H", 
        "v_call": "36-60.1.46"
      }
    ], 
    "cell_id": "1003", 
    "repertoire_id": "0f00cca2c93a21d25a2413b06bba3e81fd5ef3bd5b7ed407c7c189e3771dbf3d", 
    "tissue": "spleen"
  }
]



```

bcorrie:
Added the miairr true, set 6 reference as per #272

Fixes #272

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -800,7 +800,10 @@ PCRTarget:
                 - TRB
                 - TRD
                 - TRG
-            description: Designation of the target locus according to IMGT nomencleature
+            description: >
+                Designation of the target locus. Note that this field uses a controlled vocubulary that is meant to
+                provide a generic classification of the locus, not necessarily the correct designation according to
+                a specific nomenclature.
             example: IGK
             x-airr:
                 miairr: true
@@ -1443,7 +1446,26 @@ Rearrangement:
                 adc-api-optional: true
         locus:
             type: string
-            description: Gene locus (chain type). For example, IGH, IGI, IGK, IGL, TRA, TRB, TRD, or TRG.
+            enum:
+                - IGH
+                - IGI
+                - IGK
+                - IGL
+                - TRA
+                - TRB
+                - TRD
+                - TRG
+            description: >
+                Gene locus (chain type). Note that this field uses a controlled vocubulary that is meant to provide a
+                generic classification of the locus, not necessarily the correct designation according to a specific
+                nomenclature.
+            example: IGH
+            x-airr:
+                miairr: false
+                required: true
+                nullable: true
+                adc-api-optional: false
+                name: Gene locus
         v_call:
             type: string
             description: >
--- a/lang/python/airr/specs/airr-schema.yaml
+++ b/lang/python/airr/specs/airr-schema.yaml
@@ -800,7 +800,10 @@ PCRTarget:
                 - TRB
                 - TRD
                 - TRG
-            description: Designation of the target locus according to IMGT nomencleature
+            description: >
+                Designation of the target locus. Note that this field uses a controlled vocubulary that is meant to
+                provide a generic classification of the locus, not necessarily the correct designation according to
+                a specific nomenclature.
             example: IGK
             x-airr:
                 miairr: true
@@ -1443,7 +1446,26 @@ Rearrangement:
                 adc-api-optional: true
         locus:
             type: string
-            description: Gene locus (chain type). For example, IGH, IGI, IGK, IGL, TRA, TRB, TRD, or TRG.
+            enum:
+                - IGH
+                - IGI
+                - IGK
+                - IGL
+                - TRA
+                - TRB
+                - TRD
+                - TRG
+            description: >
+                Gene locus (chain type). Note that this field uses a controlled vocubulary that is meant to provide a
+                generic classification of the locus, not necessarily the correct designation according to a specific
+                nomenclature.
+            example: IGH
+            x-airr:
+                miairr: false
+                required: true
+                nullable: true
+                adc-api-optional: false
+                name: Gene locus
         v_call:
             type: string
             description: >
--- a/specs/airr-schema.yaml
+++ b/specs/airr-schema.yaml
@@ -800,7 +800,10 @@ PCRTarget:
                 - TRB
                 - TRD
                 - TRG
-            description: Designation of the target locus according to IMGT nomencleature
+            description: >
+                Designation of the target locus. Note that this field uses a controlled vocubulary that is meant to
+                provide a generic classification of the locus, not necessarily the correct designation according to
+                a specific nomenclature.
             example: IGK
             x-airr:
                 miairr: true
@@ -1443,15 +1446,26 @@ Rearrangement:
                 adc-api-optional: true
         locus:
             type: string
-            description: Gene locus (chain type). For example, IGH, IGI, IGK, IGL, TRA, TRB, TRD, or TRG.
+            enum:
+                - IGH
+                - IGI
+                - IGK
+                - IGL
+                - TRA
+                - TRB
+                - TRD
+                - TRG
+            description: >
+                Gene locus (chain type). Note that this field uses a controlled vocubulary that is meant to provide a
+                generic classification of the locus, not necessarily the correct designation according to a specific
+                nomenclature.
             example: IGH
             x-airr:
-                miairr: true
-                required: false
+                miairr: false
+                required: true
                 nullable: true
-                set: 6
-                subset: data (processed sequence)
-                name: Gene locus            
+                adc-api-optional: false
+                name: Gene locus
         v_call:
             type: string
             description: >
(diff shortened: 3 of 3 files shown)


schristley:
If we add `locus` to miairr set 6, should we also make it a required field in the AIRR TSV?

javh:
> If we add `locus` to miairr set 6, should we also make it a required field in the AIRR TSV?

Yeah, I think we have to make it required if it's added to MiAIRR.

Out of curiosity, what's the justification for adding `locus` to the MiAIRR required fields? I read #272, but... is this for cases where you cannot derive locus from the various `*_call` fields? Meaning, gene annotations that aren't covered by the new guidelines in #217?

bcorrie:
> Out of curiosity, what's the justification for adding `locus` to the MiAIRR required fields?

Hmmm, maybe I jumped the gun on making the change, lets discuss around this pull request and make sure we are all agreed 8-)

schristley:
Part of me is thinking, if we are going to promote one field up to MiAIRR, maybe we should consider the other ones as well, and do them together. We discussed how AIRR TSV has a larger set of required fields than MiAIRR, and we should promote all at once. For example, `sequence` is another to consider? `productive`? Others?

javh:
I think this is a question for minimal standards. The more requirements in MiAIRR, then the larger the burden is for labs to share their data. Plus, we to figure out how to cram it all into GenBank/TLS.

schristley:
> Out of curiosity, what's the justification for adding locus to the MiAIRR required fields? I read #272, but... is this for cases where you cannot derive locus from the various *_call fields? Meaning, gene annotations that aren't covered by the new guidelines in #217?

In general, while this might work for human nomenclature (parsing v_call to get locus), this might not work for the new nomenclatures being worked on for mouse and other species. I think we added `locus` to discourage parsing of the gene name???

The question in my mind is why isn't `locus` a `required` field in AIRR TSV?

schristley:
> I think this is a question for minimal standards.

@bussec might have mentioned this elsewhere but it sounds like minimal standards wants to pass the requirements for MiAIRR set 6 over to DataRep...

javh:
> The question in my mind is why isn't `locus` a `required` field in AIRR TSV?

Not sure, actually. Though, we weren't thinking very hard about single-cell data at the time, so probably just under-appreciating the need. I think enough tools output the information that it would be a big deal to make it required - we should double check that though.

schristley:
`locus` field is also tangentially related to #258 which essentially reserves field names for the different pieces of the gene nomenclature. Seeing as we don't have a schema for germline annotations that tools can use to lookup...

bcorrie:
> > I think this is a question for minimal standards.
> 
> @bussec might have mentioned this elsewhere but it sounds like minimal standards wants to pass the requirements for MiAIRR set 6 over to DataRep...

I think the desire is to have DataRep manage the spec and then MinStd would select the fields that it thought should go in MiAIRR. So I don't think the desire is to hand over what is in Set 6 per se - that is still the mandate of MinStd (I think). I think the main concern around "handover" is that there should be one, and only one spec, and that the MiAIRR fields are a subset of the spec. The goal is to make it clear that MiAIRR is not different than the spec, but rather is a subset of the spec.

I agree that we should revisit what is in Set 6 and make some recommendations to MinStd if we think something should be added... I think anything that has adc-api-optional: false should be considered (from the docs):

locus, v_call, d_call, j_call, c_call, productive, junction_aa, junction_aa_length

as well as the _id fields that are rearrangement specific such as:

clone_id, cell_id, pair_id

schristley:
> Out of curiosity, what's the justification for adding locus to the MiAIRR required fields? I read #272, but... is this for cases where you cannot derive locus from the various *_call fields? Meaning, gene annotations that aren't covered by the new guidelines in #217?

After some thought, I'm not sure why `locus` needs to be in MiAIRR. I can understand that we want it in the DataRep schema, and even in the ADC API, but @javh is correct that this can be "derived" or at least "looked up in the germline database".

Adding the `x-airr` attribute is fine but have `miairr:false`. From @srlak issue, it sounds like the question more revolves around the ADC API versus MiAIRR. I think it's reasonable that we make it a queryable field for the ADC API. I thought we did already (it's listed in the docs), but I see it is missing `x-airr` so yeah that is the issue. This is my suggestion for schema. It would be good to specify the allowable enum values as well.

```
        locus:
            type: string
            enum:
                - IGH
                - IGI
                - IGK
                - IGL
                - TRA
                - TRB
                - TRD
                - TRG
            description: Gene locus (chain type).
            example: IGH
            x-airr:
                miairr: false
                required: false
                nullable: true
                adc-api-optional: false
```

The open question to me is `required:false` or `required:true`? It is not `required` by anything, so technically a data repository doesn't need to return it, even though it does have to support querying on the field. Does queryable (`adc-api-optional:false`) imply required?

bcorrie:
> The open question to me is `required:false` or `required:true`? It is not `required` by anything, so technically a data repository doesn't need to return it, even though it does have to support querying on the field. Does queryable (`adc-api-optional:false`) imply required?

I think I would vote for required:true, primarily because it suggests that it should be included. You don't have to, but required:true implies value, and we want to encourage people to provide such valuable fields. I would think adc-api-optional:false would imply required:true???

I worry about confusion when running queries... Lets say Repository 1 (R1) has locus and Repository 2 (R2) does not. Lets say that both R1 and R2 have IGH rearrangements.

Lets say my research only considers IG heavy chains so I query R1 and R2 for the IGH locus. I get a bunch of data from R1, but nothing from R2... This seems problematic to me, no?

srlak:
> Adding the `x-airr` attribute is fine but have `miairr:false`. From @srlak issue, it sounds like the question more revolves around the ADC API versus MiAIRR. I think it's reasonable that we make it a queryable field for the ADC API. I thought we did already (it's listed in the docs), but I see it is missing `x-airr` so yeah that is the issue. This is my suggestion for schema. It would be good to specify the allowable enum values as well.

Yes. I agree with @schristley that the issue #272 was concerning whether the locus field is queryable using ADC-API . So if this field is queryable then it is not required to be defined in miairr (Set 6).  However, in terms of defining x-airr for the locus field our suggestion would be quite similar to @schristley schema except to keep `required : true. `

      locus:
            type: string
            enum:
                - IGH
                - IGI
                - IGK
                - IGL
                - TRA
                - TRB
                - TRD
                - TRG
            description: Gene locus (chain type).
            example: IGH
            x-airr:
                miairr: false
                required: true
                nullable: false
                adc-api-optional: false


However, the question is similar to @bcorrie whether keeping the field `nullable:false` be problematic for users who doesn't have/provide that information?

bussec:
Just pushed an updated version representing the current state of our discussion.

I agree that for now we (i.e. @srlak and me) will be fine if `locus` becomes available via the API, we do not need it as part of MiAIRR. As `locus` holds a controlled vocabulary it allows direct and unambiguous identification of the locus, which is not possible by parsing one of the other strings. Precedent for this is the inclusion of `junction_aa`, which is even part of MiAIRR.

MiniStd discussed the "handover of Set 6" in the call last Friday and there is a consensus -- but no final decision yet -- that MiniStd considers this to be the best way to avoid confusion around these fields. So all decisions regarding names, formats and ontologies for the fields in the `rearrangement` object will be made by DataRep. MiniStd will only decide which of these fields are considered minimal information and thus will be included into MiAIRR Set 6.

We still might want to include `locus` as part of MiAIRR at a later point, but we will open another issue for that. If you think that there are any additional fields that MiniStd should consider for inclusing in MiAIRR, I would suggest to do this in a new issue and close this one as a season's present to ourselves. 

Are we good to go?

Fixes #272

bussec:
As mentioned in #272, MiniStd has decided to hand over governance over the Set 6 fields and does not consider `locus` to be minimal information at this point. This is represented by the current diffs.

Are we good to go?

bcorrie:
> As mentioned in #272, MiniStd has decided to hand over governance over the Set 6 fields and does not consider `locus` to be minimal information at this point. This is represented by the current diffs.
> 
> Are we good to go?

I think so - I think we can defer the discussion of whether ```nullable = false or true``` to when we resolve the gene naming extension (#295)

PR #285, opened 2019-11-27
Fix name: fields for age fields

bcorrie:
age_min, age_max, and age_unit

Noticed these are incorrect in the documentation table.


schristley:
Looks good. You need to copy the spec file to the python and R subdirectories so that is passes the CI checks.

Issue #276, opened 2019-11-08
New data_processing_files field in DataProcessing


bcorrie:
The DataProcessing object captures the processing that has been performed on some sequencing data. From a file perspective, we track the files from the output of a sequencing run. From a data provenance perspective, it seems helpful to provide a field to track the files produced by a DataProcessing step. We certainly track this, and from a data provenance perspective (for a data curator) it is very helpful to be able to link a file to a set of rearrangement data. 

I would suggest something like 

```
        data_processing_file:
            type: string
            description: File name for data produced by a data processing step.
            example: MS10R-NMonson-C7JR9_S1_R1_001.igblast.airr.tsv
            x-airr:
                miairr: false
                required: false
                nullable: true
                set: 5
                subset: process (computational)
                name: Processed data file name
```
This is related to #275 where I am suggesting that it would be useful to have something like:

"data_processing_file":"TW05B_T_naive_CD8.igblast.airr.tsv"

as part of the data_processing record for the Florian data.

schristley:
> files produced by a DataProcessing step

what's a step? You said "files" as in one or more ;-D

I probably understand the intent though. I have something similar in VDJServer (a custom field) to record the actual file name that holds the data before it's loaded into the data repository. It might have usefulness to a broader audience somewhere down the line, but right now it is pretty specific to the VDJServer process.

> it would be useful to have

Why would it be useful? Is it useful to anybody other than the original data curator?

If we can come up with clear semantics and usefulness to general users, then I think it belongs in the schema, otherwise it might be better as custom field.

bcorrie:
> > files produced by a DataProcessing step
> 
> what's a step? You said "files" as in one or more ;-D

I was kind of thinking forward to capturing something clone processing as a data processing object as well. For a given repertoire, one would have annotation data processing step, and record the files for that data processing output (e.g. TW05B_T_naive_CD8.igblast.airr.tsv). One could also have a clonal analysis data processing step, store all of the relevant information in that data processing object, including output that data processing step (e.g. TW05B_T_naive_CD8.changeo.out).

> Why would it be useful? Is it useful to anybody other than the original data curator?

Possibly not, but if you find it useful and we find it useful, that seems like it is at least worth considering. I also think it is as useful as recording the information that we record in the SequencingFiles object, which are also primarily useful to the data curator, no? So there is a precedence for this.

At the same time, I think it is important that we help data curators manage data provenance in the standard as well, as they are a primary user of the data standards.

[...]

Issue #268, opened 2019-10-17
IMGT Support for AIRR Rearrangement Schema


javh:
Veronique from IMGT has posted a review table of mappings for support of the AIRR Rearrangement schema here:
http://www.imgt.org/IMGT_vquest/vquest_airr

The main limitation they have with the current fields is the lack of support for additional D segments. They have a list of custom fields for these in the bottom of the table (yellow). Do we want to add these to the schema? (w/ or w/out modifications?)

These few stuck out to me as ones that we might want to take a closer at look at, as either needing to be renamed/remapped or excluded from the official schema (left as custom IMGT fields):

* `5prime_trimmed-n nb`
* `3prime_trimmed-n nb`
* `Sequence_analysis_category`
* `insertions`
* `deletions`

(See also #139 regarding D-D rearrangements.)

scharch:
I would put `d_number` as an excluded field, as well. Also perhaps all of the `d3_X` fields, as this seems unique to IMGT and I'm not sure if there is biological evidence for such rearrangements.

javh:
From the 2019-11-11 call:

* We will incorporate the proposed `d2_*` and `np3_*` fields into the spec, with any additions needed to complete the set of fields.
* The `d3_*` and `n_*` will remain IMGT custom fields for now at least. If more tools end up needing them, then we can add them later. 
* `d_number`, `5prime_trimmed-n nb` `3prime_trimmed-n nb` `Sequence_analysis_category, `insertions`, `deletions` will be IMGT custom fields.

javh:
Added the agreed fields to the spec in #295.

javh:
Resolved in #315.

PR #287, opened 2019-11-28
Added data_processing_files

bcorrie:
Added a data_processing_files field, non MiAIRR, optional as per the discussion in #276

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1219,6 +1219,15 @@ DataProcessing:
                 set: 5
                 subset: process (computational)
                 name: Data processing protocols
+        data_processing_files:
+            type: string
+            description: File name (or names) for data produced by a data processing step (e.g. annotation).
+            example: ERR1278153_aa.txz, ERR1278153_ab.txz, ERR1278153_ac.txz
+            x-airr:
+                miairr: false
+                required: false
+                nullable: true
+                name: Processed data file names
         germline_database:
             type: string
             description: Source of germline V(D)J genes with version number or date accessed.
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1230,7 +1230,7 @@ DataProcessing:
         data_processing_files:
             type: array
             items:
-                type string
+                type: string
             description: Array of file names for data produced by a data processing step (e.g. annotation).
             example:
                 - 'ERR1278153_aa.txz'
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1231,7 +1231,7 @@ DataProcessing:
             type: array
             items:
                 type: string
-            description: Array of file names for data produced by a data processing step (e.g. annotation).
+            description: Array of file names for data produced by this data processing.
             example:
                 - 'ERR1278153_aa.txz'
                 - 'ERR1278153_ab.txz'
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1228,9 +1228,14 @@ DataProcessing:
                 subset: process (computational)
                 name: Data processing protocols
         data_processing_files:
-            type: string
-            description: File name (or names) for data produced by a data processing step (e.g. annotation).
-            example: ERR1278153_aa.txz, ERR1278153_ab.txz, ERR1278153_ac.txz
+            type: array
+            items:
+                type string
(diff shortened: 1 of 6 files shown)


javh:
@bcorrie, do you mind if I throw the changes for #268 into this PR as well? They're semi-related.

bcorrie:
> this should be an array.

Change made - waiting on Travis checks to pass...

bcorrie:
> > this should be an array.
> 
> Change made - waiting on Travis checks to pass...

Hmmm, @schristley @javh it looks like there are some new Travis checks that have been added... Can someone fill me in on what needs to be updated for these...

javh:
The tests for the python reference library are failing:

```
======================================================================
FAIL: test_load_repertoire (tests.test_interface.TestInferface)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/airr-community/airr-standards/lang/python/tests/test_interface.py", line 78, in test_load_repertoire
    data = airr.load_repertoire(self.rep_good, validate=True)
KeyError: 'data_processing_files'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/home/travis/build/airr-community/airr-standards/lang/python/tests/test_interface.py", line 80, in test_load_repertoire
    self.assertTrue(False, 'load_repertoire(): good data failed')
AssertionError: False is not true : load_repertoire(): good data failed
======================================================================
FAIL: test_repertoire_template (tests.test_interface.TestInferface)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/airr-community/airr-standards/lang/python/tests/test_interface.py", line 52, in test_repertoire_template
    result = airr.schema.RepertoireSchema.validate_object(rep)
KeyError: 'data_processing_files'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/home/travis/build/airr-community/airr-standards/lang/python/tests/test_interface.py", line 55, in test_repertoire_template
    self.assertTrue(False, 'repertoire_template(): repertoire template failed validation')
AssertionError: False is not true : repertoire_template(): repertoire template failed validation
```

bcorrie:
> The tests for the python reference library are failing:

Are these new tests or I have I never made a change to the spec that caused these tests to fail (which I find odd) 8-)

javh:
> Are these new tests or I have I never made a change to the spec that caused these tests to fail (which I find odd) 8-)

There haven't been any commits to `test_interface.py` in 4 months... At least in main.

schristley:
@bcorrie For handling arrays with simple data type items, the validate code can simply check that each entry (row) in the array is the correct type. I cut/paste code from lines 381-388 which check basic types. I guess maybe the try/catch is needed...

```
   elif spec['items'].get('type') == 'boolean':
      try:
         self.to_bool(row, validate=True)
      except ValidationError as e:
         raise ValidationError('array field %s has %s' %(f, e))

and so on for integer and number (string?)
```

There should still be an `else` to capture unhandled types like "object".

```
                    else:
                        raise ValidationError('Internal error: array field %s in schema not handled by validation. File a bug report.' %(full_field))
```

bcorrie:
> @bcorrie For handling arrays with simple data type items, the validate code can simply check that each entry (row) in the array is the correct type. I cut/paste code from lines 381-388 which check basic types. I guess maybe the try/catch is needed...
> 
> ```
>    elif spec['items'].get('type') == 'boolean':
>       try:
>          self.to_bool(row, validate=True)
>       except ValidationError as e:
>          raise ValidationError('array field %s has %s' %(f, e))
> 
> and so on for integer and number (string?)
> ```

So what is the validation for string??? to_bool etc assume a string and convert to bool. So at this point are we assuming that this is a string so no validation is required?

bcorrie:
@schristley I changed the array processing of the validation code, can you have a review of that. It looks like it works...

schristley:
> So what is the validation for string??? to_bool etc assume a string and convert to bool. So at this point are we assuming that this is a string so no validation is required?

Um, Yes?? I was thinking the same thing when I wrote that... I guess the old code (rearrangement validation) is assuming you are reading from a TSV file and thus yes the fields come in as strings, and get converted to appropriate types.

With repertoire metadata as json, we should probably explicitly check for string type.

And now that I think about it, we probably shouldn't be using the `to_` functions to convert, because we don't convert repertoire metadata when reading/writing, thus we should just check the type directly:

```
   elif spec['items'].get('type') == 'boolean':
      if not isinstance(row, boolean):
         raise ValidationError(...
```

bcorrie:
Do you want to merge this in to master and create a separate issue for that fix?

schristley:
> Do you want to merge this in to master and create a separate issue for that fix?

nah, I committed the needed changes to this PR.

javh:
From the call:
* Can this be deferred to v2.0 with an overall of DataProcessing?
* If not, good to merge.

Issue #246, opened 2019-09-11
Uniqueness of _id fields in airr_schema.yaml


bcorrie:
@schristley I was looking at several of the _id fields in the schema, and I note in the descriptions we do not mention uniqueness criteria for many (any?) of them. I think this is a problem, isn't it??? Am I missing something?

If I go to the rearrangement level, we have several _ids (pair_id, clone_id, cell_id, rearrangement_id, repertoire_id, and data_processing_id). We don't advise or specify at what level something like a clone_id is unique... Or even a repertoire_id or data_processing_id. MiAIRR specifies that study IDs should be unique (typically an INSDC study related identifier) with subject_ids and sample_ids unique within studies. 

It is not well defined what the relationship between _ids is from this level down (pair_id, clone_id, cell_id, rearrangement_id, repertoire_id, and data_processing_id)

One can probably infer (if you know the AIRR spec well) that repertoire_id should be unique at least within a study, maybe a subject. The reality is that the repertoire_id should be unique at the repository level (as they are the IDs returned by the /repertoire API endpoint), but that isn't actually stated in the spec unless I am missing something...

data_processing_id should be unique withing a repertoire_id at least.  It feels like data_processing_id should be unique at the repository level as well, so you can easily identify a set of rearrangements that have been processed with the same data_processing without having to do a combined repertoire_id x data_processing_id query but again, nothing is explicitly stated in the spec. 

pair_id, clone_id, and cell_id should probably be unique at least unique within a repertoire_id/data_processing_id pair. If data_processing_id is unique within the repository, then it is sufficient to say unique within the data_processing_id.

Finally, rearrangement_id should be unique to the repository as well, that is it is the internal identifier for the repository for a single rearrangement entry. This is the only one that states anything about uniqueness at the moment.

Should we review this?

schristley:
> descriptions we do not mention uniqueness criteria for many (any?) of them

The global uniqueness for `repertoire_id` and uniqueness of `data_processing_id` within a repertoire is in the documentation for the repertoire schema (look under Linking Data)

http://docs.airr-community.org/en/metadata-docs/datarep/metadata.html

but you are right that most of the other _ids aren't well specified.

schristley:
> MiAIRR specifies that study IDs should be unique

I'm not sure this is true, where does it say that?

javh:
[...]

PR #294, opened 2019-12-10
minimal clones and trees schema for initial release

eharkins:
As per [our discussion](https://github.com/airr-community/airr-standards/pull/251#issuecomment-563521018) I have created this PR so that we can have a place to discuss the minimal set of fields necessary for the initial version of the clones and trees schema, to be included in the next release.

I have removed fields and objects that are tool-specific or that we have discussed leaving out of an initial version. 

I have also assigned `required` to a subset of fields being proposed for these new schema.

Please comment or commit to this PR with feedback and/or corrections.


specs/airr-schema.yaml
```diff
@@ -2259,118 +2253,152 @@ Clone:
...
             type: integer
             description: Number of amino acids in the junction sequence.
+            x-airr:
+                miairr: false
         germline_alignment:
             type: string
             description: >
                 Assembled, aligned, full-length inferred germline sequence spanning the same region
                 as the sequence_alignment field (typically the V(D)J region) and including the same set
                 of corrections and spacers (if any).
```
scharch: It seems like this definition has been copied from the rearrangements schema, but I don't think it makes sense in this context. To avoid confusion, I would recommend changing the field name, as well. Maybe:
```
naive_ancestor:
    type: string
    description: Full-length inferred sequence of the naive ancestor of this clone.
```
Also, I think it should be a required field?

Related, `germline_alignment_aa` below should become `naive_ancestor_aa`

specs/airr-schema.yaml
```diff
@@ -2259,118 +2253,152 @@ Clone:
...
             type: string
             description: >
                 Amino acid translation of the assembled germline sequence.
+            x-airr:
+                miairr: false
         v_alignment_start:
             type: integer
             description: >
                 Start position in the V segment in both the sequence_alignment and germline_alignment fields
                 (1-based closed interval).
```
scharch: Same type of comment on these fields, too, I guess, There is no `sequence_alignment` field in this object, so at least the description has to change. I think it would make more sense to just have the fields be named `v_start`, `v_end`, etc, since with only one sequence (the inferred naive ancestor) there isn't really an alignment...

specs/airr-schema.yaml
```diff
@@ -2400,9 +2439,10 @@ Node:
...
+            x-airr:
+                miairr: false
         sequence:
             type: string
             description: DNA sequence of this node
-        meta:
-            type: object #????
-            description: A place for users to add custom info
+            x-airr:
+                miairr: false
```
scharch: Should this be a required field, since (if I understand correctly) initially there is no link to a `Repertoire` object that will allow recovery of the sequence otherwise?

scharch:
I missed the call, but does this means we are not having any link between a `Clone` and the `Repertoire`(s) that holds the original sequences in the initial release?

javh:
> I missed the call, but does this means we are not having any link between a `Clone` and the `Repertoire`(s) that holds the original sequences in the initial release?

No, I think `repertoire_id` is just missing from `Clone` in this schema file. We still want to the link to `Repertoire`, because we'll have to store some of the custom fields in `Repertoire`. 

It's just the repertoire grouping / clone set object we wanted to wait on. So have clones defined on a single Repertoire, while we test the single Repertoire case and sort out how to handle the grouping issue.

> It seems like this definition has been copied from the rearrangements schema, but I don't think it makes sense in this context. To avoid confusion, I would recommend changing the field name, as well.

I thought `naive_ancestor` and `germline_alignment` were the same thing - the inferred, unmutated ancestor of the clone. You might redoing the inference with the lineage tree method, instead of the reference aligner, but I think it's suppose to be the same thing.

@eharkins?

scharch:
> I thought `naive_ancestor` and `germline_alignment` were the same thing

They are allowed to be the same thing, but I think in practice the typical usage of `germline_alignment` in the rearrangements schema is to have gaps replacing N insertions, because most tools (ie not partis) are doing per-sequence alignments and therefore can't generate `naive_ancestor`s.

specs/airr-schema.yaml
```diff
@@ -2259,118 +2253,152 @@ Clone:
...
             type: integer
             description: Number of unique rearrangements represented by this clone
+            x-airr:
+                miairr: false
         clone_count:
             type: integer
             description: Number of instances for this clone
+            x-airr:
+                miairr: false
         seed_id:
```
scharch: How does `clone_count` differ from  `rearrangement_count`?

specs/airr-schema.yaml
```diff
@@ -2259,118 +2253,152 @@ Clone:
...
+            x-airr:
+                miairr: false
+                required: true
         cdr3_end:
             type: integer
             description: CDR3 end position in the query sequence (1-based closed interval).
+            x-airr:
+                miairr: false
+                required: true
         rearrangement_count:
```
scharch: Do we want a list of `sequence_id`s from the original `Repertoire` that are included in this clone?

javh: IIRC, those are supposed to be in each `Node`. I pushed a change replacing `node_id` with `sequence_id`, moving the junction sequence fields to Node as well.

But... that's only going to make sense if you have a Tree object. Otherwise, you'll have to go back to the `clone_id` column in the Rearrangement data to get the list of `sequence_id` values. Which might be fine?

specs/airr-schema.yaml
javh: I'm confused about this distinction too. 

I would assume `clone_count` to be at the `Repertoire` level? Ie, the total number of observed clonal families in the Repertoire. That doesn't match the description here though.

scharch: ...and wouldn't belong in the `Clone` object, either

specs/airr-schema.yaml
scharch: `node_id` is not necessarily equivalent to `sequence_id`, because of internal nodes. In addition, while we could require that (for leaves) `node_id` has to match a `sequence_id` in `Repertoire`, I don't think that's currently documented as part of the schema.

Finally, while this is mostly relevant to olmstead, at the moment, if you build trees from a subset of the sequences in a clone, than even accessing the/a `Tree` object and collecting all of the `Nodes` doesn't guarantee that you will catch all of the members.

scharch: All of the annotation properties should belong to the `Clone` either because they by definition refer to the whole clone (`v_call` etc) or because they refer to the `naive_ancestor` (`junction_start` etc).
If the clone contains sequences with indels relative to the inferred `naive_ancestor`, those should (for the moment) probably be added as `Node`-level properties. For alignments etc of individual sequences, a user should either cross-reference the original `Rearrangement` or compare to the `naive_ancestor` using the indel information present on the `Node`. I don't think we want/expect this kind of mostly redundant information to be re-encoded at every `Node`.

javh: > node_id is not necessarily equivalent to sequence_id, because of internal nodes. In addition, while we could require that (for leaves) node_id has to match a sequence_id in Repertoire, I don't think that's currently documented as part of the schema.

I think it's fine to give internal nodes their own unique `sequence_id`, because `sequence_id` is already pretty context dependent in terms of what it means. There's also a use case where you'd want to dump inferred nodes out to a Rearrangement TSV to analyze them as you would any other sequence (particularly the MRCA), so you wouldn't have to do any translation of column names if we just stick with `sequence_id`.

> All of the annotation properties should belong to the Clone either because they by definition refer to the whole clone (v_call etc) or because they refer to the naive_ancestor (junction_start etc).

I was kind of scratching my head on this one too. I'm not sure whether it's a problem. How it's setup now it looks like there's an assumption that all members of a clone have the same junction length. That's a common approach by tools, but not universal. But... everything is supposed to be operating on a multiple alignment of the sequences/germline, so indels are handled by that requirement.

Maybe the solution is to not make some of the fields required in `Clone` and copy them into `Node` as well. So that you can chose whether to defined certain field across all Nodes in a clone or on a Node-by-Node basis.

All those positional fields shouldn't be required though, if they aren't essential to the tree building process and aren't required in the Rearrangement schema.

schristley:
> Is that correct?

Maybe. It changes `data_processing` though. Before `data_processing` would have a single entry which was the `primary_annotation: true`. You are suggesting that different processing steps are recorded as individual entries.

Given a repertoire, how do we know which data_processing is for clonal analysis? What if there is more than one clonal analysis processing, we need the specific `data_processing_id` to retrieve the appropriate clone records?

javh:
On the call, we briefly discussed doing some sort of abstraction/inheritance thing for DataProcessing, so that you have a single DataProcessing for alignment/cloning (eg, MiXCR) or multiple task specific ones like AlignmentProcessing/CloneProcessing/TreeProcessing that are just copies of DataProcessing objects with reserved names.

Though, that might be more complex than necessary.  Seems like you could get the same effect by reworking `DataProcessing` to be more general and adding a property that defines the task it describes (eg, `processing_tasks: "clonal clustering"`). Which, might address the problem @schristley brings up about mapping the elements in a `data_processing_id` list to specific tasks.

Related... Does `data_processing_id` need to be in `Clone`. Can this just be in `Repertoire`? And then we resolve the ambiguities in the` DataProcessing` object itself, rather than requiring the `data_processing_id` to be copied downward?

specs/airr-schema.yaml
eharkins: I think @schristley added it so maybe could clarify here?

Perhaps it's the number of times this same clone was observed/inferred across repertoires since we talked about wanting to account for the possibility that you might see a clone in different repertoires (hence the emphasis on the discussion of the clone-repertoire relationship)?

specs/airr-schema.yaml
eharkins: > Should this be a required field

`sequence_id` is required, is that what you are referring to @scharch ?

Or are you saying that `sequence_alignment` should link to a separate `Alignment` object and be required. I was interpreting `sequence_alignment` here to be the actual sequence being stored in the node.

Maybe this was already resolved by @javh's commits?

specs/airr-schema.yaml
eharkins: **node_id:**
> this is mostly relevant to olmstead, at the moment, if you build trees from a subset of the sequences in a clone, than even accessing the/a Tree object and collecting all of the Nodes doesn't guarantee that you will catch all of the members.

In Olmsted we don't currently keep track of a clone's sequences that we didn't include in a tree for that clone except summary statistics at the clone level (e.g. how many sequences belong to the clone). For our use case I'm ok with either
>  give internal nodes their own unique sequence_id, because sequence_id

, or having `node_id` be separate.


**annotation properties:**
I think it makes sense to keep these at the clone level as @scharch says. It seems 
> everything is supposed to be operating on a multiple alignment of the sequences/germline, so indels are handled by that requirement.

means that this model covers the case the @javh is pointing out, if I understand correctly.

> All those positional fields shouldn't be required though, if they aren't essential to the tree building process and aren't required in the Rearrangement schema.

I think I made them required since some of them are required by or are products of the clone inference process in partis. If this is not broadly applicable, please remove the required property from any of them.

eharkins:
Regarding `DataProcessing`, can we just allow any object to have its own  `DataProcessing` (or an id referencing one) if its fields are generic enough and it allows custom fields? This seems like it might simplify the need for different types of  `DataProcessing` objects.

Maybe this doesn't make it easy to look among a collection of `DataProcessing` objects to find ones that refer to clone processing, but going the other direction (finding the tool(s) used to create a given object) seems more valuable.

specs/airr-schema.yaml
schristley: > How does clone_count differ from rearrangement_count?

I pulled these from immunedb so my understanding might be somewhat limited. `clone_count` is the abundance of that specific clone. It might be as simple as adding up the abundances for all the rearrangements for the clone, or something more complicated. `rearrangement_count` is the number of rearrangement records for that clone, so if `rearrangement_count=7` and you did a query on the rearrangement table for that clone, it should return 7 entries.

bcorrie:
> > Is that correct?
> 
> Maybe. It changes `data_processing` though. Before `data_processing` would have a single entry which was the `primary_annotation: true`. You are suggesting that different processing steps are recorded as individual entries.

Yes... Which does require a bit of a generalization of this object which @javh brought up.

> Given a repertoire, how do we know which data_processing is for clonal analysis? What if there is more than one clonal analysis processing, we need the specific `data_processing_id` to retrieve the appropriate clone records?

In my model, there are two ways...

If you are starting at the Clone object, it points to a specific data_processing_id that was used to generate the clone.

If you are starting at the Repertoire, the only way in my model is the "software_processing" that is performed, but that is a very weak link. @javh suggestion of having a task tag would help, but I think we need to think a big about generalizing DataProcessing if we really want that...

bcorrie:
> Maybe. It changes `data_processing` though. Before `data_processing` would have a single entry which was the `primary_annotation: true`. You are suggesting that different processing steps are recorded as individual entries.

A given ```Repertoire``` could always have more than one ```DataProcessing``` object in it, as it is an array? This is how we are able to capture that you have a single ```Repertoire``` that has two rearrangements from two annotations. One ```DataProcessing``` could be from igblast and one from MiXCR. The ```primary_annotation``` field allowed the ```Repertoire``` to have a "preferred" ```DataProcessing```.

So we already have different processing steps as different ```DataProcessing``` objects in a ```Repertoire```. My assumption was that we were adding another ```DataProcessing``` object to capture that the repertoire had been processed by igblast, mixcr, and Brian's Clone Tool v0.00001 8-)

This points out a problem of course... As we lose track of whether the clone analysis was applied to the igblast rearrangements or the mixcr ones 8-(

I suppose it is possible to have a single ```DataProcessing```, that captures the entire data processing pipeline that is applied. For example, we might have

```
{"repertoire_id":"REP1", 
    ... , 
    "data_processing":
    [
        {"data_processing_id":"D1", ..., "software_versions":"igblast 1.8.0, Brian's Clone Tool v0.0001"}
        {"data_processing_id":"D1", ..., "software_versions":"MiXCR vXXX, Brian's Clone Tool v0.0001"}
    ]
} 
```
and only have one ```DataProcessing``` object that captures both the annotation and clonal analysis processes. In this case the Rearrangement and Clone objects generated by this data processing would have the same data_processing_id. You could still have different annotations and track clonal processing of those annotations with multiple ```DataProcessing`` objects. The problem here is that there is no real mechanism other than parsing software_versions to figure out if a Repertoire has clonal anlaysis.

@schristley was that more along the lines of what you were originally thinking?

bcorrie:
> Related... Does `data_processing_id` need to be in `Clone`. Can this just be in `Repertoire`? And then we resolve the ambiguities in the `DataProcessing` object itself, rather than requiring the `data_processing_id` to be copied downward?

I think so... A ```Repertoire``` can have multiple ```DataProcessing``` objects, so you need to be able to identify which ```DataProcessing``` object goes with a specific clone, which tells you how that clone was "assigned/created". Even without the dual use of ```DataProcessing``` for both annotation and clonal assignment, I think we want to support multiple clonal analysis processing on a single Repertoire, no?

specs/airr-schema.yaml
scharch: My summary of this thread:
* `Clone` should have an array of `clonal_sequence`s (?) that match `sequence_id`s in the `Repertoire`.
* `germline_alignment` should be a required field for `Clone`. I'd still like the description to be tweaked slightly to point out that it the inferred ancestor for the entire `Clone` and that the relevant `sequence_alignment`s are those of the `clonal_sequence`s (as opposed to in the Rearrangements TSV, where both fields are at the same level).
* `v_call` etc should be fields at the `Clone` level but not required
* `node_id` should be defined to match `sequence_id` in the source `Repertoire` where relevant/possible. (NB this could get dicey if analyzing multiple `Repertoire`s together.)

@javh @eharkins look ok?

specs/airr-schema.yaml
scharch: So it's just the sum of `duplicate_count` for all the rearrangements in the clone? And what about `rearrangement_count` vs `unique_rearrangement_count`?

Given the lack of clarity and the fact that this initial release is supposed to be "minimal," I propose dropping both `unique_rearrangement_count` and `clone_count`, at least for now.

specs/airr-schema.yaml
javh: 1. I don't think this is necessary, because the list of sequences in a clone will be contained in the Node objects. You can also get the `sequence_id` in the Clone from the Rearrangement object by just keying on `clone_id`. That said, having a non-required field that contains a list of `sequence_id`s seems fine to me. I would name it `sequence_id` though, because the semantics are the same as `sequence_id` in the Rearrangement object - it's just a list instead of a scalar.
2. Agreed. I think tweaking the description is a good idea. That's simpler than making a new field.
3. Agreed.
4. Yes, but I would just use `sequence_id` instead of making a new `node_id` field that contains largely the same information.

specs/airr-schema.yaml
scharch: @eharkins The question was whether `sequence` should be required. Further thoughts:
1) The field name should be `sequence_alignment` and it should be defined to match the `germline_alignment` at the `Clone` level. 
2) I think it should _not_ be a required field, as it's easy to imagine this causing an explosion in file sizes. The preferred course should be to extract the `sequence_alignment` from the source `Repertoire`.
3) The above will not work when either (a) the alignment in the `Repertoire` differs from that used to build the tree, maybe because the former was done in the context of a single sequence or (b) for internal nodes, which typically won't be part of any `Repertoire`.

specs/airr-schema.yaml
scharch: > 1. I don't think this is necessary, because the list of sequences in a clone will be contained in the Node objects.

I disagree, because there is not guarantee that all clonal members are used to build the tree (which could be subsampled) and therefore no guarantee that they will be represented by a `Node` object. However, I guess I am ok with making it a non-required field.

Totally fine with the nomenclature refinements you suggest @javh.

javh: @scharch, okay, then I agree we should include the `sequence_id` list in Clone then. This is for testing purposes after all, so we can always tweak it later.

eharkins: This sounds good to me. @scharch do you want to push a commit with the changes you listed above with @javh's amendments:
> 1 ... having a non-required field that contains a list of sequence_ids seems fine to me. I would name it sequence_id though, because the semantics are the same as sequence_id in the Rearrangement object - it's just a list instead of a scalar.

> 4. Yes, but I would just use sequence_id instead of making a new node_id field that contains largely the same information.

if we agree on them?

specs/airr-schema.yaml
eharkins: @scharch ok with me. Since we allow custom fields, tools like Olmsted (which is where `unique_rearrangement_count` came from I believe) can define these fields for themselves. 

Quantifying the members of a clonal family in different ways seems like it will always application dependent anyway.

specs/airr-schema.yaml
eharkins: 1. 
> The field name should be sequence_alignment 

seems like this is already the case here. Go ahead and update the description to reflect

> it should be defined to match the germline_alignment at the Clone level.

2.
> I think it should not be a required field, as it's easy to imagine this causing an explosion in file sizes

sounds good to me. Go ahead and make it not required when you get the chance.
> The preferred course should be to extract the sequence_alignment from the source Repertoire.

makes sense; could one also use the non-required `Clone.sequence_id(s)` field we discussed in [this thread](https://github.com/airr-community/airr-standards/pull/294#discussion_r358451825)?

3. 
> The above will not work when either (a) the alignment in the Repertoire differs from that used to build the tree, maybe because the former was done in the context of a single sequence or (b) for internal nodes, which typically won't be part of any Repertoire.

Does the addition of `Clone.sequence_id(s)` help with this (especially if it can reference sequence objects created in the process of inferring internal nodes which don't belong to a repertoire)?

specs/airr-schema.yaml
```diff
@@ -2255,154 +2249,223 @@ Rearrangement:
...
+            x-airr:
+                miairr: false
+                required: true
         newick_string:
             type: string
             description: Newick string of the tree edges.
+            x-airr:
+                miairr: false
+                required: true
         node_list:
```
eharkins: Should this be called `nodes` or `node_dict` or `nodes_by_seqid` or something like that to reflect its structure?

specs/airr-schema.yaml
eharkins: Looks like this never got addressed. It makes sense to me, so hopefully @scharch is able to commit directly to this branch now? I think @javh was going to try to fix whatever was preventing that but not sure if that happened yet. If not I can try

specs/airr-schema.yaml
eharkins: I will change this now if no one objects. Any preference among the choices here or suggestions?

specs/airr-schema.yaml
eharkins: @scharch should I resolve this thread? #302 accounts for 1 & 2 above, but I wasn't sure if 3 was answered here.

specs/airr-schema.yaml
scharch: I think it should be `node_id`, similar to how `Clone` has `sequence_id`

specs/airr-schema.yaml
scharch: I think it got settled -we left the names as they were but I updated the descriptions in #302

specs/airr-schema.yaml
eharkins: From a consistency perspective this makes sense, but actually both cases kind of confuse me by using a singular name to refer to a structure containing many of those things. Is there a reason for doing it this way?

specs/airr-schema.yaml
scharch: I think you can close it. 3 was really just an observation of the limitation of 2 (and hence a case where the field should be used despite the default preference against it). Perhaps something for the documentation.

specs/airr-schema.yaml
scharch: They get the same name because they are referencing the same thing. The values in `Clone.sequence_id` are exactly those in `Rearrangement.sequence_id` and are used to cross-reference. Same here. OTOH, I could also see using `sequence_ids` or `sequence_id_list` in `Clone` if you find that more salutary. Check with @javh, I think it was originally his suggestion to use the same field name.

eharkins: Got it. I am in favor of using the same name for the same thing but unless there is a reason to have a list of `thing` have the exact same name as `thing` (e.g. programming/querying use case requiring them to have the same name) without some plural suffix, then I think a plural suffix like you suggest woud make sense.

specs/airr-schema.yaml
eharkins: Ok, I'll close. Do you have a standard for recording changes introduced by a PR that should be reflected in the documentation?

eharkins:
Other than a few naming changes and threads which are in the process of being resolved here, it seems like the only discussion left is surrounding `DataProcessing` and its relationship(s) with various objects. 

In the spirit of this "minimal" initial schema version, I propose keeping things as simple as possible and then extending the schema to accommodate more complex use cases for `DataProcessing`.

Does [my suggestion above](https://github.com/airr-community/airr-standards/pull/294#issuecomment-565598969) seem reasonable or would someone like to propose an alternative?

scharch:
> Does my suggestion above seem reasonable or would someone like to propose an alternative?

Fine with me...

specs/airr-schema.yaml
scharch: erm @javh? should someone submit a pr on the docs page? I assume we'll need a Clone schema page over there anyway...?

javh: I hadn't thought about the docs yet... Yes, at some point we should make a Clone schema page.  Think we can get away with a skeleton for now, similar to the Alignment schema page:
https://raw.githubusercontent.com/airr-community/airr-standards/master/docs/datarep/alignments.rst

specs/airr-schema.yaml
javh: We have a `type` property, so we've been avoiding type identifying names (Hungarian notation). So just `nodes` (or `node`), similar to `sample` in `Repertoire`,  `sequencing_files` in `SequencingRun`, etc.

We don't have another dictionary-like object in the schema, but this looks correct to me according to:
https://swagger.io/docs/specification/data-models/dictionaries/
I'll have to defer to @schristley though.

And, yeah, I think `sequence_id` as the key is correct.

specs/airr-schema.yaml
javh: Yeah, @scharch and I resolved this.

javh:
> Does [my suggestion above](https://github.com/airr-community/airr-standards/pull/294#issuecomment-565598969) seem reasonable or would someone like to propose an alternative?

Seems fine to me too. (At least for now.)

specs/airr-schema.yaml
eharkins: Does this still need addressing, @scharch?

specs/airr-schema.yaml
eharkins: 👍 should I resolve this thread then since it sounds like the issue has been raised / noted or does the merging of this minimal schema version need to wait for complete documentation?

javh: I say resolve. We can add the docs after the merge.

bcorrie:
> Does [my suggestion above](https://github.com/airr-community/airr-standards/pull/294#issuecomment-565598969) seem reasonable or would someone like to propose an alternative?

I am not sure I quite follow the suggestion completely. I think the concern is that you don't want to have ```data_processing_id``` (and now ```sample_processing_id``` as I added here: https://github.com/airr-community/airr-standards/pull/294/commits/73166b05401b73b072bdb33ad2023362c2b41409) at the ```Clone``` object level, correct?

The problem is that if the ```Clone``` object just refers to a ```Repertoire``` through a ```repertoire_id``` then there is no way to determine which ```DataProcessing``` object (without a ```data_processing_id```) or ```SampleProcessing``` object (without a ```sample_processing_id```) from within that ```Repertoire``` that the specific ```Clone``` object of interest comes from... The problem is that a ```Repertoire``` can have many ```DataProcessing``` objects. The main use case for this is that you want to compare two sets of clones generated from two different tools from the same repertoire... So the ```repertoire_id`` is the same for all clones, but some clones will have one ```data_processing_id``` (from tool A) and the other clones will have a different ```data_processing_id``` (from tool B).

Without the ```data_processing_id``` at the ```Clone``` level you are stuck!!!

bcorrie:
I should note that we had exactly this same discussion for the ```Rearrangement``` object and came to the conclusion that with ```Repertoire``` having an array of ```SampleProcessing``` and ```DataProcessing``` objects, the only way to disambiguate something that belongs to a ```Repertoire``` is to have IDs that point to the ```Repertoire``` object and the ```DataProcessing```, and ```SampleProcessing``` objects within that ```Repertoire```.

eharkins:
@bcorrie 
> I think the concern is that you don't want to have data_processing_id (and now sample_processing_id as I added here: 73166b0) at the Clone object level, correct?

That wasn't my concern, I was suggesting abstracting the `DataProcessing` object so that it could be attached to any object using a `data_processing_id` field. All such objects, whether they pertained to a `Sample` or to a `Clone` or anything else, would have the same schema, which would be flexible and not require fields describing clone-processing-specific information, for example.

bcorrie:
> > I think the concern is that you don't want to have data_processing_id (and now sample_processing_id as I added here: [73166b0](https://github.com/airr-community/airr-standards/commit/73166b05401b73b072bdb33ad2023362c2b41409)) at the Clone object level, correct?
> 
> That wasn't my concern, I was suggesting abstracting the `DataProcessing` object so that it could be attached to any object using a `data_processing_id` field. All such objects, whether they pertained to a `Sample` or to a `Clone` or anything else, would have the same schema, which would be flexible and not require fields describing clone-processing-specific information, for example.

OK, gotcha, was getting the two DataProcessing discussions confused 8-)

I think abstracting the DataProcessing is a great idea but I think it is a bit of a complex topic... I think we want to do it well, and doing it well may not coincide with a quick release of the initial clone spec... Do we need the two to go hand-in-hand? 

Did you have an idea of what the DataProcessing abstraction might look like???

eharkins:
> I think abstracting the DataProcessing is a great idea but I think it is a bit of a complex topic... I think we want to do it well, and doing it well may not coincide with a quick release of the initial clone spec... Do we need the two to go hand-in-hand?

No, I don't think they need to go hand in hand.

> Did you have an idea of what the DataProcessing abstraction might look like???

I had not thought much about it but was planning to err on the side of flexibility so that it could easily accommodate any type of data processing info. The extreme of that would be to not have any required fields at all in `DataProcessing`, and just use it as a place to stick pertinent info in any format. This would be rather open-ended and leave any parsing of `DataProcessing` up to the application as opposed to the schema, but it would allow us to move forward more quickly with an initial release.

If that is too extreme perhaps we could settle on a few open ended required fields for any `DataProcessing` that were broadly applicable. E.g. things like `software` and `version`.

@bcorrie (and others): thoughts on this?

specs/airr-schema.yaml
scharch: nope all good

scharch:
The (slight) problem here is that we probably want to distinguish between the two usages. One refers to a `data_processing` object that is really part of the `Repertoire`, with the id at the `Clone` level just for disambiguation as @bcorrie points out. The other is `data_processing` object (the one that @eharkins is abstracting) has a primary reference to the `Clone` itself, describing interesting things one might have done to that specific clone (eg tree building). I'm not sure how to do this without getting kludgy, though.

scharch:
@bcorrie related to the disambiguation question - do we need to include `study_id` also, or we assuming globally unique `repertoire_id`s?

javh:
A clone isn't necessarily going to be restricted to a single sample, so I don't think `sample_processes_id` being a member of `Clone` makes sense. Yes? No?

IIRC, we decided to punt on the problems with (and abstraction of) `DataProcessing`, including `data_processing_id` as a link for now and just assuming `DataProcessing` can handle clones, trees, etc. Sorting out `DataProcessing` seems tied to the larger issue of Repertoire grouping and might be a semi-large task...

bcorrie:
> @bcorrie related to the disambiguation question - do we need to include `study_id` also, or we assuming globally unique `repertoire_id`s?

We assume repertoire_id's are unique at the repository level for sure, with a rather long discussion, still ongoing, about the uniqueness of repertoire_id globally (see https://github.com/airr-community/airr-standards/issues/246#issuecomment-557918355).

scharch:
> > @bcorrie related to the disambiguation question - do we need to include `study_id` also, or we assuming globally unique `repertoire_id`s?
> 
> We assume repertoire_id's are unique at the repository level for sure, with a rather long discussion, still ongoing, about the uniqueness of repertoire_id globally (see [#246 (comment)](https://github.com/airr-community/airr-standards/issues/246#issuecomment-557918355)).

I remember that conversation... :) 
For present practical purposes, unique-in-repository is enough that we don't need to include `study_id`, right?

scharch:
> A clone isn't necessarily going to be restricted to a single sample, so I don't think `sample_processes_id` being a member of `Clone` makes sense. Yes? No?

Yes, but actually doesn't the same logic apply to `data_processing`, as well? This makes me think more: @bcorrie a `sequence_id` is unique within a `Repertoire` even if there are multiple `data_processing`s applied to the same raw reads, right? If so, maybe it would be better to access this information through the `sequence_id`s of the `Node`s instead of from the `Clone` directly?

scharch:
> IIRC, we decided to punt on the problems with (and abstraction of) `DataProcessing`. For now, including `data_processing_id` as a link and just assuming `DataProcessing` can handle clones, trees, etc. Sorting out `DataProcessing` seems tied to the larger issue of Repertoire grouping and might be a semi-large task...

@javh, I think this is right, but it doesn't seem compatible with @bcorrie's proposed change of the description of `data_processing_id` to "Identifier of the data processing object _in the repertoire metadata for this clone_" because it might not be (only) that. I understand why @bcorrie wants to have this specifically, but the better answer might be going through `sequence_id` as I proposed above and reverting the description of `data_processing_id` here to the more generic "punted" version.

javh:
> Yes, but actually doesn't the same logic apply to `data_processing`, as well?

Yeah, it does. Hence the punt, because we don't really have a `DataProcessing` object that can represent clone/tree tasks on a collection of Repertoires. But, we can have a skeletal `DataProcessing` object with just `data_processing_id`, `software_versions`, and some as-of-yet-unknown custom fields for clone/trees which is associated with a single `Repertoire`.

schristley:
> @schristley was that more along the lines of what you were originally thinking?

I've been thinking a bit about `DataProcessing`, specifically how we might generalize it. The one solution I was exploring is what has been suggested a few times, that is, to use the multiplicity (array) of `DataProcessing` to describe the different steps. Then, the simplest would be to attach tags to them like `clone assignment`, `vdj assignment`, etc. With a well-established set of tags, you could retrieve the specific processing you want.

However, there is a problem with this design, and it happens downstream based upon the policy this encourages. Specifically, it will encourage users to use multiple `DataProcessing` to describe different parts/steps of a single processing pipeline/workflow. As soon as they start doing that, people will need to answer questions like, which steps go together? what is the ordering of the steps, and so on? Before you know it, we have to design a workflow language that describes how all of those different steps go together. This is not what we (well me at least) want to do...

Therefore, I feel we should stick with the original design of `DataProcessing` which envisions a single object that covers the whole process.

Here we have two choices, within that single `DataProcessing` object, we can add additional fields for annotating clonal assignment, lineage trees, and other stuff.

The other choice is to not overload `DataProcessing` too much, but add directly to `Repertoire`. That is, we add a separate `clone_processsing` array as a sibling to `data_processing`. This seems okay to me because it is an immune repertoire specific concept, but we might get into the situation where we annotate more and more types of processing, and this becomes unwieldy. Then, it may be better to encapsulate everything in `DataProcessing` so `Repertoire` doesn't become polluted.

javh:
Because generalizing DataProcessing seems like a big topic that's somewhat tangential, I spawned a new issue from @schristley's comment: #313.

bcorrie:
> > > @bcorrie related to the disambiguation question - do we need to include `study_id` also, or we assuming globally unique `repertoire_id`s?
> > 
> > 
> > We assume repertoire_id's are unique at the repository level for sure, with a rather long discussion, still ongoing, about the uniqueness of repertoire_id globally (see [#246 (comment)](https://github.com/airr-community/airr-standards/issues/246#issuecomment-557918355)).
> 
> I remember that conversation... :)
> For present practical purposes, unique-in-repository is enough that we don't need to include `study_id`, right?

Yes, I believe so...

eharkins:
Thanks @javh. For the purposes of this minimal schema version, is this
> we can have a skeletal DataProcessing object for clones/trees which is associated with a single Repertoire that just has data_processing_id, software_versions, and some to-be-determined custom fields.

acceptable? If so, @javh can you commit something like that for us to iterate on since it seems like you had an idea of where to start? Otherwise I can add something simple to get the ball rolling and we can make any necessary corrections.

javh:
> acceptable? If so, @javh can you commit something like that for us to iterate on since it seems like you had an idea of where to start? Otherwise I can add something simple to get the ball rolling and we can make any necessary corrections.

I don't think we need to do anything for this. 

If we're agreed, then I think it's fine as is with `data_processing_id` in `Clone`. We can handle the internals of `DataProcessing` separately and restrict this PR to `Clone`, `Tree` and `Node`. Maybe we can simplify the description of `data_processing_id` in `Clone` as @scharch mentioned above.

eharkins:
> > IIRC, we decided to punt on the problems with (and abstraction of) DataProcessing. For now, including data_processing_id as a link and just assuming DataProcessing can handle clones, trees, etc. Sorting out DataProcessing seems tied to the larger issue of Repertoire grouping and might be a semi-large task...

> @javh, I think this is right, but it doesn't seem compatible with @bcorrie's proposed change of the description of data_processing_id to "Identifier of the data processing object in the repertoire metadata for this clone" because it might not be (only) that. I understand why @bcorrie wants to have this specifically, but the better answer might be going through sequence_id as I proposed above and reverting the description of data_processing_id here to the more generic "punted" version.

@scharch would 

> reverting the description of data_processing_id here to the more generic "punted" version.

mean changing it back to `Identifier for the data processing object.`? I can make this change if everyone agrees with that.

scharch:
> @scharch would
> 
> > reverting the description of data_processing_id here to the more generic "punted" version.
> 
> mean changing it back to `Identifier for the data processing object.`? I can make this change if everyone agrees with that.

I think completely reverting https://github.com/airr-community/airr-standards/commit/73166b05401b73b072bdb33ad2023362c2b41409 tbh.

eharkins:
Makes sense, shall I go ahead with this @bcorrie / others?

We can reference that commit in #313 or duplicate @bcorrie's comment on it to preserve that effort for the issue focused on `DataProcessing`

bcorrie:
> Makes sense, shall I go ahead with this @bcorrie / others?
> 
> We can reference that commit in #313 or duplicate @bcorrie's comment on it to preserve that effort for the issue focused on `DataProcessing`

I am not sure that will work... 8-)

A data_processing_id is not unique (at least currently in our spec/docs). It is unique within a Repertoire only. To identify a specific ```DataProcessing``` entity, you need the ```reperotire_id``` of the ```Repertoire``` and the ```data_processing_id``` of the ```DataProcessing``` entity you are referring to within that ```Repertoire```. A ```data_processing_id``` without a ```repertoire_id``` is ambiguous and is not guaranteed to be resolvable...

From here: http://docs.airr-community.org/en/latest/datarep/metadata.html

```
If a Repertoire has multiple DataProcessing then data_processing_id should be used to distinguish the appropriate DataProcessing within the Repertoire. The Rearrangements contains data_processing_id for this purpose. The data_processing_id is only unique within a Repertoire so repertoire_id should first be used to get the appropriate Repertoire object and then data_processing_id used to acquire the appropriate DataProcessing.
```

specs/airr-schema.yaml
schristley: The issue with `sequence_id` is it is only unique within a "file" and may not be persistent, so we should also allow `rearrangement_id` to be used.

scharch:
> I am not sure that will work... 8-)
> 
> A data_processing_id is not unique (at least currently in our spec/docs). It is unique within a Repertoire only. To identify a specific `DataProcessing` entity, you need the `reperotire_id` of the `Repertoire` and the `data_processing_id` of the `DataProcessing` entity you are referring to within that `Repertoire`. A `data_processing_id` without a `repertoire_id` is ambiguous and is not guaranteed to be resolvable...

I hear you @bcorrie, but it seems like a minor enough issue that we can let it stand for now, and it will get resolved naturally  as part of the introduction of a `CloneRepertoire`-type object and/or the evolution of `DataProcessing` itself.

specs/airr-schema.yaml
scharch: But `sequence_id` is required, while `rearrangement_id` is not...
I could see using `rearrangement_id` as a kind of future-proofing, though, since it will be important if identifying `Clone`s across multiple `Repertoire`s.

eharkins:
It does seem true that linking a `DataProcessing` to a `Clone` will depend in part on the `Clone`s relationship with a `Repertoire`, and since we have officially deemed resolving the Clone-Repertoire relationship to be outside the scope of this version of the clone-tree schema, I am in favor of allowing `DataProcessing` to be fully resolved in a later version / tackled in a separate effort (#313). 

If that is alright with you @bcorrie? Are there amendments you'd want to make still for this version if we revert 73166b0 as @scharch suggests?

e.g. noting in the description that :
> A data_processing_id without a repertoire_id is ambiguous and is not guaranteed to be resolvable

bcorrie:
> If that is alright with you @bcorrie? Are there amendments you'd want to make still for this version if we revert [73166b0](https://github.com/airr-community/airr-standards/commit/73166b05401b73b072bdb33ad2023362c2b41409) as @scharch suggests?

There is still the addition of sample_processing_id in that commit, but....

Since the main purpose of repertoire_id, data_processing_id, and sample_processing_id is to link a ```Clone``` back to a ```Repertoire``` and we are deferring the resolution of that to the ```CloneRepertoire``` discussion, I think this is probably OK for now...

In the interests of having something, we can probably go with this for now... I guess maybe I would prefer to have sample_processing_id added with the same terminology description, even though that may make less sense. At least that way a ```Clone``` can be made resolvable to a specific ```SampleProcessing``` and ```DataProcessing``` within a ```Repertoire``` if desired?

I am OK either way for now though, as long as we go back and resolve this 8-)

javh:
`repertoire_id` is currently a required field in `Clone`, so we could leave the `data_processing_id` as is (with the repertoire metadata references). The `data_processing_id` for the `Clone` may actually be in the undefined grouped repertoire object instead in the final version. But, we don't have that yet, so maybe it doesn't makes a difference for the single Repertoire test case we're starting with.

A `Clone` having multiple samples is a pretty common use case, so we're not going to be able to link to single `SampleProcessing` object in `Clone`. I'm inclined to leave `sample_processing_id` out for now and come back to it later.

I just pushed some suggested changes:

* Leave `data_processing_id` description unchanged, but set `required: false`.
* Remove `sample_processing_id` from `Clone`.
* Remove `junction` and `junction_aa` from `Clone`. (Clone's won't have a single junction sequence... I can't recall if this was originally included due to some discussion).
* Rename `node_list` to `nodes`.
* Rename `newick_string` to `newick`.
* Deleted `specs/immunedb-clones.yaml` and `olmsted-airr-schema.yaml`.

That's all I see that I personally have to comment on. So, I'm good to go with this for an experimental release.

I think there's a good argument for making `v_call`, `d_call` and `j_call` required fields in `Clone`, but I didn't change that, because there are situations were every sequence in a clone doesn't have the same alignment results (even if this isn't biologically realistic). We could reword this as a consensus V(D)J, but not every tool does that.

scharch:
>     * Remove `junction` and `junction_aa` from `Clone`. (Clone's won't have a single junction sequence... I can't recall if this was originally included due to some discussion).

Yes: they (and the `*_call` fields) were specifically defined as being in reference to the naive ancestor `Clone.germline_alignment`. If the member sequences have different assignments/calls/whatever, those should be accessed via the original `Rearrangement` TSV (which is why `sequence_id` is supposed to match) or included as a custom field under `Node`.

eharkins:
> repertoire_id is currently a required field in Clone, so we could leave the data_processing_id as is (with the repertoire metadata reference). The data_processing_id for the Clone may actually be in the undefined grouped repertoire object instead in the final version. But, we don't have that yet, so maybe it doesn't makes a difference for the single Repertoire test case we're starting with.

> A Clone having multiple samples is a pretty common use case, so we're not going to be able to link to single SampleProcessing object in Clone. I'm inclined to leave sample_processing_id out for now and come back to it later.

@bcorrie does this ^  work for now?

> >  * Remove `junction` and `junction_aa` from `Clone`. (Clone's won't have a single junction sequence... I can't recall if this was originally included due to some discussion).

> Yes: they (and the *_call fields) were specifically defined as being in reference to the naive ancestor Clone.germline_alignment. If the member sequences have different assignments/calls/whatever, those should be accessed via the original Rearrangement TSV (which is why sequence_id is supposed to match) or included as a custom field under Node.

@scharch so do we need to keep the junction and *_call fields and make clear in their definitions that they are 
>  in reference to the naive ancestor Clone.germline_alignment

? Maybe that was already in their definitions but I couldn't find it.

@javh if we are sticking with not having the junction fields you removed, should we also remove the corresponding fields like `junction_length`, `junction_aa_length`?

javh:
> Yes: they (and the `*_call` fields) were specifically defined as being in reference to the naive ancestor `Clone.germline_alignment`.

Ah, my bad then. I'll put them back in and push it.

javh:
Okay, I added `junction` and `junction_aa` back in and added wording to the junction and `*_call` fields referencing the inferred ancestor.

bcorrie:
> > A Clone having multiple samples is a pretty common use case, so we're not going to be able to link to single SampleProcessing object in Clone. I'm inclined to leave sample_processing_id out for now and come back to it later.
> 
> @bcorrie does this ^ work for now?

Yes, I think so... I think this indicates pretty strongly that we do need a CloneRepertoire that is different than a Repertoire, at least if I understand correctly - but that is a different conversation...

eharkins:
Should I merge this into the `clones-and-trees` branch then?

I have an AIRR Data Rep WG call on my calendar for Monday, if that is accurate then we can discuss next steps for including in the 1.3 release during the call?

javh:
No objections to merging from me.

Yes, we have our first call of the year on Monday. I'll look over what we have and send out proposed topics this weekend - I suspect it'll be dominated by v1.3 release topics.

I think the only thing extra we'll need for the v1.3 release is a skeletal docs page, but we can take care of that separately. That's all that occurs to me. We'll see what everyone else thinks.  I'm hoping we can just merge all the way into master after the call.

PR #299, opened 2019-12-17
Extend the Junction AA description

bcorrie:
The Junction NT description notes that the junction is the CDR3 plus the conserved codons. The Junction AA description should have a similar definition.

Simple change, should be able to merge right away...

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1597,8 +1597,7 @@ Rearrangement:
         junction_aa:
             type: string
             description: >
-                Junction region amino acid sequence, where the junction is defined as
-                the CDR3 plus the two flanking conserved codons.
+                Amino acid translation of the junction.
             example: CARAGVYDGYTMDYW
             x-airr:
                 miairr: true
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1545,7 +1545,8 @@ Rearrangement:
         junction_aa:
             type: string
             description: >
-                Junction region amino acid sequence.
+                Junction region amino acid sequence, where the junction is defined as
+                the CDR3 plus the two flanking conserved codons.
             example: CARAGVYDGYTMDYW
             x-airr:
                 miairr: true
--- a/lang/python/airr/specs/airr-schema.yaml
+++ b/lang/python/airr/specs/airr-schema.yaml
@@ -1597,8 +1597,7 @@ Rearrangement:
         junction_aa:
             type: string
             description: >
-                Junction region amino acid sequence, where the junction is defined as
-                the CDR3 plus the two flanking conserved codons.
+                Amino acid translation of the junction.
             example: CARAGVYDGYTMDYW
             x-airr:
                 miairr: true
--- a/lang/python/airr/specs/airr-schema.yaml
+++ b/lang/python/airr/specs/airr-schema.yaml
@@ -1545,7 +1545,8 @@ Rearrangement:
         junction_aa:
             type: string
             description: >
-                Junction region amino acid sequence.
+                Junction region amino acid sequence, where the junction is defined as
+                the CDR3 plus the two flanking conserved codons.
             example: CARAGVYDGYTMDYW
             x-airr:
                 miairr: true
--- a/specs/airr-schema.yaml
+++ b/specs/airr-schema.yaml
@@ -1597,8 +1597,7 @@ Rearrangement:
         junction_aa:
             type: string
             description: >
-                Junction region amino acid sequence, where the junction is defined as
-                the CDR3 plus the two flanking conserved codons.
+                Amino acid translation of the junction.
             example: CARAGVYDGYTMDYW
             x-airr:
                 miairr: true
--- a/specs/airr-schema.yaml
+++ b/specs/airr-schema.yaml
@@ -1545,7 +1545,8 @@ Rearrangement:
         junction_aa:
             type: string
             description: >
-                Junction region amino acid sequence.
+                Junction region amino acid sequence, where the junction is defined as
+                the CDR3 plus the two flanking conserved codons.
             example: CARAGVYDGYTMDYW
             x-airr:
                 miairr: true


javh:
Hrm. The rest of the `*_aa` fields just read as references to the base nucleotide field, which has a detailed-ish description, so it would be more consistent to change the description to something like:

"Amino acid translation of the junction field."

bcorrie:
> Hrm. The rest of the `*_aa` fields just read as references to the base nucleotide field, which has a detailed-ish description, so it would be more consistent to change the description to something like:
> 
> "Amino acid translation of the junction field."

Hrm back 8-)

Is that because that was the easy path for us to define the fields or because that is the right definition? I think it makes more sense for each field to be defined properly in and of itself, not in terms of the definition of another field in the spec... If a derivation is likely, it could be mentioned, but I am not sure that that is the right description in general.

My rationale and use case is this... We use the AIRR spec to define an awful lot of what is in our user interface... We don't want to define terms, we want to use the AIRR spec descriptions for those terms. So for example, in our user interface, a table with junction_aa as column, the column uses the name in the spec as the column name (```IMGT-JUNCTION amino acid sequence```) and the description of the field as the help text for that column (currently ```Junction region amino acid sequence.```). If we used the description you suggest we would have help text that read ```Amino acid translation of the junction field.```

I would much prefer the description for the field to complete in and of itself, so it would read ```Junction region amino acid sequence, where the junction is defined as the CDR3 plus the two flanking conserved codons.```

The AIRR Community, through the spec, thinks it is important enough to note that the Junction contains the CDR3. I think the spec is more complete if it does that for both fields...

bcorrie:
I suppose what I am suggesting is that all the other fields might need to be changed as well - but I think the description for CDR3 and Junction are particularly important - as people can get confused in particular when they don't see the word "CDR3" in the user interface when using a Junction search element...

bcorrie:
I guess I am also saying that yes, I will do that work if we think it is the right thing to do 8-)

bcorrie:
> That said, I don't recall why we did it that way originally. It was probably just less typing.

That is what I suspect... But I am happy to go through them, as this is quite important for us and I think for completeness it makes sense (and a spec should be complete IMHO 8-)...

javh:
> That is what I suspect... But I am happy to go through them, as this is quite important for us and I think for completeness it makes sense (and a spec should be complete IMHO 8-)...

Okay. I suggest we resolve and merge in #295 first. Modifying all the `*_aa` field descriptions may generate some annoying merge conflicts in the spec if we don't merge that one first.

javh:
From the call:
* Few opinions, but we prefer to leave the `*_aa` as simple translation references to the nucleotide fields.
* Maybe tooltips can be facilitated by stripping `_aa` to get the text of the "parent" field or by adding a new x-airr field.

bcorrie:
> From the call:
> 
> * Few opinions, but we prefer to leave the `*_aa` as simple translation references to the nucleotide fields.

So the opinion was that we keep the junction_aa field description as:

```Amino acid translation of the junction field```

rather than

```Junction region amino acid sequence, where the junction is defined as the CDR3 plus the two flanking conserved codons. This is typically an amino acid translation of the junction field```

What is the downside of having the more complete definition (the later definition)? The downside of having it incomplete is that if someone takes the definition of a field on its own, the top description above isn't very helpful. This seems like the wrong approach to take for a specification.

The only downside I can see of using the more complete definition is that we are duplicating text. I recognize that this is potentially error prone and a pain, but in my opinion we want to have a more complete definition for the sake of clarity. For me, that is what a spec is for...

Even for the longest _aa translation (```v_germ_alignment_aa```) this is not particularly onerous...

scharch:
> What is the downside of having the more complete definition (the later definition)?

The downside is that it's (arguably) incorrect! These are derived/secondary fields, so at the very least "typically" is definitely incorrect. Our feeling is that the first definition you quoted is, in fact, the exact technical definition, which is why we prefer it, despite potential usability concerns.

javh:
Yeah, the concerns were:

1. Maintenance. Having to keep field descriptions in sync is a manual, error prone process.
2. Obfuscating the requirement for the `*_aa` fields to be derivative of the corresponding nucleotide fields.

(2) Could be addressed by changing the definition you have above to:

> Junction region amino acid sequence, where the junction is defined as the CDR3 plus the two flanking conserved codons. This must be an amino acid translation of the junction field.

(1) Is a trade-off I don't see a way around. Though, as this is a standard, hopefully we rarely update these descriptions. If this was software, I'd be much more concerned about this.

My opinion isn't very strong. I'm more of a lean towards keeping it as is. I see the usability concern, so if that's a major blocker, then I personally don't mind replicating the descriptions.

bcorrie:
> > What is the downside of having the more complete definition (the later definition)?
> 
> The downside is that it's (arguably) incorrect! These are derived/secondary fields, so at the very least "typically" is definitely incorrect. Our feeling is that the first definition you quoted is, in fact, the exact technical definition, which is why we prefer it, despite potential usability concerns.

I used typically in that instance on purpose because in the spec it is not necessary to provide the ```junction``` field. So you could have a valid AIRR TSV file that has a ```junction_aa``` defined as a nucleotide translation of the ```junction``` field without a ```junction``` in the file. So typically meant that it should be the translation of the ```junction``` field if it exists in the file.

I accept that ```junction_aa``` is the nucleotide translation of the "junction sequence", but for a given file it is also not correct to say that it is translation of the ```junction``` field as that field may not exist in that file.

javh:
`junction` is required, so I would say if you have an empty string for `junction`, but an amino acid sequence in `junction_aa` then that file is malformed. Maybe not technically, as everything is nullable, but in spirit at least.

bcorrie:
> `junction` is required, so I would say if you have an empty string for `junction`, but an amino acid sequence in `junction_aa` then that file is malformed. Maybe not technically, as everything is nullable, but in spirit at least.

Interesting... 8-)

What about this scenario. I am a researcher with some data, and I annotate my data using MiXCR. I ask for AA CDR3 features only, because I don't need the nt sequences for these regions for my analysis. I correctly note that if I store the nt sequences they increase the size of my data set by a factor of 3!

Based on the above, if I output this data to an AIRR TSV file with ```junction_aa``` but no ```junction``` this is not a valid AIRR TSV file (in spirit at least)? Is that we want?

8-)

javh:
> Based on the above, if I output this data to an AIRR TSV file with junction_aa but no junction this is not a valid AIRR TSV file (in spirit at least)? Is that we want?

Yeah, that's what we want, because that file has been stripped of valuable information that is difficult to derive without realigning it. Similarly to having an empty `germline_alignment` field, which is not uncommon, but something we want to discourage as it presents significant obstacles to analysis of the data.

Now, if that researcher had instead chosen to remove `junction_aa`, there would have been no loss of data.

(This reminds me that we should ping the MiXCR folks about the status of AIRR Rearrangement support...)

scharch:
> Is that we want?

Yes. A lot of analysis can only be done in nucleotide space and the point is to encourage reusability. Besides, `sequence`, `sequence_alignment`, and `germline_alignment` are all required nt sequences, so your storage savings wouldn't actually be very big.

scharch:
> (This reminds me that we should ping the MiXCR folks about the status of AIRR Rearrangement support...)

Software WG has been talking with them about it, is coming....

PR #302, opened 2019-12-18
Updates based on discussion

scharch:
Most of the changes are based on the consensus of discussions with @eharkins and @javh. A few other minor things:
* I added `repertoire_id` and `data_processing_id` as required in `Clone`
* Do we need explicit `required: false` tags the way we seem to be doing for `miairr`?
* Is list a valid `type` for `sequence_id` in `Clone`?

### DIFF
--- a/specs/airr-schema.yaml
+++ b/specs/airr-schema.yaml
@@ -2274,7 +2274,7 @@ Clone:
                 miairr: false
                 required: true
         sequence_id:
-            type: list
+            type: array
             description: List of sequence_ids that are members of the clone.
             x-airr:
                 miairr: false
--- a/specs/airr-schema.yaml
+++ b/specs/airr-schema.yaml
@@ -2251,9 +2251,8 @@ Clone:
     type: object
     required:
         - clone_id
-        - v_call
-        - d_call
-        - j_call
+        - repertoire_id
+        - data_processing_id
         - germline_alignment
     properties:
         clone_id:
@@ -2267,9 +2266,16 @@ Clone:
             description: Identifier to the associated repertoire in study metadata.
             x-airr:
                 miairr: false
+                required: true
         data_processing_id:
             type: string
             description: Identifier for the data processing object.
+            x-airr:
+                miairr: false
+                required: true
+        sequence_id:
+            type: list
+            description: List of sequence_ids that are members of the clone.
             x-airr:
                 miairr: false
         v_call:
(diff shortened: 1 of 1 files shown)


scharch:
@eharkins @javh I don't have write permissions.... TIA

eharkins:
Thanks @scharch! Not sure but I assume I don't have the authority to give you permissions to the airr-standards repo if that's what you mean?

> I added repertoire_id and data_processing_id as required in Clone

👍 
> Do we need explicit required: false tags the way we seem to be doing for miairr?

I didn't add any just to be consistent with existing `x-airr` fields in this schema. Up to the AIRR group but seems like it just depends on whether the schema is validating that the `required` field exists or not. 
> Is list a valid type for sequence_id in Clone

I think it should be [array](https://cswr.github.io/JsonSchema/spec/arrays/) for json schema.

scharch:
> Not sure but I assume I don't have the authority to give you permissions to the airr-standards repo

@eharkins no, I just meant I can't commit directly to your PR, so this is a PR on your PR that you have to merge in

eharkins:
@scharch got it. Let's see what @javh  and others have to say and then I'll gladly merge. Thanks for putting it together.

specs/airr-schema.yaml
```diff
@@ -2267,9 +2266,16 @@ Clone:
...
                 miairr: false
+                required: true
         data_processing_id:
             type: string
             description: Identifier for the data processing object.
+            x-airr:
+                miairr: false
+                required: true
+        sequence_id:
+            type: list
```
eharkins: @scharch can you change this to be [array as in this link](https://cswr.github.io/JsonSchema/spec/arrays/) and then I can merge?

scharch: @eharkins done :)

eharkins: @scharch thanks! I'll merge now. Should we add the type of items in the array like they do in the docs I linked:
```
 "items": {
           "type": "string"
          }
```
(or other restrictions on what the array can contain)?

scharch: > Should we add the type of items in the array like they do in the docs I linked:

dunno, sounds like a question for @schristley on what our typical format/standard for such things is...

eharkins: The jsonschema array docs suggest you might be required to specify the `items` definition, and regardless it seems like a good idea if we know the expected type of things in the array as we probably do with `sequence_id` here. Also most of the array fields in the existing AIRR schemas have an `items` definition [e.g.](https://github.com/airr-community/airr-standards/blob/master/specs/airr-schema.yaml#L1290)

I can add one directly to the merged `clones-and-trees-release` branch.

PR #315, opened 2020-01-13
Fields for IMGT support

### LINKED ISSUE #268: IMGT Support for AIRR Rearrangement Schema

### LINKED ISSUE #309: Tell partial sequences right away from rearrangement schema
If we define whole/complete/full sequences as those spanning the first base of the (germline) V gene until the last base of the (germline) J gene [as per https://github.com/airr-community/airr-standards/issues/11], how to tell right away whether sequences are full length from the [rearrangement schema](http://docs.airr-community.org/en/stable/datarep/rearrangements.html#fields)? One can spot right away sequences with partial 5' end by looking for `v_germline_start` > 1. However, it is harder to do something similar on the 3' end (looking at `j_germline_end`) since the length of different germline J gene varies and is not immediately available.

The use case I have in mind is submissions to IgBlast, with output AIRR format. I recall from long time ago that partial submission to IMGT V-quest do get flagged as partial 5' or partial 3' with a clear label. 

Am I missing a way to easily infer partial 3' from the rearrangement schema? Would there be value in adding two boolean columns for partial 5' and partial 3' sequence flagging?

javh:
This is a clone of #295, with the following changes:

* Removal of v/d/j fields `*_subgroup`, `*_gene` and `*_allele`, which have been pushed back to v2.0.
* Incorporation of additional gene nomenclature text from @williamdlees in Rearrangement docs.
* An attempt to resolve the impasse on the inclusion of examples in the existing `*_call` fields.

The main purpose here is to resolve #268, adding the extra fields to Rearrangement need to properly support IMGT adding Rearrangement support to IMGT/V-QUEST.

I think we can also resolve #309, but I didn't add those two fields yet, as I think they need discussion first.

### DIFF
--- a/docs/api/adc_api.rst
+++ b/docs/api/adc_api.rst
@@ -758,7 +758,7 @@ in the AIRR Schema.
       - Description
     * - rearrangement_id, repertoire_id, sample_processing_id, data_processing_id, clone_id, cell_id, pair_id
       - Identifiers; rearrangement_id allows for query of that specific rearrangement object in the repository, while repertoire_id, sample_processing_id, and data_processing_id are links to the repertoire metadata for the rearrangement. The clone_id, cell_id, and pair_id are all identifiers that group rearrangements based on clone definition, single cell assignment, and paired chain linking.
-    * - locus, v_call, d_call, j_call, c_call, productive, junction_aa, junction_aa_length
+    * - locus, v_call, d_call, j_call, c_call, v_gene, d_gene, j_gene, c_gene, v_subgroup, d_subgroup, j_subgroup, c_subgroup, productive, junction_aa, junction_aa_length
       - Commonly used rearrangement annotations.
 
 **Repertoire/rearrangement object size**
--- a/docs/api/adc_api.rst
+++ b/docs/api/adc_api.rst
@@ -758,7 +758,7 @@ in the AIRR Schema.
       - Description
     * - rearrangement_id, repertoire_id, sample_processing_id, data_processing_id, clone_id, cell_id, pair_id
       - Identifiers; rearrangement_id allows for query of that specific rearrangement object in the repository, while repertoire_id, sample_processing_id, and data_processing_id are links to the repertoire metadata for the rearrangement. The clone_id, cell_id, and pair_id are all identifiers that group rearrangements based on clone definition, single cell assignment, and paired chain linking.
-    * - locus, v_call, d_call, j_call, c_call, v_gene, d_gene, j_gene, c_gene, v_subgroup, d_subgroup, j_subgroup, c_subgroup, productive, junction_aa, junction_aa_length
+    * - locus, v_call, d_call, j_call, c_call, productive, junction_aa, junction_aa_length
       - Commonly used rearrangement annotations.
 
 **Repertoire/rearrangement object size**
--- a/docs/datarep/alignments.rst
+++ b/docs/datarep/alignments.rst
@@ -4,7 +4,7 @@ Alignment Schema (Experimental)
 ===============================
 
 An Alignment is the output from a V(D)J assignment process for a
-single V, D, J, or C gene segment for a sequence. It is not necessary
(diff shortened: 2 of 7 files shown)


javh:
@bussec, I'm content to include whatever consensus language you and @williamdlees (and others) can arrive at, even if that's just deleting the expanded text.

From the DataRep perspective, we need to preserve the first bit of text about there being no actual requirements for what appears in the gene call fields. The rest is just recommendations.

As for gene vs segment, I vaguely recall a big discussion on this a long time ago and we settled on using "gene" for any reference to a germline gene and "segment" for any reference to subregion in an alignment. But... I'd have to go back through the notes and see if that's the case.

javh:
From the call:
* Reword the existing "segment" references to "gene" in the schema.
* For #309 look at NucleicAcidProcessing.complete_sequences, though the semantics look wrong. A single boolean specifying whether the alignment is full length may be sufficient. (`complete_vdj`).

williamdlees:
I suggest that we delete lines 80-90 in the documentation.

For info, in IARC/OGRDB we use 'sequence' for anything that isn't a reference gene, and 'sequence type' to differentiate between V,D,J. Over time we expect to be handing inferences for constant regions, leader, and maybe other things, which would mitigate against using 'segment' in our context. If you're just looking at differentiating V D and J, 'segment' seems very reasonable. To my knowledge there isn't a suitable term in the IMGT lexicon.

The remaining issue we have, from my point of view, is that in the examples you use a compound name for the comonents of a gene name, whereas in the germline schema and OGRDB we separate it into parts, so that an allele name would be, say, IGHV1-69*01 in the proposed scheme but is 01 in OGRDB currently. I can't claim that it is a vast amount of work to change OGRDB in this respect, but it does represent some work, and I think we should determine whether the expected benefit is worth it.

Perhaps we also need to consider how we can smooth the path going forward. I'm conscious that I have not given the germline schema much love over the last 18 months. Perhaps we could find a volunteer who would be prepared to take it on board and also take on the wider job of determining and rectifying terminology across all of the places it is used - working with GLDB and Data Rep.

javh:
I deleted that extra nomenclature text (lines 80-90) from the Rearrangement docs.

I expunged "segment" from the schema and the docs, I replaced it with either "gene" or "gene alignment", as appropriate. With the exceptions of:
* The Alignment schema. `segment` is an explicit field, so this would be a large change. Though, as far as I know, this schema never saw any use. So either way... probably not worth messing with it.
* The NCBI submission docs, because this is the terminology used there.

@williamdlees, as far as the compound names issue goes, #295 is still open, so we can tackle how to handle the `*_subgroup`, `*_gene` and `*_allele` fields in that PR separately for v2.0.  This PR would be for v1.3.

I can't see us ever changing the `*_call` fields, those will surely remain as is, supporting compound gene identifiers and identifiers that can't be broken down into parts at all. (They are mapping to the reference sequence the alignment tool assigned as the top hit.)

williamdlees:
I agree that the *_call fields should remain compound.

javh:
If there are no objections from @schristley or @bcorrie, then I'd to merge this in.

scharch:
@javh are you going to use this to close #309?

javh:
No, I'll open a separate PR for #309. Seems like it needs opportunity for independent discussion.

PR #318, opened 2020-01-17
Human population genetics Extension

### LINKED ISSUE #264: Human Population Genetics Extension
Background: MiAIRR currently contains a number of fields that do not necessarily meet the stringent definition for being "minimal", in the sense of being *independent* of the specifics of a study (e.g. human subjects). To increase the usability and acceptance of MiAIRR and prevent potentially erroneous or misleading annotation (for the sake of "standard compliance"), it seems appropriate to move such fields out of the "core" standard and into separate "extensions". These will then be required to be used in a modular fashion if certain conditions are met by a study. This will also solve (some) of the issues related to a single field requiring multiple ontologies, depending on the context.

Proposal: Move the following fields, which are primarily aimed at human populations, into a "Human Population Genetics" extension. The name represents that the involved fields focus on the information regrading the ancestry of a subject. Note that the fields will be **removed** from the MiAIRR v2 Core Standard.

* `ancestry_population`
* `ethnicity`
* `race`
* `country_birth` (defined in #265 )
* `collection_country` (defined in #265  )

### LINKED ISSUE #265: Add country information to subject and sample
Background: Geographical information about the origin of a subject is important information for human population genetics of Ig/TCR loci. However, to be boardly applicable, we should capture information that is frequently available (GPS coordinates are not), allows unambiguous encoding and does not present major challenges for privacy.

Proposal: Add the following two fields into the AIRR Schema, however do **not** include them in MiAIRR (see Note 2 below):

* Field `country_birth` on the `subject` level
   *  Contains the self-reported country of birth of the subject
   *  Encoding: [ISO3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) Alpha-2 
* Field `collection_country` on the `sample` level
   *  Contains country in which the sample was taken
   *  Encoding: ISO3166-1 Alpha-2

Notes:
1. Choice of encoding:
    * ISO3166-1 contains also historical entities and allows for subdivision of federal states via ISO3166-2.
    * Alpha-2 coding is well known, as it is also used for internet top-level domains.
    *  Alternatives:
         a. ISO3166-1 Alpha-3: Less known, increase in human-readability is marginal
         b. ISO3166-1 English short name (ALL CAPS): Manual encoding error-prone, requires implementation as controlled vocabulary.
         c. UN Stats Division code (currently [M49](https://unstats.un.org/unsd/methodology/m49/)): Numerical code, not human-readable
2. Non-MiAIRR status: The two fields are expected to be moved (along with a set of other human-specific fields) into the proposed Human Population Genetics Extension (#264 ), which would remove them (again) from the MiAIRR core specification. Hence it seems appropriate not to include them at this point.

bussec:
Closes #264 
Closes #265

This is the first stab to implement this extension. It will likely require review and documentation of the way how XT are integrated into the schema. Also the GAZ ontology is a draft at this point.

### DIFF
--- a/docs/miairr/metadata_guidelines.rst
+++ b/docs/miairr/metadata_guidelines.rst
@@ -232,3 +232,24 @@ experiments SHOULD be annotated as follows:
       reliably determined.
    *  ``reverse_pcr_primer_target_location``: Locus and position
       according to the respective set of reverse primers.
+
+
+Geolocation information
+-----------------------
+
+For questions regarding population genetics, the information about the
+membership of a given individual in a (potentially) distinct population
+group is of high interest. However, no reasonably complete ontology
+exists for the concept of "ethnicity", which in addition is limited
+to humans (not to mention the challenge of assigning consistent values
+to this property). Therefore, the AIRR Schema uses annotation of the
+location of birth of an individual and the location where a given sample
+was taken as imperfect but at least operationalizable substitute.
+
+The respective properties ``Subject``.``location_birth`` and
+``Sample``.``collection_location`` are ontology controlled and expect
+values that are geographic locations. The provided value SHOULD
+allow to resolve at least the country-level information of the annotated
+location. Higher granularity SHOULD only be provided, if this does not
+exposed the data subject to the risk of being re-identified via based on
+or assisted by this information.
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -332,7 +332,6 @@ Attributes:
                             type: string
                             description: Ontology name for the top node term
 
-
 # AIRR Data File
 #
 # A JSON data file that holds Repertoire metadata, data processing
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -79,6 +79,14 @@ CURIEMap:
         map:
             ENSG:
                 iri_prefix: "https://www.ensembl.org/Multi/Search/Results?q="
+    GAZ:
+        type: ontology
+        default:
+            map: OBO
+            provider: OLS
+        map:
+            OBO:
+                iri_prefix: "http://purl.obolibrary.org/obo/GAZ_"
     IEDB_RECEPTOR:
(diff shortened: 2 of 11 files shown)


williamdlees:
GAZ is certainly exhaustive. Perhaps it is too exhaustive for the purposes of country_birth and collection_country? If these fields are truly country fields (which personally I believe is the correct level of granularity) it would be easier for submitters to engage with ISO-3166. It's also a little concerning that GAZ has not been updated in the last 9 months: are we confident that it will be well supported?

As a gazeteer, it may not be the best choice for ancestry_population. What is the reason for preferring it in this context over HANCESTRO?

A problem that I have found with both GAZ and HANCESTRO is that they are difficult to browse efficiently. Perhaps there are tools that I am not aware of that make them easier to use, but half an hour's searching has not revealed them to me). Searches in the referenced browsers are extremely slow and search terms do not always provide helpful results. For example searching GAZ for the term 'France' provides >600 results. I am worried that using a large ontology may frustrate submitters and/or result in records that are difficult to search meaningfully, unless we provide facilities that make it easy to use the ontologies at reasonably high levels of abstraction.

schristley:
The tests are probably failing for the new repertoire validation. If you add fields, especially `miairr=true`, then need to modify `lang/python/tests/data/good_repertoire.airr.yaml` and `lang/python/airr/specs/blank.airr.yaml` to include those fields.

ssnn-airr:
@schristley  Not sure if related to this issue but I noticed in master `good_repertoire.airr.yaml` has duplicated 'cell_subset' info for the same sample: [good_repertoire.airr.yaml#L49](https://github.com/airr-community/airr-standards/blob/master/lang/python/tests/data/good_repertoire.airr.yaml#L49)

schristley:
> has duplicated 'cell_subset' info...

that's a typo, maybe caused by a merge error, please delete the extra entry.

bussec:
As discussed in one of the calls this spring, the "extension" (XT) mechanism has been deprecated and the respective parts of the schema have been removed.

javh:
From the call:
* Change `country_` to `location_` (consistent with GAZ).
* Update docs to state that country level is preferred.
* Then merge.

Issue #280, opened 2019-11-23
Inconsistent use of array in spec?


bcorrie:
We have a couple of places where arrays are used to contain multiple values for a field. In the spec this is done inconsistently, and when one tries to map this to a query the semantics of the query are not clear.

We currently have:
- keywords_study - which is an array of strings
- pcr_target - which is an array of objects, where each object is a key/value pair with a key value of "pcr_target_locus" and value type of string.

In #279 I suggest changing read_length, which is a string that contains an array like string with two numbers to an array of integers using the "keywords_study" approach.

I got confused when trying to generate a query using either approach, as I don't think the semantics of the queries are clear in either case.

There is an example in the docs for pcr_target_locus:
```
                "op":"=",
                "content": {
                    "field":"sample.pcr_target.pcr_target_locus",
                    "value":"IGH"
                }
```
We are assuming that this means that the query clause evaluates to true if any of the "value" elements in the pcr_target array contain the element IGH??? So this is an "OR" operation for "=" across each array element???

One can assume the same for keywords_study, but in doing so we actually have the same query operating on two different data structures, which is weird.

BTW, in looking at the spec, it is unclear why there is a need for more than one PCR Target. This may be obvious to others, but I am a bit unclear. This might need a clarification in the description for the field?

bussec:
> BTW, in looking at the spec, it is unclear why there is a need for more than one PCR Target. 

You can run multiplex PCRs on multiple loci at the same time. It is not the most frequently used protocol, but technically possible. In this case the resulting libraries would be mixed, which is what we tried to encode in this field.

bcorrie:
> > BTW, in looking at the spec, it is unclear why there is a need for more than one PCR Target.
> 
> You can run multiplex PCRs on multiple loci at the same time. It is not the most frequently used protocol, but technically possible. In this case the resulting libraries would be mixed, which is what we tried to encode in this field.

Ah, lack of clarity - I blame it on working on GitHub issues while on a long flight 8-)

What I meant by this, more than clarifying for me, is that we should probably clarify this in the spec in the description for this field... So someone reading the spec would know what this might be???

schristley:
> We are assuming that this means that the query clause evaluates to true if any of the "value" elements in the pcr_target array contain the element IGH??? So this is an "OR" operation for "=" across each array element???

Yes, that's my understanding of how the query works.

schristley:
[...]

Issue #298, opened 2019-12-17
Functionality of is operator


bcorrie:
In looking at the implementation of the ADC API "is" operator in the ADC API on iReceptor, we have stumbled on an issue... I think this is probably an @schristley question 8-)

In the document, the "is" operator checks to see if a field is "missing". The question is, what is the definition of missing?

This was touched on here: https://github.com/airr-community/airr-standards/issues/239#issuecomment-531391095

The ADC API test for this is here (https://github.com/airr-community/adc-api-tests/blob/master/repertoire/pass-is-op.json):
```
{
  "filters": {
    "op":"is",
    "content": {
      "field":"sample.cell_number"
    }
  }
}
```
This returns a list of repertoires where the sample.cell_number field is missing. My understanding from #239 is missing in this case means that the field **is not in the repository** or the field **is in the repository** but is **null** for a given repertoire.

Put a different way, the above query would return the repertoire where "cell_number" == null

```{"repertoire_id":"1", ... , "sample" : { "cell_number":null, ... }, ... }```

I wanted to confirm, from my understanding of #239, that this interpretation is correct?

**Comment by bcorrie** [comment]
I would like to also revisit the question as to whether we should change these operators from:

"is" to "missing"
"not" to "notmissing"

or something similar... I find these are extremely confusing even when they are well documented 8-)

If I want to query as to whether a field has data in it or not for a given repertoire, the query below:
```
{
  "filters": {
    "op":"missing",
    "content": {
      "field":"sample.cell_number"
    }
  }
}
```
to be much more clear and intuitive than the query:
```
{
  "filters": {
    "op":"is",
    "content": {
      "field":"sample.cell_number"
    }
  }
}
```

schristley:
> means that the field is not in the repository or the field is in the repository but is null for a given repertoire.

Mongo allows you to differentiate these but I don't think the API can, as there is no way to check for a `null` value, is that right? That might be something useful though... Anyways, The GDC [docs](https://docs.gdc.cancer.gov/API/Users_Guide/Search_and_Retrieval/#filters-specifying-the-query) doesn't give much help, so it's up to us to define. If we don't add a way to explicitly check for `null` then I agree with your definition, that it covers both cases.

bzimonja:
Another question then is, what happens in fields that are defined as "N/A" - do they exist as 'not applicable/not available' and if so, is that different from them being 'null'?

bussec:
[...]

Issue #279, opened 2019-11-23
Should read_length be an array of integers?


bcorrie:
Currently read_length is specified as a string of the format [300,300]

Computationally, this isn't very useful. I want to do a search for all read_lengths > 300 and there is no effective way to do this in the sense that the data is a string and could have odd formatting issues...

I suppose this might be a good time to confirm that there should be one field for this, with an array of either one element or two elements depending on if it was paired-read sequencing???

bcorrie:
Hmm, given that we have a keyword based spec, maybe we keep read_length and add read_length_reverse of the reverse read exists, null if it isn't paired-read sequencing? Both as integers.

bcorrie:
Double Hmmmmm. I see that we have another array in the spec, keywords_study. What does a query look like in the ADC API if I want to query for a study that "contains_ig"?

Are the correct semantics:
```
{"op":"=", "content": {"field":"keywords_study", "value":"contains_ig"}}
```
with this evaluating to true if one of the array elements evaluates to true?

bcorrie:
> I suppose this might be a good time to confirm that there should be one field for this, with an array of either one element or two elements depending on if it was paired-read sequencing???

Related question, what is the "correct" or "definitive" way to determine if a SequencingRun uses paired reads given the spec we have now?

bussec:
Introducing a second keyword for the reverse read length should be ok, as the array is not expected to contain more than two entries. This does of course not fix the general handling of JSON style arrays in string-type variables.
The least ambiguous way to distinguish single from paired read setups is `length(read_length)>1`. The information could also be deduced from `sequencing_kit` but as this is free-text it is only human-readable.

schristley:
> Related question, what is the "correct" or "definitive" way to determine if a SequencingRun uses paired reads given the spec we have now?

Well, I don't know about "definitive", but I check if `paired_filename` has a value.

schristley:
I tend to like the idea of having a second field instead of making `read_length` an array.

bcorrie:
Simple change??? Sort out for ADC v1 and AIRR v1.3?

schristley:
> Kind of begs the question as to whether this field should be in the RawSequenceData object rather than the SequencingRun object.

That seems reasonable to me.

> Simple change??? Sort out for ADC v1 and AIRR v1.3?

sure, seems simple enough to me.

bcorrie:
Made changes on a new branch, created a pull request (#324) to finalize the changes...

[...]

PR #322, opened 2020-01-27
Clarify operations on arrays

bcorrie:
As per the discussion in https://github.com/airr-community/airr-standards/issues/280#issuecomment-578860448

```
                "op":"OP",
                "content": {
                    "field":"study.keywords_study",
                    "value":"STRING"
                }
```
The semantics of this query are that for each element i in the array of N keywords_study the correct return value would be an OR operation on the results of N tests keywords_study[i] OP "STRING"

### DIFF
--- a/docs/api/adc_api.rst
+++ b/docs/api/adc_api.rst
@@ -75,11 +75,11 @@ specific functionality as summarized in the following table:
     * - ``/v1/rearrangement/{rearrangement_id}``
       - Retrieve a rearrangement given its ``rearrangement_id``
       - ``GET``
-      - Upon success, returns the ``Rearrangement`` information in JSON or TSV format according to the :ref:`Rearrangement schema <RearrangementSchema>`.
+      - Upon success, returns the ``Rearrangement`` information in JSON format according to the :ref:`Rearrangement schema <RearrangementSchema>`.
     * - ``/v1/rearrangement``
       - Query rearrangements
       - ``POST``
-      - Upon success, returns a list of ``Rearrangements`` in JSON or TSV format according to the :ref:`Rearrangement schema <RearrangementSchema>`.
+      - Upon success, returns a list of ``Rearrangements`` in JSON or AIRR TSV format according to the :ref:`Rearrangement schema <RearrangementSchema>`.
 
 .. __: https://swagger.io/tools/swagger-editor/
 
@@ -569,6 +569,23 @@ The Fields section below describes the available queryable fields.
 The ``value`` operand specifies one or more values when evaluating the
 operator for the ``field`` operand.
 
+*Queries Against Arrays*
+
+A number of fields in the AIRR Data Model are arrays, such as
+``study.keywords_study`` which is an array of strings or
+``subject.diagnosis`` which is an array of ``Diagnosis`` objects. A
+query operator on an array field will apply that operator to each
+entry in the array to decide if the query filter is satisfied. The
+behavior is different for various operators. For operators such as
+``=`` and ``in``, the filter behaves like the Boolean ``OR`` over the
+array entries, that is if **any** array entry evaluates to true then
+the query filter is satisfied. For operators such as ``!=`` and
+``exclude``, the filter behaves like the Boolean ``AND`` over the
+array entries, that is **all** array entries must evaluate to true for
+the query filter to be satisfied.
+
+*Examples*
+
 A simple query with a single operator looks like this:
 
 .. code-block:: json
@@ -616,19 +633,12 @@ A more complex query with multiple operators looks like this:
     }
   }
 
-**Queries Against Arrays**
-
-A number of fields in the AIRR specification are arrays, such as ``study.keywords_study``. 
-A query operator applied to an array field applies that operator to each field and returns
-a boolean ``OR`` of the results of this set of operations. That is, the operator will return
-true if the operation returns true on any of the individual array elements.
-
 **Format Query Parameter**
 
-Specifies the format of the API response. JSON is the default format
-and the only format available for all endpoints except for the
-``rearrangement`` endpoint that accepts AIRR for the :ref:`AIRR TSV
-<FormatSpecification>` format.
+Specifies the format of the API response. ``json`` is the default
+format and is available for all endpoints. The ``rearrangement``
+``POST`` endpoint also accepts ``tsv`` which will provide the data in the
+:ref:`AIRR TSV <FormatSpecification>` format.
 
 **Fields Query Parameter**
 
--- a/docs/api/adc_api.rst
+++ b/docs/api/adc_api.rst
@@ -600,6 +600,13 @@ A more complex query with multiple operators looks like this:
     }
   }
 
+**Queries Against Arrays**
+
+A number of fields in the AIRR specification are arrays, such as ``study.keywords_study``. 
+A query operator applied to an array field applies that operator to each field and returns
+a boolean ``OR`` of the results of this set of operations. That is, the operator will return
+true if the operation returns true on any of the individual array elements.
+
 **Format Query Parameter**
 
 Specifies the format of the API response. JSON is the default format
--- a/docs/examples/queries/query1_rearrangement.json
+++ b/docs/examples/queries/query1_rearrangement.json
@@ -24,5 +24,5 @@
 	]
     },
     "fields":["repertoire_id","rearrangement_id","v_call","productive"],
-    "format":"airr"
+    "format":"tsv"
 }


bcorrie:
Please comment and/or add changes...

bcorrie:
We need to determine if we are happy with these semantics when using the ```in``` and ```exclude``` operator.

study.keywords_study in ["contains_tcr", "contains_ig"] would return true if any of the array elements contain either value.

I don't think this works for operators that perform "negation" operations - not surprising since they are boolean operations I suppose 8-)

study.keywords_study exclude ["contains_tcr", "contains_ig"] would return true if any of the array elements exclude the values, which is probably not the intent...

Consider the array ["contain_single_cell", "contains_tcr"]

In the above scenario, we want to exclude this repertoire because it "contains_tcr" but the operation will perform an exclude of ["contains_tcr", "contains_ig"] on "contain_single_cell" and return true, and therefore the operation will return true...

schristley:
The doc looks good to me. I'll add that `diagnosis`, `sample` and `data_processing` are also arrays.

Regarding `in` and `exclude`, is this how it actually works (that is you tested with queries) or how it might work?

bcorrie:
> Regarding `in` and `exclude`, is this how it actually works (that is you tested with queries) or how it might work?

I was talking strictly semantics based on the definition above... I was more concerned about the logical expectation as to what should happen- haven't looked at any implementation yet...

The question came up when we were checking our implementation and it wasn't clear what should actually happen 8-)

bcorrie:
> The doc looks good to me. I'll add that `diagnosis`, `sample` and `data_processing` are also arrays.

Is it expected that in the ADC API you would be able to perform a query on an object like this??? I have been assuming not... Hence it wasn't necessary to talk about these arrays from a query perspective... 

There is currently one other field array, data_processing_files - but as far as I am aware there are no other fields that one would be able to query on that are arrays (other than the possible discussion in #279)

schristley:
> Is it expected that in the ADC API you would be able to perform a query on an object like this??? I have been assuming not... Hence it wasn't necessary to talk about these arrays from a query perspective...

Yes, like `sample.cell_subset` and all the other sample fields.

schristley:
> study.keywords_study exclude ["contains_tcr", "contains_ig"] would return true if any of the array elements exclude the values, which is probably not the intent... I think the logical intent of this operation would be to exclude the data from this study if any of the keywords were present.

I see. Yes, the description you wrote is not really correct then. The `exclude` operator shouldn't work this way as an `OR`, it should work as the users would want it to, by [excluding documents](https://stackoverflow.com/questions/16221599/find-documents-with-arrays-not-containing-a-document-with-a-particular-field-val) with a specific value. However, we might be missing a test for this on an array. @bcorrie I'm not sure how you want to modify the docs, it might be better to put the array info along in the operator table instead of its own section, or maybe keep its own section but add table with each operator and its array behavior.

bcorrie:
> > Is it expected that in the ADC API you would be able to perform a query on an object like this??? I have been assuming not... Hence it wasn't necessary to talk about these arrays from a query perspective...
> 
> Yes, like `sample.cell_subset` and all the other sample fields.

What I meant is performing a query on the field `sample` itself given that it is an array of objects.

That is, can you do something like:
```
                "op":"OP",
                "content": {
                    "field":"sample",
                    "value":SOMEVALUE
                }
```
Given that `sample` is an object, it is unclear what `SOMEVALUE` would be. This does not make much sense to me.

This is related to the NULL question that was raised in https://github.com/airr-community/airr-standards/issues/280#issuecomment-559452960 as that seems to be the only value that makes sense in this case. I could see *maybe* that you could do something like:

```
                "op":"OP",
                "content": {
                    "field":"sample",
                    "value":null
                }
```
That is essentially what you are talking about for subject.diagnosis in that you have set it to null in some cases, is that correct?

schristley:
> What I meant is performing a query on the field sample itself given that it is an array of objects.

I see. No, I don't think many operators/queries would make sense directly on `sample`. The exception is maybe to check if it does or does not exist.

bcorrie:
> > What I meant is performing a query on the field sample itself given that it is an array of objects.
> 
> I see. No, I don't think many operators/queries would make sense directly on `sample`. The exception is maybe to check if it does or does not exist.

I think this holds true of all Schema objects, correct, including `pcr_target` as @bussec was discussing? The bottom line here is that it is an error to try and apply an ADC API operator to a Schema object (AKA operators can only be applied to base types (not objects))

The one possible exception might be the "is missing" operator as per the discussion in #298 Does it make sense for us to allow for:
```
"op":"is missing", "content": {"field":"sample"}
```
or
```
"op":"is missing", "content": {"field":"sample.pcr_target"}
```
Personally, I think this is quite messy and potentially confusing... If a field is required but nullable, shouldn't it be present in the data as a null field? If you had a data set with 
```
{ ... "sample":[{ ... "pcr_target":null ... }] }
```
This would mean that the data was missing a required, nullable field (e.g. `sample.pcr_target.pcr_target_locus`) ... Doesn't that make it not AIRR compliant? I think it is an important part of the spec that these fields are required and should be provided (even as null) and we should avoid providing shortcuts to make it easy for researchers to leave things out...

bcorrie:
Tests for array operations have been added.

https://github.com/airr-community/adc-api-tests/commit/da9e7503fffe1d4fb5ab9419e28ae82ba8cc4d97

schristley:
> Tests for array operations have been added.

thanks, though I don't think that query works as you expect. When I run it on the test dataset in the reference implementation, it returns 0 results, which is what I would expect.

The `study.keywords_study` is on the whole study and not on individual repertoires. The study in the test dataset has both ig and tcr data so it's value is this:

```
"keywords_study":["contains_ig","contains_tcr"]
```

which would cause all those repertoires to be excluded.

bcorrie:
> > Tests for array operations have been added.
> 
> thanks, though I don't think that query works as you expect. When I run it on the test dataset in the reference implementation, it returns 0 results, which is what I would expect.

Do you have a test server with the gold dataset that you are running this query against? I wrote the test with what I think is an expected result in my head. But I have nothing to test against as we don't have this implemented. If we exclude "contains_tcr" on a study with `study.keywords_study` ["contains_tcr","contains_ig"] (which is what Florian's data has) I would expect it to return 0 repertoires.

If that is what is returned then I think the query/filter is working. no? You are asking for all of the repertoires from studies that do not have TCR data in them... Since Florian's study has TCR data, then none of the repertoires are reported... so we get 0.
> though I don't think that query works as you expect.

Is the problem you are alluding to more about the fact that there are Repertoires in the Florian study that are not TCR - and therefore you would expect to have the IG repertoires reported from this query??? I can see how this might be confusing to the end user, but that expectation doesn't match the query that was issued. The query that would provide this information would be to ask for all of the Repertoires that exclude TCR data (rather than ask for all of the repertoires from the studies that exclude TCR data).

I don't think there is a direct mechanism to actually discover which repertoires are IG or TCR through the metadata that we have  - although we have had several discussions around that (#174 and #185) 8-) In fact that is where the `keywords_study` field came from in the first place... The cell_subset ontology is the closest thing we have at the moment...

Bottom line - if the query returns 0 on the Florian data, I think the query is correct...

schristley:
> Do you have a test server with the gold dataset that you are running this query against?

Yes, but it is not publicly accessible. I just bring it up briefly to run tests then shut it down. I'm working on getting additional VMs so that I can have a more stable test environment, but that's going to take awhile. Even so, I still don't think I'm willing to open it to the public, the reference implementation has no security...

If you have a spare VM, getting up a reference implementation should be fairly straight forward (as in no harder than an iReceptor turnkey). I tried really hard to write complete and accurate [instructions](https://github.com/airr-community/adc-api#installation-procedure).

> If that is what is returned then I think the query/filter is working.

correct, I consider the query/filter to be working correctly.

> Is the problem you are alluding to ...

No, I'm just responding to your comment with the [test commit](https://github.com/airr-community/adc-api-tests/commit/da9e7503fffe1d4fb5ab9419e28ae82ba8cc4d97). I read it wrong and thought you were expecting it to return data. I agree with what you said, sounds like we are both on the same page.

I will say that having a test which returns 0 records isn't the best. There is always the chance that a bug causes all records to be excluded, which would be interpreted as a pass. Better for the test to split and return some data, though I'm not sure if there is another field we can test with, or we'd need to modify the dataset to create the condition.

bcorrie:
> If you have a spare VM, getting up a reference implementation should be fairly straight forward (as in no harder than an iReceptor turnkey). I tried really hard to write complete and accurate [instructions](https://github.com/airr-community/adc-api#installation-procedure).

8-)

We are working on putting the Gold data set on an iReceptor Turnkey so we can use it to test our service API implementation... We should have that up soon... So we probably won't do that for the reference implementation... If we think that would be valuable in general, we could consider it...

bcorrie:
> I will say that having a test which returns 0 records isn't the best. There is always the chance that a bug causes all records to be excluded, which would be interpreted as a pass. Better for the test to split and return some data, though I'm not sure if there is another field we can test with, or we'd need to modify the dataset to create the condition.

Agreed... I will try and figure out a query, but the problem is the Gold data set doesn't have a rich enough set of study_keyword. I suppose I can try and exclude of `['contains_ig', 'contains_paired_chain']`. This should return the set of repertoire that `'contains_tcr'` but not `'contains_ig'` with the `'contains_paired_chain'` being effectively a no-op as there aren't any of those to exclude...

schristley:
Some repertoires have multiple entries for the `data_processing` array... And there are other enhancements we should add to the test data set as well, I went a created an [issue](https://github.com/airr-community/adc-api/issues/4) to be put it all together. Go ahead and add any other ideas you have.

bcorrie:
@schristley can we also merge this, given that other discussions (tests for arrays - https://github.com/airr-community/adc-api/issues/4#issue-566518176), nullable schema objects - #328) are captured in other issues now.

schristley:
@bcorrie yes except we didn't resolve how to document [exclude](https://github.com/airr-community/airr-standards/pull/322#issuecomment-581006136). I don't think `OR` is proper for `exclude`, it actually operates like an `AND`, that is, ALL entries in the array must exclude those values for the document to be returned.

Works a bit like deMorgan's law:

`in`: value OR value OR value
`exclude`: !value AND !value AND !value

It's also similar for the `=` and `!=` operators.

bcorrie:
@schristley OK to delete this branch?

Issue #297, opened 2019-12-16
Formalize relations between x-airr fields


bussec:
Given the recent discussion in #213, #272, #282 and other issues/PRs, would it be helpful if we would formalize the relations between the following `x-airr` fields?
* `miairr`
* `required`
* `nullable`
* `adc-api-optional`

Where "formalizing" means
1. having a comprehensive set of `if-then` rules regarding the state of the fields
2. mapping frequently encountered requirement groups to combination of states
3. defining the default values and whether fields withs default values `MUST`, `MAY` or `MUST NOT` be included in the schema

There is already some information [in the docs](https://github.com/airr-community/airr-standards/blob/master/docs/datarep/overview.rst#airr-extension-properties) and there is also a [table from the ADC API perspective](https://github.com/airr-community/airr-standards/issues/213#issuecomment-546500048), but IMO they don't address all those point. If I missing something, please let me know.

javh:
Also:

1. What's the distinction between `x-airr: required` and the top-level `required` attributed of each object? Is there one?
2. Do these all default to `false` if unspecified?

bcorrie:
It feels to me like these should be resolved for v1.3??? Added the tag...

bcorrie:
In particular, top level required versus x-airr: required needs to sorted, as there are inconsistencies. 

locus is x-airr: required but is not in the YAML required fields

Python validation checks YAML required fields but not x-airr: required.

bcorrie:
Rearrangement is confusing... There is a long list of top level required fields, but many have no x-miairr attribute. Is there any reason why we don't have:
```
    required:
        - sequence_id
```
Imply it is an error NOT to have:
```
        sequence_id:
            type: string
            x-airr:
                required: true
```

bcorrie:
Or maybe get rid of one of them so we don't have both???

schristley:
> What's the distinction between x-airr: required and the top-level required attributed of each object? Is there one?

The primary distinction is that the top-level `required` is an integral part of the OpenAPI spec, so software/middleware will interpret and act upon that setting (for example, by disallowing data to be transferred/saved/etc because it is missing a `required` field). While `x-airr` is an extension object and needs to be manually interpreted.

The top-level `required` only exists for rearrangements (and maybe alignment too) to indicate requirement for the AIRR TSV format (and correspondingly the minimum set of annotations fields required for downstream analysis). The `x-airr: required` indicates that the field is either required for MiAIRR and/or required for ADC API.

[...]

PR #323, opened 2020-01-28
clone schema Olmsted validation

eharkins:
Working on validating clones using the AIRR clones and trees schema in Olmsted!

It seems like the way null/None values are allowed in the AIRR schema at the moment is via [the `x-airr.nullable` field](https://github.com/eharkins/airr-standards/blob/b109492b17971ed8001a911131bf5e77772602f5/specs/airr-schema.yaml#L114). Since in Olmsted, I am just loading in the Clones portion of the yaml schema into my python script and validating directly from there with the [jsonschema](https://python-jsonschema.readthedocs.io/en/stable/) package, I can't make use of that field to allow null values for certain attributes.

The two ways I can think of to allow null values instead of using `x-airr.nullable` are:
1. The way I have done in this pull request
2. [Extending the validator](https://python-jsonschema.readthedocs.io/en/stable/validate/#validating-with-additional-types) to take null in place of any field of type string in the schema. 

Some examples of things I would like to leave as null:
- `Clone.repertoire_id`, since we aren't yet adopting the AIRR Repertoire object / concept in Olmsted; because this is a required field in the AIRR schema here, I understand if you don't want to allow null values for it.
- `Clone.seed_id`: the [description](https://github.com/airr-community/airr-standards/compare/master...eharkins:olmsted_clones_validation?expand=1#diff-7d67e9142fe90cadc6c257313a117f65R2660) says 
> null if no seed

so it seems at least worth adding the `x-airr.nullable` for you own validation purposes.

Thanks!

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -2502,7 +2502,6 @@ Clone:
     type: object
     required:
         - clone_id
-        - repertoire_id
         - germline_alignment
     properties:
         clone_id:
@@ -2517,12 +2516,16 @@ Clone:
             x-airr:
                 miairr: false
                 required: true
+                nullable: true
+                adc-api-optional: false
         data_processing_id:
             type: string
             description: Identifier of the data processing object in the repertoire metadata for this clone.
             x-airr:
                 miairr: false
-                required: false
+                required: true
+                nullable: true
+                adc-api-optional: false
         sequences:
             type: array
             items:
@@ -2653,7 +2656,7 @@ Clone:
                 miairr: false
         seed_id:
             type: string
-            Description: rearrangment id for seed, null if no seed
+            Description: sequence_id of the seed sequence. Empty string (or null) if there is no seed sequence.
             x-airr:
                 miairr: false
 
@@ -2711,7 +2714,7 @@ Node:
             type: string
             description: >
                 Identifier for the Rearrangement object. May be identical to sequence_id,
-                but will usually be a univerally unique record locator for database applications.
+                but will usually be a universally unique record locator for database applications.
             x-airr:
                 miairr: false
                 required: false
--- a/lang/python/airr/specs/airr-schema.yaml
+++ b/lang/python/airr/specs/airr-schema.yaml
@@ -2502,7 +2502,6 @@ Clone:
     type: object
     required:
         - clone_id
-        - repertoire_id
         - germline_alignment
     properties:
         clone_id:
@@ -2517,12 +2516,16 @@ Clone:
             x-airr:
                 miairr: false
                 required: true
+                nullable: true
+                adc-api-optional: false
         data_processing_id:
             type: string
             description: Identifier of the data processing object in the repertoire metadata for this clone.
             x-airr:
                 miairr: false
-                required: false
+                required: true
+                nullable: true
+                adc-api-optional: false
         sequences:
             type: array
             items:
@@ -2653,7 +2656,7 @@ Clone:
                 miairr: false
         seed_id:
             type: string
-            Description: rearrangment id for seed, null if no seed
+            Description: sequence_id of the seed sequence. Empty string (or null) if there is no seed sequence.
             x-airr:
                 miairr: false
 
@@ -2711,7 +2714,7 @@ Node:
             type: string
             description: >
                 Identifier for the Rearrangement object. May be identical to sequence_id,
-                but will usually be a univerally unique record locator for database applications.
+                but will usually be a universally unique record locator for database applications.
             x-airr:
                 miairr: false
                 required: false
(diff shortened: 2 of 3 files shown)


eharkins:
p.s. I think [the fact that I've put quotes around `null` means it will be compatible in Swagger](https://github.com/pipermerriam/flex/issues/91#issuecomment-164507549)

javh:
I think we have some issues to resolve with `x-airr.nullable` (see #297), particularly at levels above Rearrangement (sample meta data and such).

Every field in the Rearrangement schema is nullable, despite not having an `x-airr.nullable = True` attribute. I think we should apply the same assumption here, as Clone/Tree are "downstream" of Rearrangement, while we sort out the issues with the `x-airr` attributes.

schristley:
OpenAPI V3 has a specific `nullable` attribute, but V2 does not. That's why we have `x-airr: nullable`. Once we move to V3, we can use the `nullable` attribute and get rid of `x-airr: nullable`

eharkins:
@javh 
>  I think we should apply the same assumption here, as Clone/Tree are "downstream" of Rearrangement, while we sort out the issues with the x-airr attributes.

Sounds good. How do you allow for this when validating if the `null` type isn't explicitly included in those fields in the schema?

javh:
> Sounds good. How do you allow for this when validating if the `null` type isn't explicitly included in those fields in the schema?

In the reference python/R API packages we treat an empty string as None/NA.

eharkins:
Ah ok. I am using the [generic jsonschema python package](https://python-jsonschema.readthedocs.io/en/stable/validate/) since specific tools for validating AIRR Clones schema in [python don't seem to exist yet](https://github.com/airr-community/airr-standards/tree/master/lang/python).

 [It seems like you are not directly using jsonschema for validation](https://github.com/airr-community/airr-standards/pull/251#issuecomment-550401412) in those API packages, so for now I will probably stick with [allowing strings to be null](https://github.com/matsengrp/olmsted/commit/80b6ec75c7bd919868f9c945f0b447ade638c913) when validating the AIRR spec using jsonschema.

javh:
I'm not familiar with jsonschema, so I'm not sure how to get around the null issue with it. 

But, I don't think we want to add multiple type declarations to fields. That sounds like it would break a lot of things. And ultimately, the nullability mechanism will be in a separate property of some kind (`x-airr: nullable` or `nullable`, as @schristley describes).

If you can't just save empty strings to these fields, can you just exclude them from your output?  `repertoire_id` is currently set in `required`, but removing it from `required` seems a simpler change.  The json/yaml/tsv should still be valid if an optional (non-required) field isn't present (that's what makes them optional).

scharch:
> specific tools for validating AIRR Clones schema in [python don't seem to exist yet](https://github.com/airr-community/airr-standards/tree/master/lang/python).

Actually, that's a good point. We will clearly want this at some point fairly soon, but it seems like it might be too late for 1.3...




> I can't recall exactly what we decided about `repertoire_id` in the clones/tree discussion before, but given recent conversations, it's possible that `repertoire_id` is in the top level `required` just because it had `x-airr: required = true`.

@javh hmmm maybe, but I think we probably want to hold off on that change until #297 is resolved, yes? And in that case, it probably *shouldn't* be `nullable`?

javh:
> @javh hmmm maybe, but I think we probably want to hold off on that change until #297 is resolved, yes? And in that case, it probably _shouldn't_ be `nullable`?

Maybe? The safest bet is probably to just copy `Clone.repertoire_id` unmodified from Rearrangement, which is absent from the top level `required` with:

```
x-airr:
    miairr: false
    required: true
    nullable: true
    adc-api-optional: false
```

Then change it when/if the `x-airr` tags are changed in the rest of the schema after #297 is sorted.

eharkins:
Not having `Clone.repertoire_id` be required sounds good to me.

schristley:
> The safest bet is probably to just copy Clone.repertoire_id unmodified from Rearrangement...

Yes, and those `x-airr` settings are correct, it should be non-miairr, required, nullable, and required for API. These settings should be on all identifiers in the schema, especially ones that link objects in the AIRR Data Model, while identifiers for non-AIRR objects can be optional.

schristley:
> the generic jsonschema python package

We should try to move to something like this. My validation code is a simple recursive hack that's not very robust, it's also slow. However, it sounds like we'd have to modify the schema slightly before passing to the validator, at least for null. Might be worthwhile to see how OpenAPI V3 does against jsonschema, maybe transition to both at the same time.

javh:
Okay, that sounds like consensus. I pushed the discussed changes. I also sync'd the `data_processing_id` x-airr tags and made minor changes to the `seed_id` description (please verify).

eharkins:
FYI I've incorporated the AIRR Clones and Trees schema into Olmsted's input validation scripts. See https://github.com/matsengrp/olmsted/pull/146

schristley:
@javh @eharkins This should be good to merge, right?

javh:
Yeah, I think we can handle the x-airr tags, rearrangement_id and the count fields separately.

PR #324, opened 2020-01-29
Change read_length to be two integer fields for paired reads

### LINKED ISSUE #279: Should read_length be an array of integers?

bcorrie:
Fixes #279

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1108,7 +1108,7 @@ RawSequenceData:
                 miairr: true
                 required: true
                 nullable: true
-                set: 3
+                set: 4
                 subset: process (sequencing)
                 name: Forward read length
         paired_filename:
@@ -1146,7 +1146,7 @@ RawSequenceData:
                 miairr: true
                 required: true
                 nullable: true
-                set: 3
+                set: 4
                 subset: process (sequencing)
                 name: Paired read length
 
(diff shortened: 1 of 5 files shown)


bcorrie:
Can someone have a look at the wording I used in the description....

@bussec this is a MiAIRR term, do you have any concerns about this change... This change implies a MiAIRR set change for the field.

bcorrie:
@bussec left the fields to be in MiAIRR set 3, although they are now in the Schema block `RawSequenceData` where all of the other fields are MiAIRR set 4. But I don't think we want this change to induce an incompatibility. If we want to change the MiAIRR set then we can do that in v2???

Otherwise I think this is ready to merge...

schristley:
We'll have to see how the MiAIRR TSV comes out in the docs. I think the [code](https://github.com/airr-community/airr-standards/blob/6acb7f2936ffafa141369e984229288518701bfc/docs/conf.py#L301) implicitly relies on the fact that the order of definitions in the schema is in set order (1, 2, 3, ...) so these fields may show up in an odd place.

bussec:
Fine with me. Regarding the `set`: Strictly spoken, the addition of the other `set: 4` fields already broke compatibility as they are new requirements that were not present in v1.0. However `set: 4` is kind of an strange place for MiAIRR anyhow (as it is just the raw data), so in reality nothing depends on this.

The remaining problem with moving `read_length` to Set 4 is the NCBI mapping, as this needs to be updated. I am fine with doing this after we merge this in, but it will be another to-do for v1.3.

bussec:
Created the issue to fix the docs: #330

PR #327, opened 2020-02-05
Add complete_vdj boolean to schema to identify truncated alignments

javh:
Added the `complete_vdj` boolean to the schema, per discussion in #309.  I preserved a lot of the language from `NucleicAcidProcessing::complete_sequences`, but it's not exactly the same:

```
complete_vdj:
    type: boolean
    description: >
        True if query nucleotide sequence spans the entire V(D)J region. 
        Meaning, the sequence includes both the first V gene codon that 
        encodes the mature polypeptide chain (i.e., after the leader sequence) 
        and the last complete codon of the J gene 
        (i.e., before the J-C splice site).
```

One concern I have it how to deal with alignments that are not complete according to his definition, because they haven't included mismatches at the 5'/3' ends, despite having a query sequence length that extends further in either direction. So, I referenced the query sequence (`sequence`) instead of the alignment (`sequence_alignment`). Not sure about this.

We should also verify that this definition is consistent with the `full_length` column of the cellranger-vdj annotation CSV.

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1483,6 +1483,8 @@ Rearrangement:
                 sequence_alignment includes both the first V gene codon that encodes the
                 mature polypeptide chain (i.e., after the leader sequence) and the last
                 complete codon of the J gene (i.e., before the J-C splice site).
+                This does not require an absence of deletions within the internal
+                FWR and CDR regions of the alignment.
             x-airr:
                 miairr: false
                 adc-api-optional: true
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1468,7 +1468,7 @@ Rearrangement:
         complete_vdj:
             type: boolean
             description: >
-                True if query nucleotide sequence spans the entire V(D)J region. Meaning, the sequence
+                True if the sequence alignment spans the entire V(D)J region. Meaning, sequence_alignment
                 includes both the first V gene codon that encodes the mature polypeptide chain
                 (i.e., after the leader sequence) and the last complete codon of the
                 J gene (i.e., before the J-C splice site).
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1468,10 +1468,13 @@ Rearrangement:
         complete_vdj:
             type: boolean
             description: >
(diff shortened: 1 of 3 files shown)


javh:
> Do we want to give it a default value? I know we haven't really been adding default values but I'm thinking its useful for `boolean` as we've run into issues with allowing `null` for them.

I'm not sure. I'm inclined to think that the default should be `null` (ie, "I didn't bother to figure it out").

schristley:
> I'm not sure. I'm inclined to think that the default should be null (ie, "I didn't bother to figure it out").

Yeah, that's a good point. Don't really want to assume one way or other if it hasn't actually been computed. Should add `x-airr: nullable: true` then.

bcorrie:
> We should also verify that this definition is consistent with the `full_length` column of the cellranger-vdj annotation CSV.

I have pinged someone at 10X...

javh:
> I have pinged someone at 10X...

Me too... Great minds. :)

javh:
From the call:
* Heard back from 10X and this definition matches their `full_length` exactly.
* Change to `sequence_alignment` context.
* Ping IgBLAST devs for comments.

javh:
Changed to:

```
        complete_vdj:
            type: boolean
            description: >
                True if query nucleotide sequence spans the entire V(D)J region. Meaning, the 
                sequence includes both the first V gene codon that encodes the mature polypeptide 
                chain (i.e., after the leader sequence) and the last complete codon of the
                J gene (i.e., before the J-C splice site).
```

bcorrie:
> > I have pinged someone at 10X...
> 
> Me too... Great minds. :)

Even better, our two 10X people agreed that it is correct 8-)

javh:
Heard back from @jianye00:

1. Doesn't see a problem with this field for IgBLAST.
2. Shares the sentiment that if the alignment is truncated at the ends due to mismatches, then that's a software problem, not a spec problem.
3. Thinks the field should explicitly require FWR1-FWR4 to be present for `complete_vdj` to be true.

My take on (3) is that that would be the ideal definition, but is probably not practical for facilitating adoption (e.g., cellranger-vdj doesn't call FWR boundaries).

javh:
Heard back from Veronique @ IMGT.  She's okay with this definition.  They can derive it from the `V-REGION partial 5prime missing nt nb` and `J-REGION partial 3prime missing nt nb` of the `3. nt-sequences` file.

I pushed a change to address a concern from @jianye00 about internal deletions.

I'm good to merge this if everyone else is cool with it as is.

PR #331, opened 2020-02-13
operator documentation

### LINKED ISSUE #298: Functionality of is operator

schristley:
fixes #298 

Also changed operator examples to be JSON, and `organism.id` is a string.

### DIFF
--- a/docs/api/adc_api.rst
+++ b/docs/api/adc_api.rst
@@ -284,7 +284,7 @@ The response will provide the ``Repertoire`` data in JSON format.
            "subject_id":"TW02A",
            "synthetic":false,
            "linked_subjects":"TW02B",
-           "organism":{"id":9606,"value":"Homo sapiens"},
+           "organism":{"id":"9606","value":"Homo sapiens"},
            "age":"25yr",
            "link_type":"twin",
            "sex":"F"
@@ -467,6 +467,10 @@ recursive algorithms are typically used for tree traversal.
 
 The following operators are support by the ADC API.
 
+.. |br| raw:: html
+
+    <br>
+
 .. list-table::
     :widths: auto
     :header-rows: 1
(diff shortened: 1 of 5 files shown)


bcorrie:
@schristley in Pull #322 I added a section on operations on arrays. I think something like this is still needed, no, as we don't capture that information in this pull request... I think combined perhaps they do the trick.

It also occurred to me that maybe we should add `array` to the `Value Data Types` column of the operator documentation

It also occurred to me that if we don't add `array` to the `in` and `exclude` operators then the weird semantics of `exclude` on arrays don't need to be explained... Do you think it is the end of the world if the spec does not allow `in` and `exclude` on fields that are arrays???

Thoughts?

schristley:
> I think something like this is still needed...

Yes, I agree, still needed.

> It also occurred to me that maybe we should add array to the Value Data Types column of the operator documentation

Ok, that sounds good.

> It also occurred to me that if we don't add array to the in and exclude operators then the weird semantics of exclude on arrays don't need to be explained... Do you think it is the end of the world if the spec does not allow in and exclude on fields that are arrays???

I don't think we can get away from it. When you are doing a query on `sample.tissue`, even though the `tissue` field is a base data type, `sample` is an array, so technically the query is being performed on an array. That is, if the repertoire has multiple samples in the array, doing a query on `sample.tissue` will operate like how you documented.

bcorrie:
> > It also occurred to me that maybe we should add array to the Value Data Types column of the operator documentation
> 
> Ok, that sounds good.

You or me... 8-) I can add this if you haven't already...

schristley:
> You or me... 8-) I can add this if you haven't already...

I can stick it in this PR real quick as it's modifying that table. I wanted to merge this as #298 has been decided, so we can check another one off the list!

schristley:
@bcorrie added array for `in` and `exclude` in table. Good to merge?

PR #335, opened 2020-02-20
include required docs

### LINKED ISSUE #213: specification of required fields for ADC API
The `fields` parameter to the API allows the caller to specify which fields to return. In the case of `/rearrangements`, with AIRR TSV format, this can produce output that would not pass validation checks, because required fields are missing. Should we enforce that `required` fields are always returned, which kind of defeats the purpose of `fields`.

A more reasonable scenario is that a user wants all of the required fields, so it pass validation, plus a few more they are interested in, but not the full set. In this case, the user has to know the full set of `required` fields to pass `fields`, which they may not easily know without looking it up. Maybe there is an easier way to handle this case.

bcorrie:
@schristley I realized that we haven't updated the docs for the include_required, even though we had closed #213 

This pull requests makes some minor changes to the docs for the include_required flag.


schristley:
thanks @bcorrie , I was waiting on #297 so that the doc could be more explicit about what "required" means, but I'm fine with putting this in and refining it later.

bcorrie:
> thanks @bcorrie , I was waiting on #297 so that the doc could be more explicit about what "required" means, but I'm fine with putting this in and refining it later.

Right... makes sense...

bcorrie:
@schristley I think we have a bit of an issue with the new x-airr field designation... There is currently no mechanism to specify which fields are required by the ADC API as we discussed here:
https://github.com/airr-community/airr-standards/issues/213#issuecomment-548480276

For your reference API "include_required" implementation, what are you considering "required"?

schristley:
> For your reference API "include_required" implementation, what are you considering "required"?

It's not implemented yet, but with the new `x-airr` tags, I'm thinking these:

- required (top-level)
- `x-airr:identifier` == true

the `x-airr:identifier` should cover the additional ADC API fields that are not marked `required`

bcorrie:
Is that a 

> > For your reference API "include_required" implementation, what are you considering "required"?
> 
> It's not implemented yet, but with the new `x-airr` tags, I'm thinking these:
> 
> * required (top-level)
> * `x-airr:identifier` == true
> 
> the `x-airr:identifier` should cover the additional ADC API fields that are not marked `required`

Not sure I follow what you mean by `x-airr:identifier`?

bcorrie:
I suppose I am left wondering a bit, doesn't it make sense for the ADC API to return ALL fields in the spec, except possibly those that are deprecated?

schristley:
[here](https://github.com/airr-community/airr-standards/blob/e253c311770c26aac55eccf8e7987235db7a4297/specs/airr-schema.yaml#L49)

It covers not just the ADC API but broadly the AIRR Data Model to insure that the object can be linked with each other.

javh:
Ah, yeah... Forgot about `x-airr: identifier`. The conversation is here:
https://github.com/airr-community/airr-standards/issues/297#issuecomment-589738337

schristley:
> I suppose I am left wondering a bit, doesn't it make sense for the ADC API to return ALL fields in the spec, except possibly those that are deprecated?

I don't like that at all, especially for Rearrangement

bcorrie:
> Ah, yeah... Forgot about `x-airr: identifier`. The conversation is here:
> [#297 (comment)](https://github.com/airr-community/airr-standards/issues/297#issuecomment-589738337)

Missed that entirely... 8-) So nothing in the spec has been marked in this fashion yet, correct?

My question above (https://github.com/airr-community/airr-standards/pull/335#issuecomment-606251949) I think still applies. Why don't we keep it simple and for the ADC API have the `include_required` field mean all fields in the spec??? It wouldn't be include_required, it would be `include_all`, which would be MiAIRR compliant and include all spec fields... It seems like that is what we really want.

schristley:
> So nothing in the spec has been marked in this fashion yet, correct?

I've added them in #364 

> It seems like that is what we really want.

That's not what I want, but I would consider an additional parameter `include_all` that does that.

bcorrie:
> > It seems like that is what we really want.
> 
> That's not what I want, but I would consider an additional parameter `include_all` that does that.

Hold on, my mistake... No field spec gets you ALL the fields, so we don't need that, you are correct...

bcorrie:
So this does raise the need to clarify exactly what the ADC API query parameter `include_required` gives you... 

Recall here is the table:
| Query parameters                    | ADC API required | ADC API Not Required | Extra fields | Requested fields|
|-----------------|:---:|:---:|:---:|:---:|
| None                                        | Y|Y|Y|N|
| fields=[,,,]                                  |N|N|N|Y|
|include_required:true,fields=[,,,]|Y|N|N|Y|
|include_required:true                 |Y|N|N|N|

@schristley you are suggesting that `include_required` will get you:

> * required (top-level)
> * `x-airr:identifier` == true

bcorrie:
> @schristley you are suggesting that `include_required` will get you:
> 
> > * required (top-level)
> > * `x-airr:identifier` == true

Is there a reason we don't just use the required at the top level for this? From an ADC API (openAPI) perspective that is what you would expect...

javh:
> Is there a reason we don't just use the required at the top level for this? From an ADC API (openAPI) perspective that is what you would expect...

The identifiers that link object in the full data model aren't necessarily required in a TSV. Eg, `repertoire_id` is not in `required` for Rearrangement.

bcorrie:
> > Is there a reason we don't just use the required at the top level for this? From an ADC API (openAPI) perspective that is what you would expect...
> 
> The identifiers that link objects in the full data model aren't necessarily required in a TSV (or other isolated usage). Eg, `repertoire_id` is not in `required` for Rearrangement.

I really don't like having to infer what is required for the ADC API response by having to deal with two attributes in the spec, in particular two attributes in the spec that are in completely different parts of the spec... 

1. the top level "required" attribute
2. the x-airr identifier:true attribute

This just seems really clunky...

schristley:
We should have a different name for this parameter, how about `include_core`, `include_principal`? Or something more explicit like `include_data_model_fields`?

bcorrie:
> We should have a different name for this parameter, how about `include_core`, `include_principal`? Or something more explicit like `include_data_model_fields`? Or `include_required_and_identifiers`?

Are you talking about changing the name of parameter to the API query to get the these fields, which is currently `include_required`?

I don't think we have quite settled on what this is returning, which would impact the name we use 8-)

I think my own preference for the semantics of what this parameter would mean is that it returns everything that is defined in the spec... @schristley I know for rearrangements you mentioned that this was not desirable, but is that really true...

Right now, using `include_required` on rearrangements will get you:
```
Rearrangement:
    discriminator: AIRR
    type: object
    required:
        - sequence_id
        - sequence
        - rev_comp
        - productive
        - v_call
        - d_call
        - j_call
        - sequence_alignment
        - germline_alignment
        - junction
        - junction_aa
        - v_cigar
        - d_cigar
        - j_cigar
```
plus the identifier fields
```
        cell_id:
        clone_id:
        repertoire_id:
        sample_processing_id:
        data_processing_id:
```
Although this is AIRR compliant (it is a subset of the fields, the rest of which are not required), it isn't really what you would want to output into an AIRR TSV file as it has many missing AIRR spec fields which the repository might have data for. There is no way to get "just" all of the AIRR TSV fields that are in the repository with the above semantics.  If you provide no parameters to the query, you get all fields, which include any extra fields that the repository might include... 

I think we need an ADC API parameter that would result in returning all of the fields that are in the AIRR spec - essentially a fully fleshed out AIRR spec response. If we don't use `include_required` for this, maybe we do need an `include_airr` or something similar. I guess I wonder about the use case for using `include_required` as defined above. From a DataRep perspective, we rapidly came to the conclusion that the MiAIRR fields (the `required` fields) weren't sufficient for most use cases... So I wonder about the value of returning just those fields.

My preference is that we have a parameter that would ensure:

- For the /repertoire endpoint getting a Repertoire JSON object that has all fields that are in the AIRR spec, null if they are nullable and the repository doesn't have a value. This should include the "optional" fields like `primary_annotation`, `data_processing_files`, `repertoire_name`, `repertoire_description`, etc.
- For the /rearrangement endpoint, getting all of the fields in that are in the AIRR spec, again null if they are nullable and don't exist in the repository. This, combined with the format:TSV parameter, will give you an AIRR TSV file with all of the AIRR field columns in the file.

schristley:
Do I sense feature creep? ;-D

So I originally wanted `include_required` for the very specific purpose that when the `fields` parameter was used, the user could also get the "core" fields so that validation would pass. Essentially the use case is, the user wants all the main important fields plus these others specified with the `fields` parameter, no more no less. I don't want everything, otherwise what's the point of limiting the set with the `fields` parameter? This is meant to be a minimal set without the user having to remember the actual names of those "core" fields.

Now what I gather is that you are suggesting there are additional use cases beyond that one. That's fine, just please don't get rid of that use case.

My suggestion is then to change `include_required` to something like `include_fields` which is an enumeration. We could have have multiple values like `required` for strictly required, `airr` for the core fields for validation (required and identifiers), `all` for everything in the schema, etc.

I'm fine with that but I don't know what other use cases you want. You will need to define them.

It needs to be clear how custom fields are handled. Are they always returned/never returned? Do some of the settings exclude/include them?

bcorrie:
> My suggestion is then to change `include_required` to something like `include_fields` which is an enumeration. We could have have multiple values like `required` for strictly required, `airr` for the core fields for validation (required and identifiers), `all` for everything in the schema, etc.

I think this is reasonable... Gives us flexibility... We need to decide the enum types 8-)

One question. Is it true that a field is `required` if and only if `x-airr:miairr=[essential|important|defined]`.  That is, if a field is MiAIRR then it is required and if the field is required it is MiAIRR??? Maybe a question for @bussec in particular?

If so, then perhaps the enum should be `miairr`, `validation`, and `defined`

bussec:
@bcorrie That's an interesting question that I also have on my tick list. So from the MiAIRR perspective:

* Having an `xairr:miairr=(essential|important)` property is sufficient for a field to be `required`
* Having an `xairr:miairr=defined` property is *not* sufficient for being `required` (i.e. unless there is something else that makes it required, it is not).

If I understand your "if and only if" term correctly, this would imply that the `miairr` property is currently to only thing that can make a field required. But what about `identifier` fields? Is it implied in the `identifier` tag that they are required?

schristley:
There are fields in rearrangements that are required but non-MiAIRR, e.g. `productive`

schristley:
We don't want to restrict ourselves to purely MiAIRR and its definitions because we have many fields that are not part of MiAIRR.

Actually though, the enum value of `miairr` might be useful. It gives all of the MiAIRR fields, which is a subset of the `required` fields.

I like the enum value of `required` as it can mean strictly the fields marked `required` in the schema. This is the MiAIRR fields plus the additional ones marked required.

The next level is the required fields plus the identifier fields, this combo should pass the validation check but I find the enum value of `validation` to not be very meaningful from a user perspective. I kinda like `airr` to signify the AIRR Data Model. This is presumably the use case that will be used most often.

Lastly, we can have `full` or `all` which will fill out all of the fields that are in the schema. Note that this is not the same not specifying `fields` in the query, that returns all of the fields in the database, not necessarily all fields in the schema. So this setting would force all those fields to present.

Now how to handle custom fields, if you want those included?

Maybe the idea of levels is not best? Should we make `include_fields` be an array, versus a string, then each enum value could be a set that's combined together.

`include_fields=["miairr","required","identifiers","airr","custom","all"]`

I don't know, how crazy do you want to get, do we need all these options?

bcorrie:
> Actually though, the enum value of `miairr` might be useful. It gives all of the MiAIRR fields, which is a subset of the `required` fields.
> 
> I like the enum value of `required` as it can mean strictly the fields marked `required` in the schema. This is the MiAIRR fields plus the additional ones marked required.

Given @bussec description above, `miairr` is not a subset of `required` as there are some fields in `miairr` that are not `required` (x-airr.miairr:defined != required).

bcorrie:
> Maybe the idea of levels is not best? Should we make `include_fields` be an array, versus a string, then each enum value could be a set that's combined together.
> 
> `include_fields=["miairr","required","identifiers","airr","custom","all"]`
> 
> I don't know, how crazy do you want to get, do we need all these options?

I think the minimal we need would be:
- `miairr`: All of the MiAIRR fields (x-airr.miairr != null/missing) This is the "I want the MiAIRR fields and that is it, damn it" request.
- `required`: All of the Required fields (json.required). This is the "I want all the fields required to pass AIRR validation and that is it, damn it" request.
- `spec`: All of the fields in the AIRR spec. This is the "I don't have any strong opinions, so I want everything that the AIRR Community thinks is important and worth defining" request.

We don't need an all, because not providing `include_fields` means you get EVERYTHING!

One question is whether someone would want the `miairr` fields without the `x-airr.miairr == defined` fields... I am not sure why one would want that since MiAIRR is already minimal...

schristley:
> We don't need an all, because not providing include_fields means you get EVERYTHING!

That's not exactly true. You will get everything that is stored in the repository, which isn't necessarily all schema fields, e.g. the repository might not store fields that are null. Even so, I'm not sure if `all` is really needed but you had suggested earlier wanting to force the output to contain all schema fields, i.e. giving them null values if necessary.

bcorrie:
> > We don't need an all, because not providing include_fields means you get EVERYTHING!
> 
> That's not exactly true. You will get everything that is stored in the repository, which isn't necessarily all schema fields, e.g. the repository might not store fields that are null. Even so, I'm not sure if `all` is really needed but you had suggested earlier wanting to force the output to contain all schema fields, i.e. giving them null values if necessary.

Hmmm, good point. If you want all spec fields (`spec`) + all extra fields not providing include_fields doesn't cover it...

I also am not sure that is needed...

bcorrie:
@schristley should we try to finalize these terms...

How about:
- miairr - meaning a MiAIRR field
- validation - meaning a field that is in JSON.required (or airr-validation)
- defined - meaning a field that is defined in the spec (or airr-defined)

schristley:
@bcorrie I like...

* miairr - same as you
* airr - airr beyond the "mi", expected to be the most common option used
* all - force all schema fields to be there

bcorrie:
> @bcorrie I like...
> 
> * miairr - same as you
> * airr - airr beyond the "mi", expected to be the most common option used
> * all - force all schema fields to be there

My only concern with that is I find it a bit confusing on two fronts...

My naive interpretation of `airr` would be everything in the AIRR schema (your `all`)... To confirm, I think you are saying that `airr` should return all of the JSON required fields plus all of the fields that are tagged with the `identifier:true` attribute, correct? I think it is confusing to call that AIRR when really it is those fields required to pass AIRR validation. 

I find `all` also confusing because we have another level for all when one doesn't provide a `required_fields`, which returns all fields in the repository... I think it is better to be explicit in the enum for the `required_fields` parameter you provide as to exactly what it means.

schristley:
> My naive interpretation of airr would be everything in the AIRR schema (your all)... To confirm, I think you are saying that airr should return all of the JSON required fields plus all of the fields that are tagged with the identifier:true attribute, correct? I think it is confusing to call that AIRR when really it is those fields required to pass AIRR validation.

Maybe, but I also tend to think you are thinking too technically about the words, too keyed into specific attributes and their values. I'm just thinking about simplicity.

Imagine this, a GUI with some checkboxes to download data, and a user (like say Felix) has to decide what to download, in one UI the choices are 1) MiAIRR, 2) Validation and 3) Defined. Felix is like what? So there's the "help" text with Validation that says 'all JSON required fields as well as required=true so that AIRR validation will pass". Felix is still like what?

Now the other UI has the choices 1) MiAIRR, 2) AIRR and 3) All. With the "help" text for AIRR saying something simple like "Core AIRR fields sufficient for most purposes".

Of course I'm exaggerating but I feel pretty confident that any values will be confusing to somebody... ;-D

bcorrie:
> Imagine this, a GUI with some checkboxes to download data, and a user (like say Felix) has to decide what to download, in one UI the choices are 1) MiAIRR, 2) Validation and 3) Defined. Felix is like what? So there's the "help" text with Validation that says 'all JSON required fields as well as required=true so that AIRR validation will pass". Felix is still like what?
> 
> Now the other UI has the choices 1) MiAIRR, 2) AIRR and 3) All. With the "help" text for AIRR saying something simple like "Core AIRR fields sufficient for most purposes".

The only thing I don't like is having the term "AIRR" refer to something that doesn't encompass the entire AIRR spec. I think it would more confusing to poor Felix (the hypothetical biologist in this conversation) to ask for AIRR and then find out that something in the AIRR spec airr-schema.yaml wasn't returned??? 

How about `miairr`, `airr-core` and `airr-spec` using your "Core AIRR fields sufficient for most purposes" phrasing for that set of fields? Or maybe `airr-analysis` rather than `airr-core` to be more explicit about the intended purpose???

schristley:
> How about miairr, airr-core and airr-spec using your "Core AIRR fields sufficient for most purposes" phrasing for that set of fields?

ok, can we use `airr-schema` or `airr-specification` for the third option? I prefer not to use abbreviated words.

bcorrie:
> > How about miairr, airr-core and airr-spec using your "Core AIRR fields sufficient for most purposes" phrasing for that set of fields?
> 
> ok, can we use `airr-schema` or `airr-specification` for the third option? I prefer not to use abbreviated words.

Works for me... Maybe `airr-schema`?

schristley:
@bcorrie I made a pass at updating the docs...

PR #336, opened 2020-02-22
Add OntoVoc to Docs

bussec:
Currently the OntoVoc reports are only available as GDocs, but they should be part of the official AIRR documentation.

Add the report of the November 2018 sprint to the docs.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -276,12 +276,23 @@ for spec in tables:
     for k, v in properties.items():
         row = {}
         row['name'] = k
-        if v.get('type'):
-            row['type'] = v['type']
+        if v.get('type') == 'array':
+            if v['items'].get('$ref') is not None:
+                sn = v['items'].get('$ref').split('/')[-1]
+                row['type'] = '``array`` of :ref:`' + sn + ' <' + sn + 'Fields>`'
+            elif v['items'].get('type') is not None:
+                row['type'] = '``array`` of ``' + v['items']['type'] + '``'
+            else:
+                row['type'] = '``' + v['type'] + '``'
+        elif v.get('type'):
+            row['type'] = '``' + v['type'] + '``'
         elif v.get('$ref') == '#/Ontology':
-            row['type'] = 'Ontology object'
+            row['type'] = ':ref:`Ontology <OntoVoc>`'
+        elif v.get('$ref') is not None:
+            sn = v.get('$ref').split('/')[-1]
+            row['type'] = ':ref:`' + sn + ' <' + sn + 'Fields>`'
         else:
-            row['type'] = 'unknown'
+            row['type'] = '``unknown``'
         if v.get('x-airr') and v.get('x-airr').get('miairr'):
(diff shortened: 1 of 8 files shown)


bcorrie:
Looks good to me... The only thing I wonder is if we should have a section that summarizes the current state of the ontologies agreed at the current point in time, and then have the sections that document the outcomes from the various sprints. That way someone doesn't need to dig through the notes to find the ontologies that are currently agreed???

Basically add an Ontology summary section to docs/ontovoc/introduction_ontovoc.rst

bussec:
I also thought about this, but it would be redundant to the YAML. So it is ease-of-use vs. not having to keep stuff in sync (unless we autogenerate the overview table).

schristley:
> should have a section that summarizes the current state of the ontologies agreed at the current point in time...

I agree.

> I also thought about this, but it would be redundant to the YAML. So it is ease-of-use vs. not having to keep stuff in sync (unless we autogenerate the overview table).

I looked into trying to autogenerate into the [tables](https://docs.airr-community.org/en/latest/datarep/metadata.html#study-fields), and it's probably more work to code than it's worth, especially to put in a nice format with any bells and whistles. I think it might be easier to have a summary section, hopefully the manual updating won't be onerous as the ontologies don't change much. Those tables can then point to that section for detail. That would also allow us to provide more helper text that we cannot stuff in the YAML, e.g. discussion of how to code PBMC.

schristley:
I've added a summary table plus links back/forth between the schema fields and the ontology info.

@bussec Couple questions, I think the `study_type` ontology needs to be added? Also some fields are in the schema are marked as `draft: true`, are these correct and do we need to indicate that somehow in the table?

bussec:
@schristley 
> I think the study_type ontology needs to be added?

Yes. This is a "legacy ontology" we decided on before OntoVoc started. We will formally confirm it in the next sprint. Add it to the docs, labeled as `draft` (5ef27c6).

> some fields are in the schema are marked as draft: true, are these correct and do we need to indicate that somehow in the table?

Draft status of an ontology should also be represented in the docs (see above). However, the ontologies of the 11/18 sprint have been confirmed in Genoa, so they are not draft anymore.
Fixed draft status (ce625da).

bussec:
I now also included the units ontology (UO) for `age_unit` (ac9a5aa)

bussec:
`NCIT` and `UO` are now both on the review list for the 04/2020 sprint.

bussec:
@schristley @bcorrie Are we good to go?

bcorrie:
I added root nodes to cell_subset and organism... Assuming we wanted that... I did not add paths for them, as they were either very long or very complex (or both) 8-)

bussec:
@bcorrie 
> I added root nodes to cell_subset and organism... Assuming we wanted that...

Fine with me. As they were not part of the last report, I put them up to be confirmed during the next sprint.

bcorrie:
OK, I added them because they were in the spec... 8-) We can confirm and update if need be in the next spring...

PR #337, opened 2020-02-24
Changed disease_diagnosis to use ontology

bcorrie:
As per the Ontology definitions that were agreed in the last Ontology spring in 2019. We overlooked disease_diagnosis.

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -464,9 +464,11 @@ Diagnosis:
                 subset: diagnosis and intervention
                 name: Study group description
         disease_diagnosis:
-            type: string
+            $ref: '#/Ontology'
             description: Diagnosis of subject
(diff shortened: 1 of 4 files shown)


bcorrie:
@bussec can you review this as per the previous OntoVoc sprint...

I think we want to just add this, as it was overlooked when we added the other ontologies.

bcorrie:
@schristley are the docs here: https://docs.airr-community.org/en/latest/datarep/metadata.html#diagnosis-fields generated automatically from the spec? Or do they need to be manually updated?

schristley:
> generated automatically from the spec?

Yeah, [conf.py](https://github.com/airr-community/airr-standards/blob/master/docs/conf.py) gathers the data together for building the table.

I've wanted to provide more detail in the table about the ontology though, as we could pull that info out of the schema. On the other hand, the ontology should change infrequently so maybe it's easier to have manual doc like #336 and just point to it.

PR #341, opened 2020-02-28
locus should be a controlled vocabulary

bcorrie:
Matches type for pcr_target_locus etc.


schristley:
add `keywords_study` while we are at it.

PR #319, opened 2020-01-22
Change and document MiAIRR requirement levels

bussec:
* Change various requirement level to provide for easier annotation by third-party annotators.
* Add documentation for requirement levels.
* Add guideline for annotation of synthetic libraries
* Expand controlled vocabulary for `physical_linkage`

Fixes #310 
Fixes #204

PR #342, opened 2020-02-28
Rework of x-airr requirement tags

javh:
In the interesting of moving #297 forward, this is a potential rework for some of the `x-airr` tags based on discussion in #297 and #319.

Changes this:

- `required` (top level)
- `x-airr: miairr` (boolean)
- `x-airr: required`
- `x-airr: adc-api-optional`
- `x-airr: nullable`

To this:

- `required` (top level)
- `x-airr: miairr` (enum)
- `x-airr: adc-query-support`
- `x-airr: nullable`

I only changed the entries in the `Attributes` object of the schema. If we reach consensus on that, we can propagate the changes to the fields.

This is under the assumption that the top level `required` will be used to denote whether a field must be present.

### DIFF
--- a/specs/airr-schema.yaml
+++ b/specs/airr-schema.yaml
@@ -444,7 +444,7 @@ Subject:
         race:
             type: string
             description: Racial group of subject (as defined by NIH)
-            example: White, American Indian or Alaska Native, Black, Asian, Native Hawaiian or Other Pacific Islander, Other 
+            example: White, American Indian or Alaska Native, Black, Asian, Native Hawaiian or Other Pacific Islander, Other
             x-airr:
                 miairr: important
                 nullable: true
@@ -999,7 +999,7 @@ NucleicAcidProcessing:
                 If a PCR step was performed that specifically targets the IG/TR loci, the target and primer locations
                 need to be provided here. This field holds an array of PCRTarget objects, so that multiplex PCR setups
                 amplifying multiple loci at the same time can be annotated using one record per locus. PCR setups not
-                targeting any specific locus must not annotate this field but select the appropriate 
+                targeting any specific locus must not annotate this field but select the appropriate
                 library_generation_method instead.
             items:
                 $ref: '#/PCRTarget'


bussec:
@javh We discussed the MiAIRR requirement levels in the last MiniStd call and majority opinion was that the current terms (e.g. `required`, `recommended` and `optional`) are confusing as a) even `recommended` is still minimal (i.e. "required" in the common sense) and b) not all people will make the direct connection to RFC2119 and its MUST, SHOULD, MAY terms. The topic is still under discussion, but one proposal (which comes from the RDA FAIR working group) would be:

* Essential (for `required`)
* Important (for `recommended`)
* Useful (for `optional`)

We can of course still replace these terms later on, but any feedback would be welcome.

schristley:
> not all people will make the direct connection to RFC2119 and its MUST, SHOULD, MAY terms.

should we use those terms then, `must`, `should` and `may`?

javh:
I definitely prefer "essential" over "required" as it removes the perceived conflict with the top level `required` tag. I went ahead and pushed a change with these new terms.

bcorrie:
> * Useful (for `optional`)

Synonyms for useful that provide a bit more emphasis on the value of the field???

`valuable, beneficial, helpful`

Although this may be more confusing - what is higher in the hierarchy, `important` or `valuable`? 8-)

One other option might be to use a word that is "value" agnostic, but simply states that the field is `defined`, where `defined` implies that the field, being part of MiAIRR, is useful, valuable, beneficial, and helpful to provide in a `well defined` way??? So a field would either be:

"miairr":"essential"
"miairr":"important"
"miairr":"defined"

schristley:
What I took from the discussion is 1) there currently are no MiAIRR fields with the `useful` requirement level, 2) that when it does get used in the future, it will likely be for extensions fields in specific contexts (I think an example was `strain_name` which is relevant for mice but not for humans), and 3) non-MiAIRR fields will not have the `miairr` attribute at all.

Based on 2), I like the term `extension` for that level. Though it's likely that MiAIRR will want to be able to specify, in the schema somehow, when and where the extension is appropriate.

My suggestion is, if the `essential` and `important` levels are pretty much good, we should move forward with them and leave the 3rd level unspecified until we have more concrete use cases.

schristley:
I think the only thing we maybe haven't discussed sufficiently is what fields are included in `required` (top level) and what's the basis for making them required?

javh:
> I think the only thing we maybe haven't discussed sufficiently is what fields are included in `required` (top level) and what's the basis for making them required?

I think we'll just have to go through everything and make a judgement call. 

With @bussec's most recent comment in #297, I think we are ready to do that. Yes?

bussec:
Performed manual rebase to resolve merge conflicts.

PR #345, opened 2020-03-04
Fix handling of keywords by Sphinx compiler

bussec:
Currently `docs/conf.py` cannot handle the nested `enum` of `keywords_study`, which has an `items` key on the default nesting level, while the `enum` is only a key within it.

This patch fixes the handling.


bussec:
Ok, looks like @schristley was a bit faster than me ;-)

PR #350, opened 2020-03-12
Added a check for existence of the object in the array before accessing.

bcorrie:
This fails if the field is not in the file.


bcorrie:
Minor change, but since I am not the owner of this code thought I would ask for review 8-)

PR #351, opened 2020-03-12
Rearrangement set

bcorrie:
Pull request to check diffs - think this is a dead branch...


bcorrie:
Indeed, this is an old stale branch, closing and deleting...

javh:
FYI - there is a "Compare" button right next to the "Pull Request" button if you select a branch (banner on top of the code window), so you can skip the temporary PR.

PR #356, opened 2020-03-16
Added x-airr keywords for primary_annotation

bcorrie:
Assume this should be the same as data_processing_id


schristley:
This is included in #364

Issue #266, opened 2019-10-14
Rename `organism` to `species`?


bussec:
Not keen to bring this up, but working on #260 I realized that the `organism` field should actually be termed `species`, see [species](https://en.wikipedia.org/wiki/Species) and [organism](https://en.wikipedia.org/wiki/Organism). The decision for `organism` was originally based on the use of the `/organism` qualifier by INSDC. The [INSDC FT docs](http://www.insdc.org/documents/feature-table) use an "organism is-a species" construct to justify the name of the qualifier. However, this is difficult to apply when the term should be used as a suffix (`cell_species` is rather obvious, `cell_organisms` IMO not).

Therefore I wanted to get an idea what peoples' opinion on the following options are:
1. rename `organism` to `species`
2. propagate the mislabeling to other fields encoding species
3. have inconsistent terms for the same information within the AIRR Schema

scharch:
I lean toward 3, but it's a very soft lean...

bcorrie:
I would lean towards fixing it if it is incorrect, but also a very soft lean... We are going to make a major release, if it needs to be fixed better to do it sooner rather than later IMHO.

If we don't fix it then I think what we do want to get right is the description of the AIRR term, as that is what the users will see. For organism we have:

```
            description: Binomial designation of subject's species
                name: Organism
```
So if the actual keyword doesn't change, I would suggest that we change "name: Organism" to "name: Species".

bussec:
Ok, we'll discuss about the MiAIRR field name tomorrow.

bussec:
We discussed this in the MiniStd call today and the general consensus was that we should change the name from "Organism" to "Species" for MiAIRR v2. Note that this is MiAIRR name of the field (i.e. the thing which is encoded in the `name` key - as in Brian's post), not the `organism` field itself. I have a soft lean toward 1 (i.e. also renaming the field), but I am biased as the required changes wouldn't create a substantial workload on my side.

bussec:
Issue was discussed again in the MiniStd call on 2019-12-13 and the decision was made to rename the field in v2.0 of the schema.

bcorrie:
Also discussed at DataRep 2020-03-09 and we agreed that this change made sense... Jump in anyone if they want to clarify...

javh:
Field has been renamed in #359 for v1.3. We'll just need to change the `x-airr: name` tag in v2.0 to:

```
species:
    description: Binomial designation of subject's species
    x-airr:
        name: Species
```

[...]

PR #359, opened 2020-03-19
Change organism to species

javh:
Per #266, changed `organism` to `species` in Subject.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -153,12 +153,18 @@ following checklist:
 -  Upload python library to PyPI
 -  Upload R library to CRAN
 
-Deprecation
------------
+Field Deprecation
+-----------------
 
-Deprecated schema fields should remain in the schema files and be labeled
-with the appropriate ``x-airr`` tags that denote deprecation, explain the
-rationale, and specify the replacement fields (if any). For example:
+Deprecated schema fields should remain in the schema until at least
+the next ``MAJOR`` version number. Preferably, deprecated schema fields
+should be retained indefinitely, unless there is a clear need to remove
+them.
+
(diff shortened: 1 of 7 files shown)


schristley:
@javh as I understand this is V2, so we don't merge until after V1.3 is out? or maybe create a V2 branch?

javh:
I labeled with the V2.0 milestone, but I'm not sure that we need to wait... Do you remember when we introduced this field? Can we just make the description language consistent with MiAIRR "Organism" right now and update it to the new MiAIRR language with V2.0?

schristley:
I thought @bussec had said that essentially `organism` was part of the MiAIRR V1, so while it technically wasn't in the AIRR schema yet, it was in the [MiAIRR table](https://github.com/airr-community/airr-standards/blob/v1.0.0/AIRR_Minimal_Standard_Data_Elements.tsv)

javh:
So can we just have this in v1.3:

```
species:
    x-airr:
        name: Organism
```

And then change it in v2.0 to:
```
species:
    x-airr:
        name: Species
```
 Yay? Nay?

bussec:
@javh Yay, this is also my understanding.

@schristley You are right that the `organism` keyword is part of the MiAIRR table, but as it is listed in the "Formats WG" column, so DataRep can could change it.

Or, we just do away with all that "my dance space, your dance space" division, and ask whether this change is a semantic one -- which IMO it is not. Therefore one could argue that it does not change MiAIRR in an incompatible manner and we do the change for field name and key in v1.3.

schristley:
> it does not change MiAIRR in an incompatible manner and we do the change for field name and key in v1.3.

great, let's do it!

PR #362, opened 2020-03-27
Updated the "name":"Organism"

bcorrie:
Should be name:Species


javh:
Not yet. See discussion in #266 and #359. 

This should be `name: Organism` for v1.3 and `name: Species` for v2.0.

javh:
From the call:
* Make the change in OpenAPI v3 schema.
* Then merge.

PR #363, opened 2020-03-30
MiAIRR_Elements.tsv adjustments

franasa:
Following the recent changes from #319, I have reviewed and adjusted the `conf.py` file to generate the `AIRR_Minimal_Standard_Data_Elements.tsv`.

As pointed in #359 species name is still organism in 1.3 but should be reviewed for 2.0

I haven't made any changes to the `data_elements.rst`  but after the changes made on the `conf.py ` the table is generated and stored in `_download` but it does not get rendered, @javh maybe you have an idea why is this?

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -320,6 +320,16 @@ data_elements = [["Set",
                   "Example",
                   "Requirement"]]
 
+
+# func to chop down long strings of the table
+def wrap_col(string, str_length=11):
+    if [x for x in string.split(" ") if len(x) > 25]:
+        parts = [string[i:i + str_length].strip() for i in range(0, len(string), str_length)]
+        return ('\n'.join(parts) + '\n')
+    else:
+        return (string)
+
+
 # iterate over first level of yaml items
 for key, v in airr_schema.items():
     # iterate over second level of yaml items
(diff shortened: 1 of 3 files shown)


javh:
If you run `sphinx-build -b html -a -E docs docs/_build/html` locally, you should see the following error:

```
airr-standards/docs/miairr/data_elements.rst:15: WARNING: "csv-table" widths do not match the number of columns in table (9).

.. csv-table::
   :file: ../_downloads/AIRR_Minimal_Standard_Data_Elements.tsv
   :delim: tab
   :header-rows: 1
   :stub-columns: 0
   :align: left
   :widths: 5,10,10,10,10,10,30,15
```

You were just missing another width item for the updated table. I pushed a fix, but I didn't play around with the widths at all to get it to look right.

(There's also a bunch of warnings from news.rst we should probably fix. Not sure how to fix those embedded comments off the top of my head.)

franasa:
> If you run sphinx-build -b html -a -E docs docs/_build/html locally, you should see the following error:

 So far I just use`make html` and I see some of the errors and warnings from Pycharm but that .rst was just giving me some env warning,  the column number was quite obvious though , thanks for the hint and the fixes!

franasa:
eb3f221 is an idea to shrink `Format` and `Example` columns

PR #364, opened 2020-03-30
x-airr settings for ADC API support

### LINKED ISSUE #183: Location of germline_database key(s) in AIRR schema
The current AIRR schema contains two `germline_database` keys, one in the `Alignment` another one in the `Rearrangement` object. However, this create a lot of redundancy, as usually all the `Rearrangements` in a `Rearrangement_set` would use the same reference. Therefore:

1. Is there any scenerio in which we would need a `germline_database` key below the `rearrangement_set` level?
2. Do we need more that a single reference, e.g. on for software processing and one to annotate a subject's individual germline reference (as suggested by @bcorrie)?

Note that this topic has been touched upon before in a number of other discussions like:
* https://github.com/airr-community/airr-standards/issues/157#issuecomment-451604385
* https://github.com/airr-community/airr-standards/pull/156#issuecomment-447919836

### LINKED ISSUE #340: merge rearrangement_id -> sequence_id
`rearrangement_id` was added to have a more global/database context than `sequence_id`, but now having two identifiers for different contexts is really just causing problems. The plan is to eliminate `rearrangement_id` and just use `sequence_id`.

- Data repositories/tools are allowed to overwrite `sequence_id` with their own value in order to satisfy uniqueness within their database/tool processing.

- Tools should not assume that `sequence_id` contains a value that references a sequence in the raw sequencing files.

@bcorrie As the other big data repository implementation, any concerns?

### LINKED ISSUE #346: Add deprecation of rearrangement_set_id to Rearrangement
@bcorrie bought up the point in #340 that we hadn't properly deprecated `rearrangement_set_id` and indicated that it was replaced by `repertoire_id`.

schristley:
resolves #346 
resolves #340 
resolves #183

### DIFF
--- a/docs/appendix/references.rst
+++ b/docs/appendix/references.rst
@@ -1,0 +1,30 @@
+.. _References:
+
+References
+==========
+
+.. [Breden_2017] Breden F *et al*. Reproducibility and Reuse of
+   Adaptive Immune Receptor Repertoire Data. Front Immunol 8:1418 (2017)
+   `DOI: 10.3389/fimmu.2017.01418`_
+
+.. [RFC2119] Key words for use in RFCs to Indicate Requirement Levels
+   `DOI: 10.17487/RFC2119`_
+
+.. [Rubelt_2017] Rubelt F *et al*. AIRR Community Recommendations for
+   Sharing Immune Repertoire Sequencing Data. Nat Immunol 18:1274
+   (2017) `DOI: 10.1038/ni.3873`_
+
+.. [Wilkinson_2016] Wilkinson MD *et al*. The FAIR Guiding Principles
+   for scientific data management and stewardship. Sci Data 3:160018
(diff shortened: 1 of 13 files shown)


schristley:
@javh @bcorrie this should be ready to go

javh:
Looks like there are two cached Repertoire schemas in the python API now? `RepertoireSchema`
and `CachedSchema['Repertoire']`. I don't have a strong opinion on how we structure the cached schemas, but we should probable be consistent.

Otherwise, looks good. I made some minor adjustments to the lineage schemas.

schristley:
> Looks like there are two cached Repertoire schemas in the python API now?

Ah yeah, I didn't think to replace usage of `RepertoireSchema`, but we can, I imagine it's only used a few spots. I needed to be able to lookup by name for validate_object(), and `CachedSchema` seemed the easiest way to do that.

PR #366, opened 2020-04-02
Add consistency to name/nullable attributes

bcorrie:
x-airr attributes for non MiAIRR fields in the spec are inconsistently represented. Added name and nullable fields for some that are missing.


schristley:
@bcorrie I'm going to merge this into #364 because that contains so many updates, and I don't want to have to deal with all the merge conflicts. Hopefully you and @javh will be able to finish your review of that soon and I can work on all the downstream changes to ADC API.

javh:
@schristley , yes, sorry. I'll try to scan through it tomorrow... Just had a lot going on this week. And, yeah, aggregating these would be great.

PR #375, opened 2020-04-08
/info response change

### LINKED ISSUE #367: /info ADC API entry point
A couple of things... 

Is there a reason why the /info entry point doesn't respond with the info_object object in the schema? It currently response with two text strings a name and version, while the info_object has a much more informative set of information...

Also, I was wondering if we could consider adding a "last_update" field to the info_object in the ADC API schema. It would not need to be mandatory, but if present would indicate the last time the repository data had changed.  This would be useful to us as a client of the ADC API, as one could easily check if a repository had added any data since the last time one checked.

Thoughts?

bcorrie:
Fixes #367


schristley:
@bcorrie I added info for the api and the schema.

bcorrie:
I'm OK to merge this in if you are good with it...

bcorrie:
Scott, can you review and merge?

PR #379, opened 2020-04-17
ADC documentation page

bcorrie:
I noticed that although we have an ADC API documentation page, we don't have anything that describes the AIRR Data Commons - that is the network of repositories that implement the ADC API... In trying to describe the AIRR Data Commons to someone in the outside world, I realized there was not a place in our documentation for this...

@schristley and @lgcowell do you want to have a look at this and review? Most of the text I stole from other docs, so if it looks familiar that is why...


schristley:
Yeah, we need a page dedicated to ADC. What about this [page](https://docs.airr-community.org/en/latest/standards/data_submission.html#data-query-and-download)?

It was setup awhile ago. Can you integrate the two? Maybe have that Query page link to the pages you created which has more detail?

schristley:
It would be nice too if the ADC page could link the [tutorial example](https://docs.airr-community.org/en/latest/examples/api_analysis_example.html). And to the example [queries](https://docs.airr-community.org/en/latest/examples/api_queries.html), though that query page needs to be filled out more.

bcorrie:
> Yeah, we need a page dedicated to ADC. What about this [page](https://docs.airr-community.org/en/latest/standards/data_submission.html#data-query-and-download)?
> 
> It was setup awhile ago. Can you integrate the two? Maybe have that Query page link to the pages you created which has more detail?

Yes, I grabbed the repository and query section from that all ready... Although maybe we don't need this in two places...

bcorrie:
> It would be nice too if the ADC page could link the [tutorial example](https://docs.airr-community.org/en/latest/examples/api_analysis_example.html). And to the example [queries](https://docs.airr-community.org/en/latest/examples/api_queries.html), though that query page needs to be filled out more.

Done...

schristley:
> Yes, I grabbed the repository and query section from that all ready... Although maybe we don't need this in two places...

That's the top-level section setup for data submission and query, we need to keep that. It's fine to have some duplication as that top-level section will want to give some overview for readers.

Presumably a user coming for the first time would first go to [Getting Started](https://docs.airr-community.org/en/latest/standards/getting_started.html). They would see and click on the link about [Querying Data](https://docs.airr-community.org/en/latest/standards/data_submission.html#data-submission). And then from there would be links to more detail about the AIRR Data Commons.

bcorrie:
> Presumably a user coming for the first time would first go to [Getting Started](https://docs.airr-community.org/en/latest/standards/getting_started.html). They would see and click on the link about [Querying Data](https://docs.airr-community.org/en/latest/standards/data_submission.html#data-submission). And then from there would be links to more detail about the AIRR Data Commons.

Have a look at how I restructured it, I think it makes sense...

PR #384, opened 2020-04-24
Update AIRR spec to use title attribute

### LINKED ISSUE #380: Deprecate MiAIRR field name
We have maintained a split between MiAIRR field names (which are human readable) and DataRep/Schema fields (which serve as unique and unambiguous keys) for the last two years. Given the recent decisions on Set 6/Rearrangement and the fact that everything is now represented in a single schema, this split appears unnecessary and potentially confusing. In addition, the `x-airr:name` property, which currently encodes the MiAIRR name, is used as short description for non-MiAIRR fields (e.g., `study_description`), which is illogical. As @schristley commented in [#369](https://github.com/airr-community/airr-standards/issues/369#issuecomment-614389193), this information should be in a property called `title`, the introduction of which would howver duplicate the information for MiAIRR fields. We should simply this.

Proposed changes:
* Move all `x-airr:name` properties to `title`
* Officially deprecate the MiAIRR field names. There are only Schema fields, which have a human-readable `title`. For MiAIRR fields, this title will initially be what is now the MiAIRR field name.

I think this should best go into v2.0, but open for discussion if there is a reason to push for v1.3.

bcorrie:
Fixes #380

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -137,6 +137,7 @@ Study:
         study_id:
             type: string
             description: Unique ID assigned by study registry
+            title: Study ID
             example: PRJNA001
             x-airr:
                 miairr: important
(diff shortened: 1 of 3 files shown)


bcorrie:
I think this is done. All fields with a name attribute now have a title attribute as well. So OK to do the merge, although this probably should not close #380 as we still need to deprecate the name field.

PR #385, opened 2020-04-27
Introduce CURIEs for ontologies

bussec:
Introduces the changes of OntoVoc sprint 04/2020 into the schema:
* Introduce CURIEs (including resolver table)
* Accept draft ontologies

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -19,12 +19,39 @@ Ontology:
     discriminator: AIRR
     type: object
     properties:
-        id:
+        curie:
             type: string
-            description: Identifier for the ontology term
-        value:
+            description: CURIE of the concept, encoding the ontology and the local ID
+        label:
             type: string
-            description: Value for the ontology term
+            description: Label of the concept in the respective ontology
+
+
+CURIEResolution:
+  - 
+    curie_prefix: NCBITAXON
+    curie_separator: ":"
+    iri_prefix: "http://purl.obolibrary.org/obo/NCBITaxon_"
+  - 
+    curie_prefix: NCIT
+    curie_separator: ":"
+    iri_prefix: "http://purl.obolibrary.org/obo/NCIT_"
+  - 
+    curie_prefix: UO
+    curie_separator: ":"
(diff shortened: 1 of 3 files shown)


bcorrie:
@bussec see my comment on my review about v1.3 versus v2.0.

bussec:
@bcorrie Fine with me if no one has any objections to still push into v1.3. I would assume that most tools are ontology-agnostic anyhow, so the changes for them are minimal.

bcorrie:
@schristley @javh any objections to making this update to the ontologies? My concern with not having it in v1.3 is that we will do a release with everything saying that we should use ontology IDs of the form NCBITaxon_9606 in v1.3 and then say in v2.0 that that was wrong and we should use IDs of the form NCBITAXON:9606

Given these are things that are likely stored in repositories and returned by APIs, I think we will be better served by having this correct up front with our first ontology release, which is v1.3

javh:
No objections if it's ready to go. 

(I think we need a feature freeze if we're ever going to get v1.3 out. Ideally, we tag v1.3 this upcoming week.)

schristley:
Are we talking just the underscore to colon change, or also changing the field names? I've already changed my data to use the colon, so that's fine with me. Changing the field names is more work, but not that hard to do either.

bcorrie:
> Are we talking just the underscore to colon change, or also changing the field names? I've already changed my data to use the colon, so that's fine with me. Changing the field names is more work, but not that hard to do either.

I  think the more important one is to get the docs synced with the content, which is the colon separated CURIE ID and the ontology ID, but I think we are probably better off to go with the name change as well (label -> value) since label never made it into a released spec (v1.2). So best not to have it ever appear in a release at all if it is going to go away???

bussec:
> but I think we are probably better off to go with the name change as well (label -> value) since label never made it into a released spec (v1.2) [...]

@bcorrie You mean `value` to `label`?

bcorrie:
> > but I think we are probably better off to go with the name change as well (label -> value) since label never made it into a released spec (v1.2) [...]
> 
> @bcorrie You mean `value` to `label`?

Yep... 8-)

bussec:
Reset the target branch to `master` plus some fix-ups. If we want to merge this, we would be all set now.

PR #389, opened 2020-04-30
Add clarifying text and example to CIGAR docs

### LINKED ISSUE #209: CIGAR string documentation needs more clarification
From a discussion with @williamdlees with regard to partis and OGRE

> The one area of the spec that I think really does need some amplification is the CIGAR string.  An example from the spec is 'CIGAR string for the J gene alignment.'  It should specify the two sequences that are to be aligned, and state which is the query and which is the reference. Partis has implemented the J-region as reference and the sequence_alignment as query, but the spec is ambiguous: it could be interpreted as an alignment between the J-region and the J- germline gene, or the J-germline gene and sequence_alignment, and query and reference could be either way around. The spec should also say what's expected if there are gaps in sequence_alignment (such as IMGT gaps), and in particular whether there are ever any circumstances in AIRR data in which the CIGAR could contain more than one matching region separated by non-matching regions.

javh:
Resolves #209.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -254,7 +254,7 @@ if not os.path.exists(dl_path):  os.mkdir(dl_path)
 
 # Build table for each spec
 tables = ['Rearrangement', 'Alignment']
-fields = ['Name', 'Type', 'Priority', 'Description']
+fields = ['Name', 'Type', 'Level', 'Description']
 for spec in tables:
     # Get specs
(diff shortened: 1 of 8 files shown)


javh:
@schristley. Made a few more changes. Mostly minor formatting stuff. But, one important change is that I removed the toctree link to the Germline experimental schema as this seemed to empty.

Issue #397, opened 2020-05-16
MiAIRR table for v1.2 release (stable) is broken!!!


bcorrie:
Hi All,
I just noticed that the link to the table on this page:

https://docs.airr-community.org/en/stable/miairr/introduction_miairr.html

is broken. As a result, I don't think there is a way to get to the table of AIRR terms in the current spec on the stable docs branch... Not sure how long this has been broken!

bcorrie:
I notice that the table is linked to the TSV file on master branch - which no longer exists... I thought the stable branch was linked to the 1.2.1 tag (the most recent numerical tag). Not sure how these things get pushed to docs site - so not sure what the right fix is.

bcorrie:
I guess this will be less important when we release v1.3, but we should probably still fix?

javh:
Duplicate of #393. 
I'm working on it. Didn't have time during the week to finalize the fixes. It builds in the docs-tables branch though.

javh:
I guess we could restore this link by putting the old TSV back into master temporarily, but yuck.

bcorrie:
Probably not worth fixing for 1.2 and just fix through the v1.3 release???

Issue #393, opened 2020-05-09
ReadTheDocs build failing on master


javh:
RTD builds are failing with:

```
There is a programmable error in your configuration file:

Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/airr-standards/envs/latest/lib/python3.7/site-packages/sphinx/config.py", line 368, in eval_config_file
    execfile_(filename, namespace)
  File "/home/docs/checkouts/readthedocs.org/user_builds/airr-standards/envs/latest/lib/python3.7/site-packages/sphinx/util/pycompat.py", line 150, in execfile_
    exec_(code, _globals)
  File "/home/docs/checkouts/readthedocs.org/user_builds/airr-standards/checkouts/latest/docs/conf.py", line 347, in <module>
    airr_name = airr_properties[airr_property]["x-airr"]["name"]
KeyError: 'name'
```

Seems to be due to replacing `x-airr: name` with `title`.

javh:
Simple fix didn't work, so I started work on this and #390 in the `docs-tables`  branch. Give me a few days.

franasa:
I'm sorry @javh , It was on me to do that fix but I was pushing it till `name` was removed. The MiAIRR table is rendering now and your lines look much better. Let me know if I can still help somehow.

javh:
@franasa, no worries. I'm still working on it, but I think I know what needs to be done. My goal is to merge the three bits of table generation code into a single version based off yours (which is why I moved it around a bit).  I'll definitely ask you to review it once I'm done.

PR #398, opened 2020-05-17
Refactor documentation table generation

### LINKED ISSUE #390: Inconsistency between Rearrangement and metadata field tables in docs
Sorry I didn't catch this earlier, but the field tables aren't consistent across the Rearrangement, metadata (Repertoire), and experimental schema (Alignment, Clone, Tree, Node, Cell).

Rearrangement and Alignment have columns:  Name, Type, Priority, Description. I changed "Priority" to "Level" in #389. Where Level is either `required` (in top level `required`) or `optional` (not in top level `required`). 

Metadata objects and Clone, Tree, Node and Cell have "MiAIRR" instead of "Level".

The former also have a **Download as TSV** link and the latter do not.

I think these should be consistent. At the very least Clone, Tree, Node and Cell should follow whatever rules we use for Rearrangement and Alignment. I can see an argument for making the metadata field tables different. Though, I'm not certain "MiAIRR" is the right terminology to the convey the requirement levels. We can populated the column from the `x-miairr` tag, but I don't think the distinction between MiAIRR and not-MiAIRR is very useful.

Thoughts? I'm guessing this is partly due to us having some redundant code in `docs/conf.py` for building tables.

javh:
Resolves #390, #393 and #397.

I've tried to merge all table generating code for the docs into a single approach. Previously, there were separate code chunks for the Rearrangement/Alignment tables, metadata tables, and MiAIRR tables.

They didn't all work the same, so I tried to compromise where the implementations differed. Not sure if I got the ontology stuff correct, in particular.


javh:
I enabled building of this branch on RTD temporarily:
https://docs.airr-community.org/en/docs-tables/

bussec:
Thanks @javh ! The ontology stuff looks ok, we will need to see whether it breaks if we merge #385. There is still an issue with the `cell` object at the bottom of the table, we'll look into this tomorrow.

javh:
Sounds good. Oh, I forgot to mention, I cut the MiAIRR table off at Set 5 - I wasn't sure which specs, if any, from Set 6 should be in that table.

javh:
Okay, so can we merge in #385 first, then update this branch with @franasa's fixes?

schristley:
The ontology link isn't showing in the fields tables, e.g. `study_type` has a type of `string` when it should be `Ontology`

bussec:
@schristley Are we talking about the same table? Because [here](https://docs.airr-community.org/en/docs-tables/miairr/data_elements.html) the ontology is present (the same is true for the `master` [branch tip of the doc](https://docs.airr-community.org/en/latest/miairr/data_elements.html))...

schristley:
nope, talking about these [tables](https://docs.airr-community.org/en/docs-tables/datarep/metadata.html). Somehow, the deprecated `organism` field still has the ontology link but not the other fields.

javh:
I think this is one of those cases where the MiAIRR table and the metadata schema tables didn't use the same format for the "Type" column.

E.g., the MiAIRR table `study_type` is:

**Type:** string
**Format:** Ontology: { name: NCIT , top_node: { id: C63536, value: Study}, draft: True}

Metadata schema tables `study_type` is:

**Type:** [Ontology](https://docs.airr-community.org/en/docs-tables/ontovoc/introduction_ontovoc.html#ontovoc)

schristley:
I think the metadata tables need to say `Ontology` with link to ontology page, because having the type listed as `string` is confusing and not correct as the actual type is an `object`.

bussec:
> I picked the MiAIRR format. Not sure if that's ideal. What's the preference?

Linking to the Ontovoc table has probably a better readability, so I agree with @schristley. We could refer to the individual ontologies via `#fragment` in the URL if necessary.

javh:
I made the change to the Ontology type, minus the `#fragment` bit.

bussec:
So we merge #385 now and finalize this tomorrow?

javh:
Works for me.

PR #399, opened 2020-05-20
Merge ADC docs into master

bcorrie:


schristley:
@bcorrie this should all be merged already, unless you had additional changes to add? In which case, you might want to create a separate branch to avoid the rebase conflicts.

bcorrie:
> @bcorrie this should all be merged already, unless you had additional changes to add? In which case, you might want to create a separate branch to avoid the rebase conflicts.

Yeah this is weird - I think the way the rebasing was done perhaps the old branches weren't deleted, so if you look at the branches it looks like they haven't been integrated but if you check closely the commits to the branch are duplicated on master (which is the rebase I assume).

If you look at the branches here: 

https://github.com/airr-community/airr-standards/network

it looks like there are a bunch of branches that haven't been merged in...

Not sure I like the rebase approach, as it makes it very difficult to tell what has been merged and what hasn't...

schristley:
@bcorrie I'm not positive, but I guess that network view is confusing because I didn't delete the branch after the merge. I know that we had a few cases where github couldn't rebase automatically because of conflicts, and the merges had to be done manually...

schristley:
@bcorrie for this particular PR, you can see in the "Files changed" tab that no changes are shown, so there is nothing to merge...

schristley:
@bcorrie yeah, it looks like the network view is confusing when the branch is left after doing the merge...

Issue #219, opened 2019-08-06
additional documentation for library_generation_method


schristley:
MiAIRR field `library_generation_method` has a controlled vocabulary for acceptable values. These values are concise and use acronyms that users may not immediately recognize, e.g. `RT(specific)+TS+PCR`. We need additional documentation that explains each of the values in more detail.

bussec:
The terms are explained in [`NCBI_implementation/mapping_MiAIRR_SRA_library_terms.rst`](https://github.com/airr-community/airr-standards/blob/master/NCBI_implementation/mapping_MiAIRR_SRA_library_terms.rst), but I agree that they might be better placed in the docs. Any specific location to which you would like this to move?

schristley:
The rearrangement schema doc actually lists out all of the fields and puts them in a table. It does this by reading the schema file. We should do something similar for the repertoire metadata, in which case the explanation can go in one of two places:

- In the schema itself, like the `description` field
- In a `Definitions Clarification` section like the rearrangement schema has

schristley:
@bussec I've put in some rough work to generate field tables for all of the metadata schema objects and their fields.

http://docs.airr-community.org/en/metadata-docs/datarep/metadata.html#repertoire-fields

We should think of a few things to enhance:

- How to signify ontology-based and controlled_vocabulary fields? For the controlled_vocabulary (enum) we should list the acceptable values in the doc.

- Is it okay to have a `Definitions Clarification` that is only in the docs, or should it also be in the schema somewhere?

- Some of the descriptions are quite long, e.g. `complete_sequences`. Maybe this should be shortened, and the extra detail added to `Definitions Clarification`?

PR #402, opened 2020-05-20
Doc layout tweaks

javh:
I made a few tweaks to the doc and how the MiAIRR table is rendered to try to make it more readable. It's still a little messy, but maybe this is better?

The primary changes are that I merged some of the MiAIRR table columns in the html build of the docs, but left them separate when you download the TSV.

I also swapped from Bootstrap 2 to 3 (fixes the search bar dropping down). I didn't change the theme, but that's also something we could do if we prefer the look of something else (with smaller table font sizes maybe). Theme options are here:
https://bootswatch.com/3/

**Note:** I don't think we need this for the v1.3 release. Just throwing it out there for opinions.


javh:
Branch build enabled on RTD:
https://docs.airr-community.org/en/doc-tweaks/

schristley:
That search box is messing up the navigation bar again.

javh:
> That search box is messing up the navigation bar again.

On which page? (Could just be my browser/resolution likes it better)

schristley:
https://docs.airr-community.org/en/doc-tweaks/miairr/introduction_miairr.html

https://docs.airr-community.org/en/doc-tweaks/datarep/overview.html

schristley:
ah yeah, I'm getting it with Safari but it looks okay in Chrome

bussec:
Same problem here, both on Firefox and Chromium.

bussec:
And the references are screwed up:
https://docs.airr-community.org/en/doc-tweaks/appendix/references.html
https://docs.airr-community.org/en/doc-tweaks/miairr/specification_miairr_ncbi.html#references

javh:
Okay, I think the search bar thing is due to the text wide. The extra AIRR-C and GitHub links mess up the spacing of how the top bar shrinks with the browser window size. I think.

I'll see if I can find a fix.  I swapped it back to bootstrap 2... Does it still look funny? (It still happens to me as I shrink the window.)

schristley:
with bootstrap 2, the search box does get pushed onto a second line for some pages, but at least it's not overwriting the header line in the content, the header is still readable.

It seems like that navigation bar needs to be given just a little more width. Even though I make my browser window wider, and there is plenty of white space on both left and right margins, the nav bar doesn't stretch out to use the extra space.

javh:
Any opinions on the changes to the layout of the MiAIRR table?

bcorrie:
@javh where does the `boolean T|F` come from? In particular the `T|F` part. Does that get generated some how automagically?

Correct json would be `true|false`, these fields don't have examples (which would be either `true` or `false`) and the description uses TRUE in all caps...

Wondering if it would be worth making this consistent and probably best to use the JSON values for this?

javh:
Yeah, it's generated in `docs/conf.py` here:
https://github.com/airr-community/airr-standards/blob/doc-tweaks/docs/conf.py#L263

Yeah, it should be consistent. `T` and `F` is correct for the TSV formats. Do we want the MiAIRR table to follow the yaml/json formats?

bcorrie:
Not sure, too bad they aren't the same 8-) There is no correct answer I suppose... @bussec any opinions since it is a MiAIRR table?

javh:
If I remove the AIRR-C and GitHub links and shrink the search bar width to 60px it seems to behave for me. I don't like removing the links though.

I tried messing with the viewport size, but I didn't do it right an it just messed up the dynamic resizing. I'm sure we can adjust it - I just have to figure out how. 

I'm inclined to push out the release and struggle with this more for the next patch release. We could also switch to something like the [PyData](https://pydata-sphinx-theme.readthedocs.io/en/latest/), which doesn't get us quite as much horizontal width, but otherwise fits our needs.

schristley:
> I'm inclined to push out the release and struggle with this more for the next patch release.

works for me.

bussec:
> Do we want the MiAIRR table to follow the yaml/json formats?

As it MiAIRR primarily describes `Repertoire` data, which is provided as YAML, this would IMO make sense. But I agree that this is a cosmetic issue that should not block the release.

franasa:
::thumbs-up::

I've checked out locally some of the B3 themes, Flatly, Lumen and Yeti look good in my opinion but by default don't represent any change in how the tables are rendered.

> We could also switch to something like the PyData theme, which doesn't get us quite as much horizontal width, but otherwise fits our needs. 

I also tried the PyData-theme with its defaults and it looks IMHO much better in general (but would need some changes i.e. to have a custom navbar), but also the tables as it embeds the page in a sub-area that can be scrolled in both axes. I agree it's a good idea to have more opinions on this.
It would also probably make sense to shorten some of the definitions as pointed by Scott  [here](https://github.com/airr-community/airr-standards/issues/219#issuecomment-530526218), 

> Some of the descriptions are quite long, e.g. complete_sequences. Maybe this should be shortened, and the extra detail added to Definitions Clarification?

for a better look in a further release.

javh:
I like BS3 Yeti and PyData. I also like the Executable Book extension of the PyData theme quite a bit:
https://sphinx-book-theme.readthedocs.io/en/latest/

franasa:
the color palette would also fit quite good with the AIRR logo if it is going to be included.

javh:
Ooo. I like it with the logo!

bussec:
I like sphinx-book-theme, can we keep it? :-D

javh:
All right, let's keep the book theme. :)

I want to make some little changes to the toctrees and xrefs and then I'll merge these in. (Like making Rearrangement and Repertoire more visible.)

javh:
@schristley, where is api/adc.rst supposed to live?

```
airr-standards/docs/api/adc.rst: WARNING: document isn't included in any toctree
```

schristley:
> where is api/adc.rst supposed to live?

It should be under "Data Submission and Query"

schristley:
@javh and if possible, get the AIRR TSV paper in the References page.

schristley:
and even though the final proofs aren't up, we have a web link and DOI for the ADC API paper.

https://www.frontiersin.org/articles/10.3389/fdata.2020.00022/abstract

javh:
> It should be under "Data Submission and Query"

There's a link from there, but... Do you want me to copy all that content on the same page? Having a single entry toctree on the page would be weird.

We could also put it in the Community toctree?

schristley:
> There's a link from there, but... Do you want me to copy all that content on the same page? Having a single entry toctree on the page would be weird.

I think a single entry is fine for now. We can expand that out later to include data submissions entries and germline db stuff

javh:
If every is good with these changes, I'm ready to merge this in and tag the release.

PR #404, opened 2020-05-22
Introduce Receptor schema

### LINKED ISSUE #44: epitope and specificity information
This is an issue that Bjoern brought up on the CRWG mailing list, whether part of MiAIRR or just support in formats is worth discussion.

> The alternative is to describe the exact epitope recognized as part of the MIARR standard. That is absolutely possible, but will require a substantial update -  I thought the point of using the IEDB record was that this takes away that complexity. If the epitope should be described, minimally you will need 
> - epitope type (linear peptide, discontinuous amino acids, non-peptidic)
> - epitope molecular structure (3 different fields, depending on type above)
> - epitope source protein
> - epitope source organism
> 
> There are a bunch of additional fields (e.g. is this the exact epitope / an epitope containing region / a partial epitope). But as a start you might want to err on the minimal info above. 

> The primary use case at that point was that people are using e.g. MHC tetramer staining to isolate epitope specific repertoires.

### LINKED ISSUE #207: Single-cell reactivity data annotation/linkage

### LINKED ISSUE #293: Representation of bispecific antibodies
This was briefly mentioned by @javh in yesterday's DataRep call:

How can we represent immunoglobulins that consist of more than one heavy and one light chain species. Our current concept of `paired_id` assumes that we combine two chains into a single Ig.

This would most likely happen in things like bispecific constructs, but it could also apply to immunoglobulins produced by B cells with two productive light chains (although it is harder the verify the actual existence of the latter one).

### LINKED ISSUE #410: Receptor, Cell and Rearrangement schema representations
**Receptor**

A receptor represents two chains(i.e. one heavy and one light) of a cell, each with a antigen-binding site. However, an individual cells can also express multiple receptors, e.g. due to simultaneous expression of two TCR alpha chains. 

Aim of the ```Receptor``` object is to provide reactivity information for individual Ig/TCRs. 
```Receptor``` object which is currently available in the pull request #404. 

However, if the amino acid sequences of V-region of both receptor chains are identical and grouping them together will be considered as identical receptors and they would have identical reactivity information. So, the question was how to represent them in order to avoid redundancy of identical ```Receptor``` object?

**Relationship between Receptor, Cell and Rearrangement**

The current schema for ```Cell```, ```Rearrangement ```and ```Receptor``` is represented below  

• ```Receptor``` has N: N with ```Cell``` – A cell would contain multiple receptors and a receptor can be from multiple cells

• ```Receptor``` has 1:2 with ```Rearrangement``` –Two rearrangement represents a receptor and a receptor contains only two rearrangements sequences.

However the current linkage between ```Receptor``` and ```Rearrangement``` breaks because the rearranged sequences associated with the receptor might be different as receptors are determined at amino acid level whereas rearranged sequences are determined at nucleotide level there by having any synonymous substitution would lead to identical V-region of both receptor chains but their rearrangements would not be identical. Hence, 2:1 ```Rearrangement```:```Receptor``` does not hold true.

Given the schema representations that allows N:N Cell:Receptor relationship we have come up with two schema proposals and would be good to get some suggestions/comments on the proposed schemas below. 

**Proposal 1:**

In the proposed schema we define two ```Receptor``` objects i.e ```Intermediate``` and ```Observed Receptor``` where multiple ```Intermediate Receptor``` objects are related to a single ```Cell ``` object and a ```Observed Receptor``` object is linked to multiple ```Intermediate Receptor``` object and there are no direct linkage to ```Cell``` Object.  

The ```Observed Receptor``` object will then contain same properties as what we have defined currently in #409 but with addition of amino acid sequences of the receptor chains.

**Proposal 2:**

• ```Receptor``` has N:1 with ```Cell``` – A cell would contain multiple receptors and a receptor can be present in single cell

Here we then do not link``` Receptor``` to ```Rearrangement``` but as proposed for ```Observed Receptor``` we should extend the current```Receptor``` object by addition of the protein sequences of both receptor chains

### LINKED ISSUE #540: IDs for Receptor objects
_Originally posted by @schristley in https://github.com/airr-community/airr-standards/issues/524#issuecomment-868818350_

IEDB has [released](https://help.iedb.org/hc/en-us/articles/4402872882189-Immune-Epitope-Database-Query-API-IQ-API-) a beta query API for their database.

Of interest is they have a [curie_map endpoint](https://query-api.iedb.org/curie_map) which returns a structure similar to our `CURIEMap`. I've noticed however that the IRI goes to the human-readable html page instead of the API which returns machine-readable JSON. For example, `IEDB_EPITOPE: 7355` resolves to:

```
https://www.iedb.org/epitope/7355
```

versus

```
https://query-api.iedb.org/epitope_search?structure_id=eq.7355
```

The nice thing is that now we should be to add a single field in Rearrangement if we want to link to an epitope #44 

Though that opens the question on whether we should put entries into our CURIEMap for resolving `IEDB_EPITOPE` or if we should use IEDB's?

We also can consider how we might link with receptors. The IEDB API has `tcr_search` and `bcr_search` endpoints:

```
https://query-api.iedb.org/tcr_search?limit=5
https://query-api.iedb.org/bcr_search?limit=5
```

The receptor ids have `IEDB_RECEPTOR` as their CURIE though oddly it's missing from the above curie_map. However, we might want to consider how we can link our `Receptor` to IEDB's.

sharikrish:
Fixes #44
Fixes #207
Fixes #293
Fixes #410
Fixes #540

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -408,7 +408,7 @@ tables = ['Repertoire', 'Study', 'Subject', 'Diagnosis', 'Sample', 'CellProcessi
           'PCRTarget', 'SequencingRun', 'RawSequenceData', 'DataProcessing',
           'Rearrangement', 'Alignment', 'Clone', 'Tree', 'Node', 'Cell', 'CellExpression',
           'RearrangedSequence', 'UnrearrangedSequence', 'SequenceDelineationV', 'AlleleDescription', 'GermlineSet',
-          'GenotypeSet', 'Genotype', 'MHCGenotypeSet', 'MHCGenotype']
+          'GenotypeSet', 'Genotype', 'MHCGenotypeSet', 'MHCGenotype', 'Receptor']
 for spec in tables:
     with open(os.path.join(download_path, '%s.tsv' % spec), 'w') as f:
         writer = csv.DictWriter(f, fieldnames=fields, dialect='excel-tab', extrasaction='ignore')
--- a/docs/datarep/receptor.rst
+++ b/docs/datarep/receptor.rst
@@ -80,3 +80,26 @@ dual IGK/IGL expressing B cells their individual reactivity is
 not measured on a regular basis. Therefore they are currently not
 supported in the ``Receptor`` schema.
 
+
+.. _ReceptorFields:
+
+Receptor Fields
+-----------------------------
+
+:download:`Download as TSV <../_downloads/Receptor.tsv>`
+
+.. list-table::
+    :widths: 20, 15, 15, 50
+    :header-rows: 1
+
+    * - Name
+      - Type
+      - Attributes
+      - Definition
+    {%- for field in Receptor_schema %}
+    * - ``{{ field.Name }}``
+      - {{ field.Type }}
+      - {{ field.Attributes }}
+      - {{ field.Definition | trim }}
+    {%- endfor %}
+
--- a/docs/datarep/receptor.rst
+++ b/docs/datarep/receptor.rst
@@ -6,35 +6,77 @@ Receptor Schema (Experimental)
 General
 -------
 
-Receptor data SHOULD provided if it is non-trivial, i.e., if some
-information or reactivity is available (either locally or via an
-external identifier). If no information about the receptor, besides
-the translated amino acid sequence of the associated chains it
-available, the respective object MAY be created if the curator
+The purpose of the ``Receptor`` object is to represent reactivity data
+of actual *Receptors*, i.e., Ig or TCR, either directly or indirectly.
+To this end, the ``Receptor`` object has two main sections:
+
+1. The ``receptor_*`` properties: These properties describe the receptor
+   as an abstract and global concept, i.e., the actual Ig/TCR protein
+   complex MAY or MAY NOT have been observed in the current study.
+   However, the Rearrangements encoding the respective chains MUST
+   be present in the studies as well as the information linking them
+   (see below). This allows data curators to provide potential
+   reactivities of a potential receptor by referencing to its entry in
+   an external database (e.g., IEDB).
+2. The ``reactivity_measurements`` array: This structure contains
+   one or more entries, describing reactivities of the receptor that
+   have been observed in the current study.
+
+The ``Receptor`` object explicitly requires full sequence information
+of the two associated variable domains. This is considered to be an
+acceptable restriction from an AIRR-seq perspective, where sequencing
+typically precedes or takes place in combination with the determination
+of receptor reactivity.
+
+
+Relations to other AIRR Schema objects
+--------------------------------------
+
+The ``Receptor`` object is only directly linked to the ``Cell`` object,
+which then in turn contains the references to the records in the
+``Rearrangements`` that encode the respective chains of the receptor.
+Therefore a given rearrangement cannot directly reference to a receptor,
+which is also not a meaningful thing to do, as the paired chain would
+be unclear, but is necessary to determine a receptors reactivity.
+
+
+Annotation guidelines
+---------------------
+
+Reactivity information SHOULD be provided if it was either directly
+recorded in the current study or is indirectly available in a public
+database, where it can be linked. If no information about the receptor,
+besides the translated amino acid sequence of the associated chains is
+available, the respective object MAY still be created if the curator
 considers this information to be beneficial, e.g., if there is some
 evidence (surface expression, reaction to superantigens) that the
-receptor is functional and present on the surface.
-
-
-The AIRR Receptor object explicitly requires full sequence
-information of the associated V regions. This is considered to be an
-acceptable restriction from an AIRR-seq perspective, where sequencing
-typically precedes or happens in combination with the determination of
-reactivity.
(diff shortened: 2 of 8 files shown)


scharch:
In addition to adding the amino acid sequences to `Receptor` (per #409), I think you will want to identify which is the heavy/beta/delta and which is the lambda/kappa/alpha/gamma. Relatedly, it might be useful to have a top-level property that defines what type of `Receptor` this is.

For BCRs, is isotype a property of the `Receptor` or of the `Cell`? I lean toward the latter, since it doesn't affect antigen binding (and the same `Receptor` could be observed in multiple isotypes), but if the use case is synthesizing a `Receptor` of interest, it might be useful to have that information there...

bcorrie:
All, should we be adding a /cell and /receptor ADC API entry point as part of this schema change? This seems logical to me. I can make the changes if you give me the thumbs up, since I recently did this for /clone...

bcorrie:
I have created #466 as a pull request to integrate the /cell API endpoint. There is no /receptor endpoint as of yet.

bcorrie:
> I have created #466 as a pull request to integrate the /cell API endpoint. There is no /receptor endpoint as of yet.

@bussec @sharikrish @schristley do you want to review? From the API perspective, it is relatively trivial to add the endpoint. This assumes that the `Cell` object data is what the API will return.

bussec:
@bcorrie I moved the `Cell` related changes in this PR to #466 now. Unless there is any reason to implement `/cell` and `/receptor` simultaneously that (i.e. 466) should be good to go.

bussec:
Rebasing to avoid conflicts in `schema.yaml`...

schristley:
With IEDB's new query API, we can consider how we might link our `Receptor` with IEDB's. Some additional information is in the [OntoVoc comment](https://github.com/airr-community/airr-standards/pull/524#issuecomment-868818350).

bcorrie:
Couple of minor comments:

- Me thinks `reactivity_unit` should be a UO Ontology field???
- It would be nice if the receptor_gid would follow the same GID protocol (yet to be defined) of other GIDs - although your GID generation is kinda cool 8-)

bussec:
@bcorrie 

> It would be nice if the receptor_gid would follow the same GID protocol (yet to be defined) of other GIDs - although your GID generation is kinda cool 8-)

We can replace the `_gid` field with something like a UUID and rename the current fields to `_hash`. But I think it is important to  have this functionality (ids that are identical across repositories).

bussec:
@bcorrie @scharch @schristley This is now feature-complete from our side, please have a look.

bcorrie:
I don't have any objections to adding this to the v1.4 spec - as long as you science-y types are OK with the definitions etc. 8-)

In addition, if we have this object defined in the v1.4 spec, I would suggest we add it as an API endpoint in ADC V1.2 release (as discussed in #579) @schristley any objections to me adding this to the ADC API YAML spec? Repositories would need to implement, but they just return no data if they don't store Receptor object in their repository...

scharch:
Naive question: what does 
`$ref: '#/Ontology'`
mean?

Otherwise, I think it looks fine.

bussec:
@scharch 

> what does $ref: '#/Ontology' mean?

Basically, this is an `include` statement that copies the `Ontology` object (line 18ff. of the schema) into the position of the `$ref`. See [here](https://swagger.io/docs/specification/using-ref/) for details.

bussec:
> In addition, if we have this object defined in the v1.4 spec, I would suggest we add it as an API endpoint in ADC V1.2 release (as discussed in https://github.com/airr-community/airr-standards/issues/579)

Has now been added in [fcdfdd3](https://github.com/airr-community/airr-standards/pull/404/commits/fcdfdd3360f5adcdf22138fefc1abfe05b438487).

bcorrie (approved):

Issue #421, opened 2020-06-25
File format for AIRR Clone object?


bcorrie:
Have we come to a consensus as to what the file format for the `Clone` object should be? I know @eharkins has done some validation on the AIRR Clone data model, did you use the JSON format for the object for that validation. I think so as I recall some discussion about JSON in #323. This is also relevant to validation in the python and R libraries in #328. 

I suppose I am wondering if we are intending on supporting a TSV file format for Clones as well, as a compact/simple format? I believe at least the Clone object is very flat, simple object much like rearrangements. Are we intending on supporting a Clone TSV file? Did I miss such a discussion? 8-)

scharch:
I can't find it now, but I'm pretty sure we decided it would be JSON/YAML because each `Clone` can reference multiple `sequence_id`s and multiple `Tree`s, and we don't like nesting in TSVs :)

javh:
That's my recollection as well. 

Though, I do think we talked briefly about a TSV without the `sequences` field and with a single Newick string in the `tree` column  (one consensus tree, with nodes identifier by `sequence_id`) as a simplified down format. There would be not nesting required, but you'll need the accompanying Rearrangement data to get the sequence info at the Node level.

eharkins:
@bcorrie yes, we use JSON in Olmsted; [e.g.](https://github.com/matsengrp/olmsted/blob/master/example_data/full_schema_dataset.json#L48)

bcorrie:
OK - we are looking at loading clone data into the iReceptor repositories so they are searchable via the ADC API  (see #422) - so loading JSON/YAML it is... 

So one related question. This would imply that we are assuming (hoping?) that most software tools that produce clones would support this format as a data interchange format?

eharkins:
I'm not sure if this helps/answers but the most common (the only for now)  pipeline that includes Olmsted also includes a python utility to read clone info in the format it is produced (by the clone inference tools) and shape it into the Olmsted schema in JSON format.

scharch:
When I eventually get around to implementing this in SONAR, it will produce the AIRR-compatible YAML file, yes.

bcorrie:
> So one related question. This would imply that we are assuming (hoping?) that most software tools that produce clones would support this format as a data interchange format?

I guess another way of asking this question is, what are the most widely used Clone (in the sense of the AIRR Clone data model) generating software? Since we don't support clones yet, I am far from being an expert on this.

[...]

PR #422, opened 2020-06-25
Clone api

bcorrie:
Pull request to discuss addition of the Clone API

### DIFF
--- a/specs/adc-api.yaml
+++ b/specs/adc-api.yaml
@@ -370,6 +370,41 @@ paths:
           schema:
             $ref: '#/definitions/error_response'
 
+  /clone/{clone_id}:
+    x-swagger-router-controller: clone
+    get:
+      description: >
+        Returns a clone based upon its identifier.
+      tags:
+        - clones
+      consumes:
+        - application/x-www-form-urlencoded
+        - application/json
+        - application/octet-stream
+      produces:
+        - application/json
+      parameters:
+        - name: clone_id
+          in: path
+          description: ID of clone to return
+          required: true
+          type: string
+      operationId: getClone
+      responses:
+        '200':
+          description: >
+            A successful call returns the clone annotation data.
+          schema:
+            $ref: '#/definitions/clone_response'
+        '400':
+          description: Invalid request
+          schema:
+            $ref: '#/definitions/error_response'
+        '500':
+          description: Server error
+          schema:
+            $ref: '#/definitions/error_response'
+
   /clone:
     x-swagger-router-controller: clone
     post:
--- a/specs/adc-api.yaml
+++ b/specs/adc-api.yaml
@@ -124,7 +124,7 @@ definitions:
     properties:
       Info:
         $ref: '#/definitions/info_object'
-      Rearrangement:
+      Clone:
         $ref: '#/definitions/clone_list'
       Facet:
         $ref: '#/definitions/facet_list'
--- a/specs/adc-api.yaml
+++ b/specs/adc-api.yaml
@@ -111,7 +111,24 @@ definitions:
         $ref: '#/definitions/rearrangement_list'
       Facet:
         $ref: '#/definitions/facet_list'
-  
+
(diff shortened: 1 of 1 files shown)


bcorrie:
@schristley I notice a possible problem with the /rearrangement/{sequence_id} and /repertoire/{repertoire_id} endpoints in that the response object is the same as the /rearrangement and /repertoire endpoints. I don't think this is correct, as it contains a `Facets` JSON block that is not possible to get with the {_id} endpoints. I think this should be fixed, not sure if we should fix it through this pull request or not.

schristley:
The /clone/{clone_id} endpoint implies that `clone_id` has a uniqueness scope similar to repertoire_id.

schristley:
> notice a possible problem...

For the /clone endpoint, the response also suggests that `Facet:` and `Clone:` can be returned in the same response, which shouldn't happen, will that be fixed too?

bcorrie:
> The /clone/{clone_id} endpoint implies that `clone_id` has a uniqueness scope similar to repertoire_id.

Based on discussions today at DataRep, I would think of it more like sequence_id for /rearrangements. Unique within a repository...

bcorrie:
> > notice a possible problem...
> 
> For the /clone endpoint, the response also allows for `Facet:` and `Clone:` to be returned in the same response, which shouldn't happen, will that be fixed too?

Seems like one would want to perform the same types of counting operations that Facets provides for Clones, just like Rearrangements, so I would think that would be fine... That is why I left it in...

schristley:
> > The /clone/{clone_id} endpoint implies that `clone_id` has a uniqueness scope similar to repertoire_id.
> 
> Based on discussions today at DataRep, I would think of it more like sequence_id for /rearrangements. Unique within a repository...

Yes, that too is similar to repertoire_id. You made this recent [comment](https://github.com/airr-community/airr-standards/issues/421#issuecomment-651372216) so I wasn't sure of your intent.

> the clone_id is unique within a Repertoire, or perhaps more precisely within a repertoire_id/data_processing_id pair

schristley:
> > > notice a possible problem...
> > 
> > 
> > For the /clone endpoint, the response also allows for `Facet:` and `Clone:` to be returned in the same response, which shouldn't happen, will that be fixed too?
> 
> Seems like one would want to perform the same types of counting operations that Facets provides for Clones, just like Rearrangements, so I would think that would be fine... That is why I left it in...

I agree, though the size of the clone data is on the order of rearrangements, so we might want to consider similar restrictions. However, that wasn't what I meant. `clone_response` allows a response to contain both `Facet` and `Clone` like this:

```
{
  Facet: [ stuff snipped ],
  Clone: [ stuff snipped ]
}
```

Is that valid? I was assuming not, so asking if that will be fixed too?

bcorrie:
> I agree, though the size of the clone data is on the order of rearrangements, so we might want to consider similar restrictions. However, that wasn't what I meant. `clone_response` allows a response to contain both `Facet` and `Clone` like this:
> 
> ```
> {
>   Facet: [ stuff snipped ],
>   Clone: [ stuff snipped ]
> }
> ```
> 
> Is that valid? I was assuming not, so asking if that will be fixed too?

Is it possible to specify that with YAML? I think both are valid responses, so they need to be included, they just shouldn't appear together? I think that needs to be specified at the documentation level, not the spec level doesn't it?

I modeled this response on the /rearrangement response, so unless I made a mistake with the copy I think the /rearrangement and /repertoire endpoints have the same problem, no? I think the issue is probably addressed in the documentation for them - but we don't have any Clone documentation yet 8-)

schristley:
> Is it possible to specify that with YAML?

It seems so, TACC pointed out to me that openapi V3 has the `OneOf` [keyword](http://spec.openapis.org/oas/v3.0.3#properties). I don't think it was available in V2 though.

schristley:
@bcorrie This looks fine to me, want to merge it?

PR #427, opened 2020-07-01
initial pass at openapi v3 conversion

schristley:


schristley:
I tested the AIRR schema and ADC API syntax by actually loading it up with [OpenAPI V3 middleware](https://bitbucket.org/vdjserver/api-js-tapis/src/openapi-v2-to-v3/) and verified that the API service could respond to requests.

schristley:
As these are separate files, they should not interfere with any of the existing Openapi V2 code.

bcorrie (approved):
@schristley as you say, no impact on current version... Loaded it into swagger editor and it worked with no errors so seems consistent...

PR #428, opened 2020-07-01
Updates to the Rearrangement support table

### LINKED ISSUE #425: Update Rearrangement support table
The list of tools supporting the Rearrangement schema is out of date.  It needs to be updated with:

- [x] IMGT/HighV-QUEST
- [x] alakazam, shazam, tigger, scoper
- [x] scirpy (See #411)
- [x] sumrep
- [x] rahbit
- [x] immunarch
- [x] immuneSIM

We need to check on the status of the following and update the table accordingly:

- [x] MiXCR
- [ ] partis
- [ ] IGoR, OLGA
- [ ] Vidjil Web Platform
- [x] cellranger

javh:
Draft of updates to the Rearrangement support table with current application status. Resolves #425.

This isn't done yet. I'm just making the PR for discussion and to build a docs branch for review.  I still have to contact the authors for approval and status. Including figuring out what is really TBD and what won't be updated with Rearrangement support.

(The static link to the tsv file is temporary. I will need to put it back to the master branch url when finished.)


javh:
https://docs.airr-community.org/en/support-update/resources/rearrangement_support.html

javh:
Heard back from @GreiffLab. The immuneSIM addition is good.

javh:
Heard back from the MiXCR team. Rearrangement support will be in v3.0.14 which should be out in a couple weeks. I've updated the version accordingly.

bcorrie:
@javh one of the things I was thinking might be useful is to link to the tools documentation for input and output for the AIRR formats. I have done this for iReceptor. This is a bit more work on our part, but I think would both be very useful to the community (as a user can easily find how to use the AIRR format with each tool) and give a good feel for completeness of the list... No links from a tool implies undocumented support only...

scharch:
@bcorrie I like the idea, but what would it look like for a tool like SONAR which essentially _only_ uses the AIRR Rearrangements format? (Fastas will be deprecated later this summer and I don't support any other tabular formats for rearrangements.) Would you just link to the main README?

javh:
Potential issues with that:

1. The issue @scharch raises, which is also true for all of the Immcantation tools now. There's a description of the data standard, but there's no explicit input/output documentation because it's just how the tools work. That's also going to apply to a few other tools aside from SONAR and Immcantation. Paradoxically, leaving the tools with the most integrated support without proper links in that column.
2. Maintenance could be a hassle. Project and publication links should be stable, but internal documentation urls may not be.
3. Some tools may not have any documentation to link to regarding the standard.

In the case of (3), we could just not provide links, but it might look weird.

That said, I don't actually have a strong opinion on it. Depends on how many tools actually have a doc page that would make sense. If there isn't a good doc link for most of the tools, then I think it's more effort than it's worth.

bcorrie:
> Potential issues with that:

Not too tied to the idea - was a thought... For me it was about making it as easy as possible for someone to figure out how to produce AIRR files. You go the AIRR site, see a list of software, say "cool, I didn't know MiXCR could output AIRR format, how do I do that?" - click, done... 8-)

> 1. The issue @scharch raises, which is also true for all of the Immcantation tools now. There's a description of the data standard, but there's no explicit input/output documentation because it's just how the tools work. That's also going to apply to a few other tools aside from SONAR and Immcantation. Paradoxically, leaving the tools with the most integrated support without proper links in that column.

Presumable somewhere in the docs it says the data format used is the AIRR TSV format? I think that is all that would be required... 

> 2. Maintenance could be a hassle. Project and publication links should be stable, but internal documentation urls may not be.

Agreed. this is the big hurdle for me.

> 3. Some tools may not have any documentation to link to regarding the standard.

> In the case of (3), we could just not provide links, but it might look weird.

I think that is partially my point as to why it is a good idea, actually... If it isn't clear that the tool uses the AIRR format (through some sort of documentation) can you claim that the tool supports the standard? 8-)

For example both the following have strong statements in this regard.

https://github.com/scharch/SONAR
https://changeo.readthedocs.io/en/stable/standard.html

> 
> That said, I don't actually have a strong opinion on it. I could go either way. Depends on how many tools actually have a doc page that would make sense. If there isn't a good doc link for most of the tools, then I think it's more effort than it's worth. If there's an abundance of good doc links, then it seems valuable.

As I say, I am not really tied to it, but I think it would be helpful. I agree it would be a bit of work... In fact, I notice that our documentation is actually out of date as we now support both AIRR Repertoire JSON and AIRR TSV files (the iReceptor docs only talk about AIRR TSV).

bcorrie:
> In fact, I notice that our documentation is actually out of date as we now support both AIRR Repertoire JSON and AIRR TSV files (the iReceptor docs only talk about AIRR TSV).

From a overhead perspective, the above issue wasn't that bad... We just made a major release and it changed the documentation. So really the page should have been updated with the major release, which would update the docs for that release which include docs on loading both AIRR formats.

So maybe the process is a new release comes out, hopefully we notice, and then we update? Won't happen too often and therefore maybe not too much overhead?

javh:
Well, ideally, the list of tool will grows long enough that it becomes prohibitively obnoxious to keep track of...

bcorrie:
OK, I'm out at obnoxious 8-)

scharch:
@javh should the AIRR Python and R libraries be bumped to 1.3?
I guess in general the idea just to list the original compliant version, not the most recent, but those seems like special cases?

javh:
@scharch, yeah, I'd prefer to list minimal versions rather than current versions.

Also, the table is specifically for compliance with v1.2 of the schema.  That may not be strictly true though. I think scirpy, IMGT, and cellranger are actually on 1.3, but I haven't looked through the details.  Something we should probably do...

docs/resources/tables/rearrangement_support.tsv
```diff
@@ -1,18 +1,29 @@
...
-`OLGA <https://github.com/zsethna/OLGA>`_	TBD	Input and output	"`Sethna et al. Bioinformatics, 2019. <https://doi.org/10.1093/bioinformatics/btz035>`_"
-`Partis <https://github.com/psathyrella/partis>`_	TBD	Output	"`Ralph & Matsen. PLoS Comput Biol, 2016. <https://doi.org/10.1371/journal.pcbi.1004409>`_"
+`immuneSIM <https://immuneSIM.readthedocs.io>`_	0.8.7	Output	"`Weber et al. Bioinformatics, 2020. <https://doi.org/10.1093/bioinformatics/btaa158>`_"
+`iReceptor <http://www.ireceptor.org>`_	3.0	`Input <https://github.com/sfu-ireceptor/dataloading-mongo/tree/turnkey-v3/dataload>`_ and `output <http://www.ireceptor.org/node/97>`_	"`Corrie et al. Immunol Rev, 2018. <https://doi.org/10.1111/imr.12666>`_"
+`MiXCR <https://milaboratory.com/software/mixcr>`_	3.0.14	Output	"`Bolotin et al. Nat Methods, 2015. <https://doi.org/10.1038/nmeth.3364>`_"
+`RAbHIT <https://yaarilab.bitbucket.io/RAbHIT>`_	0.1.5	Input and output	"`Gidoni et al. Nat Commun, 2019. <https://doi.org/10.1038/s41467-019-08489-3>`_"
+`scirpy <https://icbi-lab.github.io/scirpy>`_	0.3	Input	"`Sturm et al. biRxiv, 2020. <https://doi.org/10.1101/2020.04.10.035865>`_"
```
grst, line 17:
```suggestion
`scirpy <https://icbi-lab.github.io/scirpy>`_	0.3	Input	"`Sturm et al. Bioinformatics, 2020. <https://doi.org/10.1093/bioinformatics/btaa611>`_"

```

javh:
I haven't contacted everyone yet, but I'm slow and I'd like to merge this in (after swapping the url for the table). We can update the missing details later.

javh:
Kind of, yeah. It's really just because @bcorrie added it awhile ago.

bcorrie:
> Why is iReceptor unique in getting links in the input/output column? Is it a vestige of [this](https://github.com/airr-community/airr-standards/pull/428#issuecomment-656747179) proposal?

Yep... My suggestion was that we link to the documentation for producing or reading AIRR files so that it was both easy to find out how to work with AIRR files directly from the AIRR site for each tool as well as demonstrate that the tool really did support AIRR files (in the sense that if it isn't documented how to read/write AIRR files, my claim was that the tool didn't really support AIRR files).

I think where we landed was that this was to hard to maintain. Would you prefer we go with no links to be consistent?

scharch:
> Would you prefer we go with no links to be consistent?

I think so, yeah. But it's not a deal breaker.

bcorrie:
> > Would you prefer we go with no links to be consistent?
> 
> I think so, yeah. But it's not a deal breaker.

I'm easy either way, although it seems like a shame to take away direct links if we have them. I suggested above that I am willing to try to create the links for the list we have now, but I agree that having a mess of links with no links without some rigor may be more confusing than helpful. So if we aren't going to try and maintain it (and that I don't think I want to commit to 8-) then we can remove the iReceptor links.

PR #438, opened 2020-07-10
Add draft fields for germline inclusion.

### LINKED ISSUE #416: Integrating Germline and MHC into AIRR data model
Trying to capture an email discussion that grew out of the Common Repository Working Group and has been on the back burner with all of us for some time... How do we annotate AIRR-seq data sets with Germline and MHC information?

This is somewhat related to #157 and #258 but I am starting a new thread/conversation specifically for this...

bcorrie:
As per the discussion - fixes #416

Thoughts/comments?


scharch:
I didn't completely follow this in the previous discussion --the idea here is to link to a database somewhere else, instead of encoding the germline sequences directly in the metadata?

bcorrie:
> I didn't completely follow this in the previous discussion --the idea here is to link to a database somewhere else, instead of encoding the germline sequences directly in the metadata?

Actually, exactly what is stored in these fields is yet to be decided - this is more to establish where in the metadata the fields would be.

I think the general agreement is that we can't wait for an external database nor can we wait for an agreed upon nomenclature that is inclusive enough to capture everything - so I think we want to pick something that makes sense in the short term so we have SOMETHING to start with...

The current thought I think is that we would store the entire set in the metadata as per @williamdlees suggestion here: https://github.com/airr-community/airr-standards/issues/416#issuecomment-648691832

bcorrie:
I made the germline field an array of strings, assuming we would use gene names of some form (standard where possible as per how we describe v_call) with the relatively easy transformation to using CURIEs once we have some sort  of look up service for such things.

Comments/criticisms welcome, just working through some ideas...

bcorrie:
Wondering if we need a germline_annotation to complement germline_database in DataProcessing.

germline_database allows us to say IMGT downloaded July 22, 2020.
germline_annotation could be a germline array that actually has the germline genes listed as used in the annotation process in the case where a standard database is not used.

bussec:
I think we shouldn't include MHC annotation as simple part of a `germline` array. This does not mean that it should not be in `Subject` (as we already discussed), but it should be separated from Ig/TCR. Or all three should be separated to simplify parsing...

Regarding `germline_annotation`: Does "listed as used" mean that you want to include the sequence information there?

bcorrie:
> Regarding `germline_annotation`: Does "listed as used" mean that you want to include the sequence information there?

No, listed as used from my perspective would be the list of germline gene descriptors (whatever that is) rather than the sequences.

bcorrie:
> I think we shouldn't include MHC annotation as simple part of a `germline` array. This does not mean that it should not be in `Subject` (as we already discussed), but it should be separated from Ig/TCR. Or all three should be separated to simplify parsing...

Are you suggesting we have germline_ig, germline_tcr, and germline_mhc? 

The problem with this is what other germline gene types might you have? You would then need a new field? So does IG/TCR/MHC capture 99.9% of what we want. If so, then this is OK. If we are going to be adding another field in 8 weeks because we want to track something else in the germline, then maybe there is another mechanism we can use. For example I could see an array of germline objects with a controlled vocabulary of "ig", "tcr", "mhc" with each containing an array of germline genes for that class of germline gene?

bussec:
@bcorrie 

1. `germline_annotation`: Ok, so this is basically a way to subset the data that is/was present in `germline_database`, correct?

2. Ig/TCR/MHC: IMO capturing these three will cover >99% of the current cases. I am not sure about the digits after the decimal point as you could think about e.g., variants in the MHC presentation machinery or minor histocompatibility antigens or neoantigens in tumors etc. But to my understanding your are looking for something that is good and reasonably comprehensive to start with, not the final version, and I think this fits the bill.

3. Structure: The array approach seems more flexible to me. It would also allow to address another issue which is that there is -- BTMK -- no single repository that would provide Ig/TCR and MHC information at the desired depth and quality. Therefore we would need to support multiple `germline_databases` or assign that information to the individual gene locus.

4. We should check whether "germline" is a common term to use for non-Ig/TCR stuff (as germline sequence = somatic sequence).

bcorrie:
> * `germline_annotation`: Ok, so this is basically a way to subset the data that is/was present in `germline_database`, correct?
> *

Yes, I think so. `germline_database` is a well know, reference database as per the AIRR Spec example "ENSEMBL, Homo sapiens build 90, 2017-10-01"

`germline_annotation` provides the user with a mechanism to say in the annotation process I used a subset of the germline genes that is is in `germline_database`, presumably because I know that this individual has a specific set of germline genes.

bcorrie:
> * Ig/TCR/MHC: IMO capturing these three will cover >99% of the current cases. I am not sure about the digits after the decimal point as you could think about e.g., variants in the MHC presentation machinery or minor histocompatibility antigens or neoantigens in tumors etc. But to my understanding your are looking for something that is good and reasonably comprehensive to start with, not the final version, and I think this fits the bill.

Yep! Lets get something we can use and then refine it if need be.

> * Structure: The array approach seems more flexible to me. It would also allow to address another issue which is that there is -- BTMK -- no single repository that would provide Ig/TCR and MHC information at the desired depth and quality. Therefore we would need to support multiple `germline_databases` or assign that information to the individual gene locus.

I agree - arrays seem better, I will reconstruct the spec in that form.

I am not suggesting that `germline_database` change at all - which is maybe bad. This specifies which `germline_database` was used for sequence annotation as part of DataProcessing.  `germline_annotation` allows you to say, hey I used that `germline_database` for annotation but I only used this subset of germline genes for the annotation because I have some prior knowledge about this subject... 

So I think we only need one germline_database, no? Are you suggesting we need a germline_annotation_ig and germline_annotation_tcr (or an array) at the DataProcessing level.

bcorrie:
> > * Structure: The array approach seems more flexible to me. It would also allow to address another issue which is that there is -- BTMK -- no single repository that would provide Ig/TCR and MHC information at the desired depth and quality. Therefore we would need to support multiple `germline_databases` or assign that information to the individual gene locus.
> 
> I agree - arrays seem better, I will reconstruct the spec in that form.

I have added the germline array structure. The `germline` object looks like this:

```
        "germline": [
          {
            "germline_class": "ig",
            "germline_genes": [
              "IGHV1-1*01",
              "IGHV2-2*02"
            ],
            "germline_process": "repertoire_sequencing"
          }
        ]
      },
```
With germline_class being either ig, tcr, or mhc.

Feedback @bussec @williamdlees 

I haven't added the array structure to the other germline fields yet.

bcorrie:
Question: if the array above makes sense, do we need an array for germline_inferred so we can have inferred ig and tcr germline in DataProcessing? 

Similarly, if we add a germline_annotation to contain the germline used for annotation we would need the array structure there as well?

williamdlees:
The structure looks good, Brian. Just one small point - 'germline_alleles' might be preferable to 'germline_genes'?

williamdlees:
Sorry, one further thing. As it's common to have reads from a small subset of chains, it might be helpful to specify the locus (IGH, IGK, IGL, TRA, TRB, TRD, TRG) rather than the class. And I agree that we should keep MHC separate - and perhaps there is already work on its specification that we could borrow from?

bcorrie:
> Sorry, one further thing. As it's common to have reads from a small subset of chains, it might be helpful to specify the locus (IGH, IGK, IGL, TRA, TRB, TRD, TRG) rather than the class. And I agree that we should keep MHC separate - and perhaps there is already work on its specification that we could borrow from?

So we would have a controlled vocabulary of IGH, IGK, IGL, TRA, TRB, TRD, TRG and MHC and then have an array with separate germline lists for each. Typically the germline data at the DataProcessing level might have only a couple of these (e.g. because the data is TRB data only) while the Subject level germline data might have anywhere to one or all of these?

williamdlees:
> 
> 
> > Sorry, one further thing. As it's common to have reads from a small subset of chains, it might be helpful to specify the locus (IGH, IGK, IGL, TRA, TRB, TRD, TRG) rather than the class. And I agree that we should keep MHC separate - and perhaps there is already work on its specification that we could borrow from?
> 
> So we would have a controlled vocabulary of IGH, IGK, IGL, TRA, TRB, TRD, TRG and MHC and then have an array with separate germline lists for each. Typically the germline data at the DataProcessing level might have only a couple of these (e.g. because the data is TRB data only) while the Subject level germline data might have anywhere to one or all of these?

Yes, but I would suggest having MHC as a completely different branch, which we do not define. Perhaps

 "receptor_germline": [
          "germline_process": "repertoire_sequencing",
          {
            "locus": "igh",
            "germline_genes": [
              "IGHV1-1*01",
              "IGHV2-2*02"
            ],
          }
        ],
     "mhc_germline": [
    ...to be defined...
        ]

bussec:
@bcorrie I agree with all of @williamdlees's points:

* We should use the same CV like for the other places where we deal with Ig/TCR loci
* `*_alleles` is probably a more accurate term than `*_genes`
* Having MHC in a separate branch also makes sense to me, e.g., as inferrence procedures might differ and we don't really have a good understanding of them. Whether this section should be called `mhc_germline` is something we should discuss at some point. but if relabeling the keys is not a major issue, I would opt for going ahead with this structure now.

bcorrie:
> * We should use the same CV like for the other places where we deal with Ig/TCR loci

Just so we don't go around in circles, can you tell me what controlled vocabulary to use here?

Also is "receptor_germline" the right term - is this not too general? Isn't this the "immune_receptor_germline"

bcorrie:
Made changes as per @williamdlees suggested, format is now like this:

```
        "germline": {
          "receptor_germline": [
            {
              "germline_class": "IGH",
              "germline_alleles": [
                "IGHV1-1*01",
                "IGHV2-2*02"
              ],
              "germline_process": "repertoire_sequencing"
            }
          ],
          "mhc_germline": [
            {
              "germline_class": "MHC",
              "germline_alleles": [
                "HLA‐C*07:29",
                "HLA-B*15:27"
              ],
              "germline_process": "repertoire_sequencing"
            }
          ]
        }
      }
```
Does that work?

bcorrie:
Met with @watsoc04 today to discuss, we think we are on the right track 8-)

Changed the data_processing germline field to be - surprise - `data_processing_germline` and to have the same structure as the `receptor_germline` field. 

@williamdlees and @bussec any comments?

bcorrie:
Note there is no longer an inferred_germline associated with a data_processing, as I think we agreed that if you are inferring a germline for a subject from a repertoire, that the proper place to store the inferred germline is in `subject.germline`

williamdlees:
I'm sorry, Brian, I've got lost. Where can I view the schema as it stands now, on this branch?

bcorrie:
> I'm sorry, Brian, I've got lost. Where can I view the schema as it stands now, on this branch?

Yes, there are two fields, a `germline` field in `Subject` and a `data_processing_germline` field in `DataProcessing`. Structure is similar except the first one in `Subject` has an MHC component while the one in `DataProcessing` does not.

bcorrie:
https://github.com/airr-community/airr-standards/blob/germline-fields/specs/airr-schema.yaml

williamdlees:
Thanks! Looks good to me.

williamdlees (approved):

schristley (commented):
The fields look good to me, but it looks like these array of arrays can get pretty large? I mean we are talking hundreds of genes. I'm assuming that with data processing, the germline would be the same for many repertoires in a study, so it can be a lot of duplicated data. What about having a top-level `Germline` object with a `germline_id`, like we are doing with `Clone` and others, and then just the `germline_id` needs to be stored in the subject and data processing?

schristley:
It would be good to merge the existing work on germline set schema #157

bcorrie:
> The fields look good to me, but it looks like these array of arrays can get pretty large? I mean we are talking hundreds of genes. I'm assuming that with data processing, the germline would be the same for many repertoires in a study, so it can be a lot of duplicated data. What about having a top-level `Germline` object with a `germline_id`, like we are doing with `Clone` and others, and then just the `germline_id` needs to be stored in the subject and data processing?

The main reason we didn't want to do something like that is that we thought that this is likely to change once there are repositories that store germline reference and possibly reference sets (OGRDB, IMGT) and we can refer to such things. In addition, DataProcessing as an object is in the midst of the <repertoire_id, data_processing_id, sample_processing_id> and Repertoire object discussion. For example, the germline replication would be resolved if DataProcessing was an entity in and of itself and not replicated as a member of a Repertoire.

Rather than come up with a solution that solves this more complex issue, which we all agreed would take some time, we wanted to get something in place that we can move forward with.

bcorrie:
> It would be good to merge the existing work on germline set schema #157

The plan is to definitely do this, but again we didn't want to wait for this to be resolved. As you can see in the early discussion for this issue we tried to push that agenda (https://github.com/airr-community/airr-standards/issues/416#issuecomment-639611583) but we came to the conclusion that having a place to store germline information (what this pull request is about) and the specific mechanics and standards for storing the information in the field (more along the lines of #157) are related but independent. So we want to have a place to store the data, and then we can evolve how we store the concise definitions of the genes once some of the intricacies of #157 are sorted out.

schristley:
I feel we are talking about different things, I'll try to be more explicit...

1. put fields in a top-level `GermlineSet` schema object.

In `airr-schema.yaml`, we have top-level objects like `Study`, `Subject`, etc. I'm asking that the fields be put into such a `GermlineSet` schema object. Then, at least, use `$ref` to refer to it. However, that leads to my second point.

2. Use an `id` instead of denormalizing the data with `$ref`.

Take the thousands of repertoires in the ADC, and multiply it by hundreds of genes, we are talking hundreds of thousands of array entries. This seems rather excessive, and is that how we really want to promote analysis tools to access this data? Especially as we assume in the future that we will store a reference to the data elsewhere. Why not just put that design in now as the draft? It's not much different with how we've been thinking about the other new draft objects.

bcorrie:
> I feel we are talking about different things, I'll try to be more explicit...
> 
> 1. put fields in a top-level `GermlineSet` schema object.
> 
> In `airr-schema.yaml`, we have top-level objects like `Study`, `Subject`, etc. I'm asking that the fields be put into such a `GermlineSet` schema object. Then, at least, use `$ref` to refer to it. However, that leads to my second point.

That is easy, and doesn't change the structure, so can do that no problem.

> 1. Use an `id` instead of denormalizing the data with `$ref`.
> 
> Take the thousands of repertoires in the ADC, and multiply it by hundreds of genes, we are talking hundreds of thousands of array entries. This seems rather excessive, and is that how we really want to promote analysis tools to access this data? Especially as we assume in the future that we will store a reference to the data elsewhere. Why not just put that design in now as the draft? It's not much different with how we've been thinking about the other new draft objects.

The problem here is not necessarily the fact that it is necessary to replicate `GermlineSet` but instead it is currently necessary to replicate the `DataProcessing` which contains the `GermlineSet`. It is really the `DataProcessing` that we are perhaps needlessly replicating here, which in turn means the `GermlineSet` is needlessly replicated. If I have one `DataProcessing` for my study then I have one `GermlineSet` for my study. If I have one `GermlineSet` for each of N subjects in my study, then I have N `DataProcessings` and therefore N `GermlineSets`.

There is already a data_processing_id that we can use do avoid the duplication of DataProcessing (and therfore `GermlineSet`), and this will hopefully fall out of the redesign of `DataProcessing`. Given that the `GermlineSet` is a part of a `DataProcessing` I don't think we need a separate AIRR spec identifier for this object. In fact, I would be more worried about the proliferation of objects that have unique IDs, which in turn suggests complex relationships between objects that people have a very hard time agreeing on 8-)

I do agree we might indeed want to point to an external identifier at some point in the future, but I don't know that that is going to be pointing at an external `GermlineSet`. That means we need a repository (or set of repositories) that stores `GemlineSets` on behalf of the community and everyone would need to use this if they work working with `GermlineSets`. I am not convinced that is likely in the near future (or even far future). One of the suggestions we had earlier was that we use a CURIE for each Gene (which handles the reference to an external definition for that gene) but still maintain a list of the genes as CURIEs in the germline. The CURIEs would be stored in the database (giving us rigour for genes) but our repositories would still be holding a list of germline genes. I think that is the more likely solution that will emerge in the next year or so...

schristley:
> The problem here is not necessarily the fact that it is necessary to replicate GermlineSet.

I haven't really heard an argument why GermlineSet ***must*** be denormalized in `Subject` and `DataProcessing`. Is having an `id` field, where the contents, syntax and uniqueness level are TBD, so controversial?

I haven't been involved in the discussions around use cases. Are you expecting queries through the ADC API on this info? Are you expecting analysis tools to utilize this info? With other denormalized objects in `Repertoire`, we have an `id` so that client/tools can re-normalize if they want. When looking at multiple repertoires, how does a client/tool know if the same germline set was used across them?

bcorrie:
> I haven't really heard an argument why GermlineSet _**must**_ be denormalized in `Subject` and `DataProcessing`. Is having an `id` field, where the contents, syntax and uniqueness level are TBD, so controversial?

Are you suggesting this:

```
        "germline": {
          "receptor_germline": [
            {
              "germline_set_id": "GLS_ID1"
              "germline_class": "IGH",
              "germline_alleles": [
                "IGHV1-1*01",
                "IGHV2-2*02"
              ],
              "germline_process": "repertoire_sequencing"
            }
          ],
```
or this:

```
        "germline": {
          "receptor_germline": [
            {
              "germline_set_id": "GLS_ID1"
            }
          ],
```

schristley:
> or this:

I would prefer the second choice. Though my concept of `GermlineSet` might not be matching to yours. I was thinking something more like this:

```
data_processing:
  germline_set_id:
    type: string
```

or if denormalization is required then

```
data_processing:
  germline:
    $ref: '#/GermlineSet'
```

in either case, there is

```
GermlineSet:
  type: object
  properties:
    germline_set_id:
      type: string
    [... all the other stuff ...]
```

williamdlees:
As I understand it, we are describing in the schema an external representation rather than an internal implementation. Externally, I think it's important that a document contains all the relevant information and does not refer out to ids of items that are not included. This is not only because of the inconvenience that can cause in interpreting the document (both when reading it and computationally), but also because references can be mistranscribed and are often not guaranteed to be immutable.

Perhaps the two issues being raised here are (1) reluctance to repeat a definition within the schema and (2) reluctance to repeat data within the document. For (1), I don't know the schema language you are using at all well, but perhaps there is a way of saying that the same element is repeated. Is this what $ref is for?  For (2), this reminds me of forms that ask for a billing address and a shipping address. Often there is a box to check to say that they are the same. We could adopt the same approach here, by having a flag to say that the germline set is the same in both cases and is not repeated.

schristley:
> Perhaps the two issues being raised here are (1) reluctance to repeat a definition within the schema and (2) reluctance to repeat data within the document. For (1), I don't know the schema language you are using at all well, but perhaps there is a way of saying that the same element is repeated. Is this what $ref is for?

Yes, exactly, $ref is a simple way to have a single schema definition that's then referenced everywhere it's used. That's why I suggested using it.

> For (2), this reminds me of forms that ask for a billing address and a shipping address. Often there is a box to check to say that they are the same. We could adopt the same approach here, by having a flag to say that the germline set is the same in both cases and is not repeated.

Yep, that is why I suggested `germline_set_id`, instead of repeating all the data, just provide the same id value, that then acts as a flag that it's the same data.

williamdlees:
OK, so, for the time being at least, germline_set_id will be an arbitrary value that must be defined by a germline set somewhere in the file?

schristley:
> OK, so, for the time being at least, germline_set_id will be an arbitrary value that must be defined uniquely by a germline set somewhere in the file?

That's right. It's like most of our other ids in that they only need to be internally consistent within the AIRR data model.

williamdlees:
Thanks. SOunds good to me.

bcorrie:
@schristley in your use of $ref above, is it possible to create this definition using a YAML/JSON spec. Each of the `GermlineSet` objects has a different controlled vocabulary (`germline_class`), and these are part of the spec. I am not sure this is possible is it? 

> GermlineSet:
>   type: object
>   properties:
>     germline_set_id:
>       type: string
>     [... all the other stuff ...]

bcorrie:
Or at least the only way to do it is have the controlled vocabulary the union of the possible controlled vocabularies of the different possible places where GermlineSet might be used... Which we tried to avoid by splitting up MHC and Immune Receptor germline objects.

bcorrie:
I have added a `germline_set_id` field so it is possible to disambiguate whether a germline set is the same or not across different Subjects or DataProcessing objects (more important across DataProcessing I assume). Given that the controlled vocabularies are different, I am not sure we can use the GermlineSet object for this so it may need to be removed.

schristley:
Added `GeneDescription` schema based on earlier work. I'm not convinced that the intermediate germline class is needed, as each gene description has a `locus` which can be used to group genes like the germline class is doing. Also, `germline_process` seems like it should be with each individual gene description versus at the germline class level.

bcorrie:
@schristley are you sure that is valid YAML? The spec is included in the api, and the API is now breaking in the swagger editor.

bcorrie:
@schristley we had the germline denormalized so that the germline was stored with the repertoire. The germline was stored in a very simple form as an array of gene calls similar to how we store gene calls at the rearrangement level. This is no longer the case. 

If the Repertoire germline is normalized, we need a mechanism to store and look up germline sets. Base on our earlier discussion, we had decided to defer the definition of the germline set, how to store it, and how to look it up, where to look it up, etc to a later discussion and instead just store a simple representation based on how we store gene calls at the rearrangement level. The primary reason for deferring that was because agreeing on a community approved representation for germline genes, where they are stored, how they are stored, and how they are looked up was considered a longer term discussion. The Germline Working Group has been working on this for some time, and we had agreed that a simple, denormalized representation was a good intermediate step forward...

In this way, the gene names in rearrangement v_call/d_call/j_call take on the same form as the gene names in the germline gene set. This makes the most sense as an intermediate step.

The suggestion was that we tackle GeneDescription as a separate issue, and then change how genes are described throughout the spec at the same time. The key thing is we did not want to wait for an agreed GeneDescription to add Germline to the spec.

schristley:
> store a simple representation based on how we store gene calls at the rearrangement level.

@bcorrie Is this a theoretical representation then or is it meant to implemented, that is, what is the timeline for researchers to specify this with their study as part of MiAIRR? To have repositories in the ADC store and support queries on this info? To have analysis tools access this info as part of the analysis process?

bcorrie:
> > store a simple representation based on how we store gene calls at the rearrangement level.
> 
> @bcorrie Is this a theoretical representation then or is it meant to implemented, that is, what is the timeline for researchers to specify this with their study as part of MiAIRR? To have repositories in the ADC store and support queries on this info? To have analysis tools access this info as part of the analysis process?

Definitely meant to be implemented. We want to move forward with something that can be used today. We have a mechanism for storing gene calls in a `Rearrangement`, at a minimum, we want to use the same mechanism to enable the ability to either:

- store the germline genes used in the annotation process (in `DataProcessing`)
- store any sequenced or inferred germline genes at the subject level (in `Subject`)

I think if we have both germline_set_id and a list of germline genes as in the spec now, we have something that works and in a way future proofs us when a more rigorous definition for the gene names is agreed and/or a more complete mechanism for storing germline sets appears on the horizon. 

The goal was to have this as part of v1.4 so that if researchers wanted to take advantage of this capability and/or repository and tool developers wanted to store and process it, they could.

schristley:
> Definitely meant to be implemented. We want to move forward with something that can be used today.

Ok, that's what I thought but wanted to be sure, which is why I disagree about the denormalization design, as it was put forward.

> * store the germline genes used in the annotation process (in DataProcessing)
> * store any sequenced or inferred germline genes at the subject level (in Subject)

> I think if we have both germline_set_id and a list of germline genes as in the spec now, we have something that works and in a way future proofs us when a more rigorous definition for the gene names is agreed and/or a more complete mechanism for storing germline sets appears on the horizon.

I completely agree with this conceptually. My primary sticking point is storing that data in a denormalized form within the Repertoire. It is too big with too much duplication. We are talking about inserting arrays with 100s of genes into every subject and every repertoire for every study, introducing 10^6 entries with large duplication across the ADC.

> If the Repertoire germline is normalized, we need a mechanism to store and look up germline sets. Base on our earlier discussion, we had decided to defer the definition of the germline set, how to store it, and how to look it up, where to look it up, etc to a later discussion and instead just store a simple representation based on how we store gene calls at the rearrangement level.

Except you are contradicting yourself here, you say we are deferring how to store/lookup/etc a germline set, then you propose exactly that, a place to store and lookup the germline set, and furthermore to actual implement it.

I proposed an alternative that keeps your basic schema, and avoids the duplication. I see no reason why a simple JSON file format cannot be devised for v1.4 to hold the information.

I'm fine with setting aside `GeneDescription` if it's problematic, that's a change I can do quickly.

bcorrie:
> > Definitely meant to be implemented. We want to move forward with something that can be used today.
> 
> Ok, that's what I thought but wanted to be sure, which is why I disagree about the denormalization design, as it was put forward.
> 
> > * store the germline genes used in the annotation process (in DataProcessing)
> > * store any sequenced or inferred germline genes at the subject level (in Subject)
> 
> > I think if we have both germline_set_id and a list of germline genes as in the spec now, we have something that works and in a way future proofs us when a more rigorous definition for the gene names is agreed and/or a more complete mechanism for storing germline sets appears on the horizon.
> 
> I completely agree with this conceptually. My primary sticking point is storing that data in a denormalized form within the Repertoire. It is too big with too much duplication. We are talking about inserting arrays with 100s of genes into every subject and every repertoire for every study, introducing 10^6 entries with large duplication across the ADC.

Worst case scenario:

200 gene names * 15 bytes per name = 3000 bytes per repertoire
If you have a Subject and Data Processing germline then you have two per repertoire = 6000 bytes/repertoire
6000 repertoires today = 36MB

You can increase the number of gene names or the number of bytes per gene name and it is still not a lot of data considering it is accompanying 4B rearrangements. So is it a lot of duplication, absolutely, but is it a lot of data, not really...

The key thing is this is a worst case scenario, and assumes that all subjects, and with full duplication therefore all repertoires, have an inferred germline. It also assumes that all repertoires are annotated with a custom germline rather than just using germline_db to capture "IMGT version X". So worst case today we have 12,000 germline sets stored.

In reality, how many studies/subjects will truly require a germline? You don't have to have a subject germline, nor do you have to have a data processing germline (we still have germline_db). So nothing needs to be added to the ADC and our current repositories. 

What it does do is give us a capability such that if someone does use a custom germline for annotation or actually does infer the germline for a subject (which in my understanding is pretty important) they could actually do it. 

So the actual number of such germline sets, at least in the foreseeable future, would be tiny.

In addition, with the addition of germline_set_id, it is possible to annotate the data to capture where the germline_set is identical across multiple repertoires. So you don't lose any capability (other than storage).

Last but not least, from a repository perspective, if you want to normalize your data for germline sets and store them internally in a collection with primary keys that link them to germline_set_id you can do that. So the repository can easily avoid the internal duplication if they want to... Similarly you could store the data in a separate JSON file and not provide the duplicated Germline if you wanted. All are possible...

> > If the Repertoire germline is normalized, we need a mechanism to store and look up germline sets. Base on our earlier discussion, we had decided to defer the definition of the germline set, how to store it, and how to look it up, where to look it up, etc to a later discussion and instead just store a simple representation based on how we store gene calls at the rearrangement level.
> 
> Except you are contradicting yourself here, you say we are deferring how to store/lookup/etc a germline set, then you propose exactly that, a place to store and lookup the germline set, and furthermore to actual implement it.

> I proposed an alternative that keeps your basic schema, and avoids the duplication. I see no reason why a simple JSON file format cannot be devised for v1.4 to hold the information.

In VDJServer, if you has a study with an inferred germline for each subject and the normalized germline set (so you only have germline_set_id as in the current normalized spec) how would you implement a query in VDJServer? For example, what if I want to do a query that gives me all of the subjects that have IGHV1-69*01 in their germline.

If germline is in the repertoire, my query with the spec from https://github.com/airr-community/airr-standards/blob/2c13ca2604495e1168fee97b89784d25611feabb/specs/airr-schema.yaml is something like this:

```
{
    "filters":{
                "op":"contains",
                "content": {
                    "field":"subject.germline.receptor_germline.germline_alleles",
                    "value":"IGHV1-69*01"
                }
	    }
}
```
How do you do this with the normalized spec?

williamdlees:
Surely we should focus on the external representation here. implementors can make their own tradeoffs when storing - what's important is how the data is represented when it's exchanged between systems or people.

bcorrie:
> Surely we should focus on the external representation here. implementors can make their own tradeoffs when storing - what's important is how the data is represented when it's exchanged between systems or people.

I agree (I think 8-). My suggestion is that the spec (which is the external representation and includes the Repertoire JSON file format and the API), at least for now, has both a germline_set_id as well as a list of the alleles for that germline set. That way all options are available to the tools and the repository. If you want to have external GermlineSets (stored outside the Repertoire object) and use just germline_set_id (and no allele lists) you can. But it also doesn't prevent the simple use of an allele list within the Repertoire object if that is what the user/researcher wants. I think we would want to encourage both so you can normalize if you want, but I don't think we should force normalization. Maybe we can agree on normalization once we have thought about how to store and retrieve germline sets, but I think we should defer that to later and not hold up the ability to start storing germline info.

Query's can be complete with both germline_set_id and allele lists populated, and fairly simple:
- Give me all of the repertoires that use germline_set_id == 42
- Give me all of the subjects that have IGHV1-69*01 in their `Subject` allele list
- Give me all of the repertoires that have IGHV1-69*01 in their `DataProcessing` allele list
- Give me all of the germline_set_id that have IGHV1-69*01 in their allele list

But you can do complicated things as well...
- Give me all of the germline_set_ids that have IGHV1-69*01 and IGHV3-53 in their allele lists
- Give me all of the subjects that have those germline_set_ids
- Give me all of the repertoires that have been annotated with those germline_set_ids

felixbreden:
hi

I really agree with Brian's statement here:
Maybe we can agree on normalization once we have thought about how to store and retrieve germline sets, but I think we should defer that to later and not hold up the ability to start storing germline info.

I would just like to add that one of my main scientific motivations to all of this is to make it possible to compare germline gene repertoires with naive and the antigen stimulated expressed repertoires and look for associations with disease etc.  so I really hope we can move ahead with a draft way to store germline information, for those that really want to get started with these comparisons???? please? thank you!!

williamdlees:
+1

bcorrie:
OK, I have made a bunch of updates to go back closer to the previous definition here: https://github.com/airr-community/airr-standards/pull/438/commits/2c13ca2604495e1168fee97b89784d25611feabb

We still have all of the IDs, but we also have the lists as part of the metadata. It is NOT normalized - assuming we can wrestle with and figure out if and how we do that later. Because we have the IDs, the implementer can use a normalized or non normalized representation internally, but the information exchange format (the files and the API) contain both the IDs and the data.

I also went back to "just" having an array of strings for allele names rather than using the GeneDescription object as per the discussion here (https://github.com/airr-community/airr-standards/pull/438#issuecomment-705658535). I used the same wording in the description as we use for v_call. That way, if one has germline data and a bunch of rearrangements, the naming of the germline alleles and the rearrangement calls should use the same nomenclature... 

So I would suggest we go forward with this as a first implementation, with the tasks still remaining being:

- Finalize GeneDescription
- Finalize how we will normalize germline gene sets
- Discuss if we want/need repositories that store germline gene sets
- Come up with an API to query germline gene sets

Let me know what people think. I am hoping that with this we can move forward with this in the short term and solve the more complex issues later.

williamdlees (approved):

schristley (approved):

bussec (changes requested):
I am ok with the overall structure, but there are couple of points are unclear to me:

- Where do the properity names in `GeneDescription` come from? I could not find them in #157 and some of them are problematic and might be require change in the future (`sequence_name` (see below), `domain`).
- What exactly should be stored in `GeneDescription`.`sequence_name`? The gene name, the gene symbol or the sequence identifier?
- For `GermlineClassSet`.`germline_alleles`: I assume that the individual elements of this array can either be 
  - a `description_id`, which I assume to be a unique ID that -- by design -- can never be a valid gene symbol at the same time; or
  - a valid gene symbol that can be looked up in `germline_database`
  Is this correct?
- I assume we do not want people to use `GeneDescription` in connection with `MHCGermlineClassSet`.`germline_alleles`, correct? In this case this should be clarified in the description of the latter, i.e. that individual elements must be a valid gene symbol.

bcorrie:
> I am ok with the overall structure, but there are couple of points are unclear to me:
> 
> * Where do the properity names in `GeneDescription` come from? I could not find them in #157 and some of them are problematic and might be require change in the future (`sequence_name` (see below), `domain`).
> * What exactly should be stored in `GeneDescription`.`sequence_name`? The gene name, the gene symbol or the sequence identifier?

The finalization of GeneDescription is going to be deferred until a later issue - and isn't part of this pull request in the sense that GeneDescription is not used in the Subject or DataProcessing objects. I have left it in the spec as a placeholder for discussion.

Would you prefer the definition of this object be removed from the pull request to avoid confusion and added in a different pull request that would discuss the above issues?

> * For `GermlineClassSet`.`germline_alleles`: I assume that the individual elements of this array can either be
>   
>   * a `description_id`, which I assume to be a unique ID that -- by design -- can never be a valid gene symbol at the same time; or
>   * a valid gene symbol that can be looked up in `germline_database`
>     Is this correct?

For now the intent is that this would be a gene symbol defined in the same way as the gene symbols for [vdj]_call symbols are defined. That is, if you have `GermlineClassSet`.`germline_alleles` array values assigned in `DataProcessing.germline.receptor_germline.germline_alleles`, it is these and only these gene symbols that should be in `[vdj]_call`. This is the set of gene symbols that were used in the annotation.

Yes, the gene symbols in `germline_alleles` and in `[vdj]_call` should be able to be looked up in `germline_database`

> * I assume we do not want people to use `GeneDescription` in connection with `MHCGermlineClassSet`.`germline_alleles`, correct? In this case this should be clarified in the description of the latter, i.e. that individual elements must be a valid gene symbol.

Yes, I think that would be correct. But GeneDescription is currently not used at all in the current spec other than in defining the object itself (the object is not referenced).

bcorrie:
@bussec would your prefer that we remove GeneDescription for clarity since it is not referenced in the rest of the spec? Can you re-review and comment on whether my clarification is sufficient?

bussec:
@bcorrie Yes, taking out the `GeneDescription` object for now and moving this to a later release would address all of these points (as this would also remove the ambiguity mentioned in third and forth point).

bussec (approved):

bcorrie:
@javh do you want to review this? We have pretty broad approval, but you commented in the thread above...

javh:
Naw, I don't need to review - there's enough eyes in here. I'd like to look over this more closely, but I'm a bit behind right now, so don't let me hold you up.

Issue #445, opened 2020-07-23
Repertoire Set


scharch:
Forked from https://github.com/airr-community/airr-standards/issues/441#issuecomment-662795261, in which I suggest the creation of a new `RepertoireSet ` object to become the unit of analysis.

What might a `RepertoireSet` look like?
* I think it is 1:N with `Study`; that is, a paper can do/contain multiple analyses, each potentially on a different set of data, eg mouse vs human or vaccinated vs control.
  * That isn't to say that `Repertoire`s themselves shouldn't continue to encode the `Study` they are originally from. Hopefully it's okay to have it in both places.
* Obviously, it contains a record of the included `Repertoire`s, but I don't think that an array is sufficient. If, for example, I want to analyze only the IgG sequences from [this paper](https://pubmed.ncbi.nlm.nih.gov/30664748/), `RepertoireSet` should contain a way to encode that filtering.
* It needs a way to describe the analyses done. `DataProcessing` should work fine for this, again assuming that multiple levels is ok.
* What if I want to analyze both clonality, and say, [patterns of SHM](https://www.biorxiv.org/content/10.1101/2020.07.21.213686v1) (to pick a hot-off-the-presses example) on the same data? I lean toward requiring them to be in separate `RepertoireSet`s. Otherwise, we're back to the N-to-N issue that #441 started with and `Clone` would have to have `clone_processing_id` in addition to a `repertoire_set_id`. I think they're light-weight enough that the duplication should be ok...

scharch:
Oh, and it probably has to have fields for finding the relevant `Repertoire`s on disk (#439)...

scharch:
Belatedly realized that this could have been filed under #374...

Issue #441, opened 2020-07-17
If I could redesign Repertoire and its buddies...


schristley:
It is said that hindsight is 20/20 and I'm not sure that I see that clearly, but given the expansion of the AIRR Data Model, some flaws are becoming apparent. Here's your chance to pile on!

What did we do right?

- `Repertoire` as the abstract unit of analysis was the correct thing to do because it breaks the "one sample/one file" mentality, which is rigid and limited. It should allow flexibility in combining samples (sequencing data) for different analysis comparisons. It should also allow easy access to metadata, both experimental and computational.

What are the major flaws?

- One flaw is that creating additional `Repertoire`s that represent different mixtures of samples (sequencing data) is inherently expensive to do, when it should be easy and inexpensive. Those additional repertoires invariably require duplicating rearrangement data, because a single rearrangement record can only link to a single `Repertoire`. The answer isn't to change that field into a list...

- Another flaw is the `Repertoire` object is trying to do too much, and thus it compromises to meet all the objectives.  It's trying to be a structured record of study metadata. It's also trying to be a structured record of data processing protocols. It should restrict itself to just defining the unit of analysis and leave the other stuff to other objects.

What would I change?

- Pull `DataProcessing` out of `Repertoire` and make it its own top-level independent object. It was initially put in so that "all study metadata" could be conveniently accessed in a single object. Then it needed to be an array to handle multiple analysis workflows. Now it's got its tendrils into everything...

- After pulling out `DataProcessing`, change `sample` from an array to a single object, then give that whole thing (study, subject, sample) a new name. Maybe it's simply a `SampleProcessing` record.

- Reorient `Repertoire` around `Clone` instead of `Rearrangement`. Clones really become the primary object for downstream analysis.

What does `Repertoire` look like then? Not sure exactly, but it would be a light weight composite object that mostly links to other objects. Creating a new repertoire should be easy and does not require updating ids, duplicating data, or otherwise changing the primary data.

For this to work, `DataProcessing` needs to be redesigned as well. While `Repertoire` is primary a grouping concept, `DataProcessing` is a transform concept that takes input and generates output, i.e. takes raw sequences to generate rearrangements, takes rearrangements to generate clones, and so forth. `SampleProcessing` is also a transform concept, going from biological stuff to a raw sequencing data. Though `SampleProcessing` is essentially constant because it's already been done, while `DataProcessing` is a changing construct as analysis proceeds. Finding the right relationships between the three would be key to any new design.

----

[...]

PR #453, opened 2020-08-25
DataProcessing as a top-level object and independent of Repertoire

schristley:
TODO: need to incorporate design decisions from the `DataProcessing` sprint.

--- this is old ---

Allowing only a single `DataProcessing` object for a `Repertoire` object implies:

- If the same set of sequencing files are data processed in multiple ways, e.g. MiXCR for one workflow and IgBlast for another, those require separate `Repertoire` objects with their own id.

- That single `DataProcessing` is implicitly the primary annotation, so there is no need for the `primary_annnotation` flag.

- Both `repertoire_id` and `data_processing_id` are not needed together to select the appropriate set of rearrangements for a Repertoire, now `repertoire_id` is sufficient. Thus, `data_processing_id` can be removed from `Rearrangement` unless it's needed there for another purpose.

- This makes the `Repertoire` concept more solid as being the annotated rearrangements for a biological repertoire for a single subject. A more abstract object, like `RepertoireSet` #445 , is needed to support flexible analysis.

This doesn't address other issues regarding the `DataProcessing` object.

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1641,10 +1641,8 @@ Repertoire:
                 nullable: false
                 adc-query-support: true
         data_processing:
-            type: array
-            description: List of Data Processing objects
-            items:
-                $ref: '#/DataProcessing'
+            $ref: '#/DataProcessing'
+            description: Data Processing object
             x-airr:
                 nullable: false
                 adc-query-support: true
--- a/lang/python/airr/specs/airr-schema.yaml
+++ b/lang/python/airr/specs/airr-schema.yaml
@@ -1641,10 +1641,8 @@ Repertoire:
                 nullable: false
                 adc-query-support: true
         data_processing:
-            type: array
-            description: List of Data Processing objects
-            items:
-                $ref: '#/DataProcessing'
+            $ref: '#/DataProcessing'
+            description: Data Processing object
             x-airr:
                 nullable: false
                 adc-query-support: true
--- a/specs/airr-schema.yaml
+++ b/specs/airr-schema.yaml
@@ -1641,10 +1641,8 @@ Repertoire:
                 nullable: false
                 adc-query-support: true
         data_processing:
-            type: array
-            description: List of Data Processing objects
-            items:
-                $ref: '#/DataProcessing'
+            $ref: '#/DataProcessing'
+            description: Data Processing object
             x-airr:
                 nullable: false
                 adc-query-support: true


schristley:
Additional changes would be needed to the validation routines, the blank repertoire template, etc., before this PR could be finalized.

scharch:
I'm sure this in the "etc," @schristley, but to keep @bussec happy let's make explicit that it should also included changes to the docs making clear how we expect `Repertoire` to be used... 😊

bcorrie:
Do we want to make this change before we sort out #441?

Making this change has a pretty dramatic follow on effect, in particular on repositories that currently might have/need a data_processing_id and repertoire_id to link rearrangements back to repertoires. This means that every record in a repository (both repertoire and rearrangement records) would need to be updated.

I would **hate** to do that process on the 2 Billion odd rearrangements that we have in our repositories and then we change our mind because we decide to change Repertoire again. I strongly suggest that we sort out how all of Repertoire, Rearrangement, Cell, Clone etc all fit together and interact with DataProcessing before we make a change with such a significant impact. This feels like a 2.0 change to me...

schristley:
> Do we want to make this change before we sort out #441?

Probably not, but if we get far enough along with #441 then it might be worthwhile to put this in with other changes to prevent the array from being used too extensively.

> Making this change has a pretty dramatic follow on effect, in particular on repositories that currently might have/need a data_processing_id and repertoire_id to link rearrangements back to repertoires. This means that every record in a repository (both repertoire and rearrangement records) would need to be updated.
> 
> I would **hate** to do that process on the 2 Billion odd rearrangements that we have in our repositories and then we change our mind because we decide to change Repertoire again.

Agreed. I suppose the good thing is that IPA only uses a single data processing, and same for VDJServer, so that covers a large percentage. There's an extra `data_processing_id` field in rearrangements, which isn't needed, but it isn't mandatory that it be removed either.

schristley:
One point to make is that even though this structure is fine for curation, when there is one final data processing that's released for publication, it doesn't support the discovery/exploration/analysis workflow as well.

For example, let's say I want to try different quality filters and other pre-processing parameters. I run my pre-processing tool on the raw sequencing files and get one set of output. Then I run the tool again with a different set of parameters and get another set of output. Let's say the tool is special, it knows about `Repertoire`, and in fact as input, you give it a list of `Repertoires` and that's how it knows which raw sequencing files to process. Furthermore, the tool knows about `DataProcessing` and automatically fills in the information in the `Repertoires` about how it was run, with its parameters and such. Every time I run the tool, it's going to replace the data processing with new info, and there really isn't a direct way to keep track of those multiple runs, for example to compare them.

I don't consider this a major issue, but it's something to think about if you are developing a tool that will _write_ metadata information versus just reading it.

scharch:
> Every time I run the tool, it's going to replace the data processing with new info

Interesting. The way I've been thinking about this for SONAR is actually updating, not overwriting, because SONAR is a suite of scripts and each step creates its own log entry already. So I actually end up with a different problem, which is how to "simplify" a huge agglomerated `data_processing` object to only the non-redundant steps at the end...

schristley:
> > Every time I run the tool, it's going to replace the data processing with new info
> 
> Interesting. The way I've been thinking about this for SONAR is actually updating, not overwriting, because SONAR is a suite of scripts and each step creates its own log entry already. So I actually end up with a different problem, which is how to "simplify" a huge agglomerated `data_processing` object to only the non-redundant steps at the end...

This is, in a way, why I added `analysis_provenance_id`, where from a VDJServer perspective, instead of filling up the `DataProcessing` object with lots of VDJServer-specific stuff, I can keep it in a separate object, referenced by that id. Now the spec says it points to a [PROV](https://www.w3.org/TR/2013/NOTE-prov-overview-20130430/) model (sadly the [python](https://prov.readthedocs.io/en/latest/) package is kinda crappy), but I don't think that's something we've discussed with any depth.

Imagine this, `DataProcessing` has this indicating that 3 tools were run:

```
software_version: "vdj_pipe-ls5-0.1.7u8 igblast-ls5-1.14u2 sonar-ls5-4.2u1"
```

The PROV model has then, at least, 3 high-level activities, for the 3 tools that were run. Here's might how I would record the provenance for running IgBlast:

```
document = prov.ProvDocument()

# an entity for the input file
e1 = document.entity('e1', (
    (prov.PROV_TYPE, "File"),
    ('ex:path', "my_sequences.fasta"),
    ('ex:creator', "Alice"),
))

# an entity for the output file
e2 = document.entity('e2', (
    (prov.PROV_TYPE, "File"),
    ('ex:path', "rearrangements.tsv"),
    ('ex:creator', "Alice"),
))

# create an activity for the igblast tool
a1 = document.activity('igblast-ls5-1.14u2', datetime.datetime.now(), None, None)

# indicate output file was derived from input file by the activity
document.wasDerivedFrom(e2, e1, a1)
```

Now in theory, this provides analysis provenance in a standard data model, and you could get as detailed and fine-grained as you want, or stay pretty high level. The resultant PROV document can be saved as JSON.

There is a [document](https://docs.google.com/document/d/1DLMUFPGrjoQCE4XsK26Bzu72Ig0-bTMI2PMlQhuhKoE/edit#heading=h.hsojms563y7g) that I wrote awhile back with my thoughts about analysis provenance.

The provenance that we maintain in VDJServer is hard-coded and non-standard, so it's still a long-time feature request to put in all in the PROV data model.

The question is, should AIRR try adopting this more formally for tools that want to _update_ an ongoing data processing with it's activities?

scharch:
Are you suggesting "replacing" `data_processing` with a PROV document? I could see that as a long-term enhancement...

schristley:
> Are you suggesting "replacing" `data_processing` with a PROV document? I could see that as a long-term enhancement...

No, only that it references a PROV document with `analysis_provenance_id`. I see `data_processing` as useful to provide some summary information.

javh (approved):
Fine by me.

bussec:
@schristley The proposed change seems reasonable to me, but this PR has been sitting here for a while and #441 was also closed a while ago. Did we discuss this more recently than 2020?

schristley:
There's been extensive discussion (and design) around `DataProcessing` as a top-level object. However, there's been minimal implementation so far. This PR predates the decision for `DataProcessing` as a top-level object, and essentially represents an alternative design but it that has some flaws, so I don't want to commit it. I didn't delete because I didn't want to forget my comments with @scharch . Instead, I've changed to draft, changed the milestone to V2, and changed the title to reflect the new design.

javh:
Thanks, @schristley. One more PR down. :)

javh:
From the call:
* Decide if this is still relevant. Last commit was 2020.
* Will leave open for discussion purposes.

bcorrie:
@schristley should we move this to v2.1 with other DataProcessing changes?

schristley:
Cleared milestone as its unknown when/if data processing will be re-designed.

javh:
> Cleared milestone as its unknown when/if data processing will be re-designed.

Seems highly unlikely we end up redesigning DataProcess at this point. I'm inclined to let this one fade.

javh:
@schristley Are you okay with us closing this?

schristley:
stale

PR #455, opened 2020-08-27
Add v_frameshift, j_frameshift and d_frame fields.

javh:
I received a request from @jianye00 (IgBLAST developer) to add a reversed field name for a boolean to specify whether the observed V segment contains an internal frameshift, due to indel mutations or alignment against a defective pseudogene. They plan to add this field to the IgBLAST V-(D)-J rearrangement summary table.

This differs from `vj_in_frame`, which considers whether the V and J are in the same frame, but not whether the query V segment is in frame with its germline reference sequence.

In the spirit of completeness, I also added a boolean for J frameshifts and a numerical field for the D reading frame. The additions are:

```
v_frameshift:
    type: boolean
    description: >
        True if the V gene in the query nucleotide sequence contains a translational
        frameshift relative to the frame of the V gene reference sequence.
j_frameshift:
    type: boolean
    description: >
        True if the J gene in the query nucleotide sequence contains a translational
        frameshift relative to the frame of the J gene reference sequence.
d_frame:
    type: integer
    description: >
        Numerical reading frame (1, 2, 3) of the first or only D gene in the query nucleotide sequence,
        where frame 1 is relative to the first codon of D gene reference sequence.
d2_frame:
    type: integer
    description: >
        Numerical reading frame (1, 2, 3) of the second D gene in the query nucleotide sequence,
        where frame 1 is relative to the first codon of D gene reference sequence.
```

Are there any concerns with adding these fields to v1.4 and/or problems with the definitions?

### DIFF
--- a/docs/datarep/rearrangements.rst
+++ b/docs/datarep/rearrangements.rst
@@ -244,8 +244,20 @@ Alignments
 There is no required alignment scheme for the nucleotide and amino acid alignment
 fields. These fields may, or may not, include numbering spacers (e.g., IMGT-numbering gaps),
 variations in case to denote mismatches, deletions, or other features appropriate to the tool that
-performed the alignment. The only strict requirement is that the query ("sequence") and
-reference ("germline") **must** be properly aligned.
+performed the alignment. The only strict requirement is that the query (``sequence``) and
+reference (``germline``) **must** be properly aligned.
+
+Frameshifts
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+For purposes of annotating alignments, a frameshift is defined as a frameshift that is
+maintained until the end of the aligned gene, where frames are designated numerically as
+1 (in-frame), 2, or 3. For example, an V gene alignment that starts in
+frame 1 and ends in frame 2, disrupting the conserved cystine, would be defined as a frameshift.
+Whereas, a V gene alignment with an internal frameshift that corrects with a second frameshift,
+back to the original frame 1 prior to the conserved cystine, would not need to be annotated
+as a frameshift.
+
 
 Fields
 ------------------------------
--- a/docs/standards/news.rst
+++ b/docs/standards/news.rst
@@ -3,6 +3,18 @@
 Schema Release Notes
 ================================================================================
 
+Version 1.4.0: (In development)
+--------------------------------------------------------------------------------
+
+**Version 1.4, In development.**
+
+Rearrangement Schema:
+
+1. Added the optional fields ``v_frameshift``, ``j_frameshift``,
+   ``d_frame`` and ``d2_frame`` defining annotations related to alignment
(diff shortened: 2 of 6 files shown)


bussec:
I am in general ok with adding these fields, however there are two things that should be clarified:

1. Is an "internal frameshift" only a frameshift that is maintained until the end of the segment (e.g. start in frame 1, end in frame 2) or would this also apply to shifts that are balanced again (e.g. start in frame 1, shift to frame 2, shift back to 1)?
2. Is the "1,2,3" numbering in line with other frame numberings we have? There are multiple ways to designate them so it should be consistent.

javh:
1. Good question about the balanced frameshift. I'm inclined to say this would apply. The use case seems to be another sort of "productivity" flag to identify sequences that may have a non-stop defect.
2. As I know, we don't have any frame specifications in the spec currently.  Is there something in the germline standard for this already? Do we want to establish a controlled vocabulary for this now?

Pinging @williamdlees.

williamdlees:
The germline schema uses a 1,2,3 designation in a related manner. I think referring to the first, second and third bases of a codon is customary.

William

codon_frame:
type: [1,2,3]  
description: Codon position of the first sequence symbol in the Coding Sequence. Mandatory for J genes. Not used for V or D genes. ('1' means the sequence is in-frame, '2' means that the first bp is missing from the first codon, '3' means that the first 2 bp are missing)
label: Codon Frame

jianye00:
You can definitely have an internal coding frame shift that is later restored to normal frame.  But it is difficult to define its effect on the Ig functionality depending on which part is affected and how long the shift length is.  What I have in mind, at least for igblast use case, is a simpler definition.  It will be labeled frame shift in V gene if the frame shift is still there at the end of FWR3.  This means the highly conserved Cys is not in the coding frame for almost all germlines which likely disrupts the Ig function.

Jian  


On Monday, August 31, 2020, 11:15:25 AM EDT, Christian Busse <notifications@github.com> wrote: 







I am in general ok with adding these fields, however there are two things that should be clarified:

    1. Is an "internal frameshift" only a frameshift that is maintained until the end of the segment (e.g. start in frame 1, end in frame 2) or would this also apply to shifts that are balanced again (e.g. start in frame 1, shift to frame 2, shift back to 1)?
    2. Is the "1,2,3" numbering in line with other frame numberings we have? There are multiple ways to designate them so it should be consistent.

schristley (commented):
Please include updates to the documentation, maybe in the Definition Clarifications section?

schristley (commented):
Also, need to put these in the OpenAPI V3 version of the schema.

javh:
From the call:

javh:
I added a Frameshifts sections the Definition Clarifications and propagated the fields into the OpenAPI V3 spec.

I think this is ready to go.

PR #458, opened 2020-09-15
Add a study_contact field

### LINKED ISSUE #444: So what about pub_authors or contact_person
We have contact information fields for who collected the data and who deposited the data, but we don't have a field to capture the contact person for the study. This has always seemed quite odd to me, but I understand the focus on MiAIRR as a standard to describe a study. But shouldn't there be a principal contact for the study - which in most cases would probably be the corresponding author on the relevant publication for the study?

As I say, this has always seemed odd to me, but with all of the traffic we are seeing on the iReceptor Gateway with the COVID-19 data, people are looking for studies based on authors and are expressing "frustration" that they can't find this. So this seems problematic.

We have hijacked the collected_by and deposited_by fields to capture relevant information about people, but you know that when you start to do that there is a flaw in the model...

So how do people feel about having one or both of pub_authors (to accompany pub_ids) or contact_person as the principal contact for the Study?

bcorrie:
Closes #444 - I think this should be a MiAIRR addition, possibly for 1.4?


schristley:
How about we put the ORCID into its own field, like `study_contact_orcid`? This would actually allow it to be used more readily without trying to parse it from the free text.

bcorrie:
> How about we put the ORCID into its own field, like `study_contact_orcid`? This would actually allow it to be used more readily without trying to parse it from the free text.

I would not object to that, but we have other contact fields (collected_by and submitted_by) and I was trying to maintain consistency with those. Note this is related to (and maybe more appropriate for) discussion in #464 where I talk about making better use of persistent IDs in general around the AIRR spec.

@bussec is orcid a widely used PID to refer to people in the EU context? Should we be adding such fields - I am thinking of these as MiAIRR fields, as for scientific reproducibility, the people are rather important 8-)

bcorrie:
Maybe we can merge this pull request into master as is, and then leave the creation of _orcid (or maybe more generally _pid to not tie is to Orcid) for this and other fields to #464?

bcorrie:
@bussec @schristley any concerns about this? Can we merge this - with the caveat that the PID part of the contact might get pulled out into a separate field through #464

schristley (approved):

schristley:
Note that the schema as written says that `study_contact` is a MiAIRR important field. Technically this is an addition to MiAIRR and thus part of MiAIRR V2. If we want this in 1.4.0 then removing it from MiAIRR is the easiest change, or we need to come up with justification for why adding a field to MiAIRR V1 is okay.

bcorrie:
OK, lets see what @bussec take is on this. Probably want to remove the MiAIRR important field now, put the field in 1.4, and the promote to MiAIRR in v2.0, although maybe we rework the contact stuff as per https://github.com/airr-community/airr-standards/issues/444#issuecomment-709562528 for v2.0.

bussec:
Looks good, just update the contact data for the Florian study (as Dr. Stibbons was not involved here).

bussec (approved):

PR #463, opened 2020-09-24
handle case when more extra data in a row

### LINKED ISSUE #457: airr-tools rearrangement validation error
It seems that validating a rearrangement file with airr-tools produces the following error if the header has fewer fields than the record proceeding it:

> Error occurred while validating AIRR rearrangement files: 'NoneType' object has no attribute 'endswith'

For example [my_airr.txt](https://github.com/airr-community/airr-standards/files/5188691/my_airr.txt) produces this error when provided to airr-tools (version 1.3.0).

    airr-tools validate rearrangement -a  my_airr.txt

schristley:
fixes #457 

This gives a proper validation error now, and also should not abort when reading such a file.

```
$ airr-tools validate rearrangement -a my_airr.txt 
Validating: my_airr.txt
my_airr.txt at record 1 has validation error: row has extra data
```

A related question is if `airr.read_rearrangement`, when not validating, should it do something under the circumstance? Print a warning, fail, remove the data entry, leave as is? Right now it returns the row entry with a null header name:

```
OrderedDict([('sequence_id', '1'), ('sequence', '2'), ('rev_comp', False), ('productive', False), ('v_call', '5'), ('d_call', '6'), ('j_call', '7'), ('sequence_alignment', '8'), ('germline_alignment', '9'), ('junction', '11'), ('junction_aa', '12'), ('v_cigar', '13'), ('d_cigar', '14'), ('j_cigar', '15'), (None, ['not_in_header'])])
```

### DIFF
--- a/lang/python/airr/interface.py
+++ b/lang/python/airr/interface.py
@@ -243,6 +243,7 @@ def load_repertoire(filename, validate=False, debug=False):
 
     # validate if requested
     if validate:
+        valid = True
         reps = md['Repertoire']
         i = 0
         for r in reps:
@@ -253,6 +254,8 @@ def load_repertoire(filename, validate=False, debug=False):
                 if debug:
                     sys.stderr.write('%s has repertoire at array position %i with validation error: %s\n' % (filename, i, e))
             i = i + 1
+        if not valid:
+            raise ValidationError('Repertoire file %s has validation errors\n' % (filename))
 
     # we do not perform any additional processing
     return md
--- a/lang/python/airr/io.py
+++ b/lang/python/airr/io.py
@@ -91,6 +91,11 @@ class RearrangementReader:
             raise StopIteration
 
         for f in row:
+            # row entry with no header
+            if f is None:
+                if self.validate:
+                    raise ValidationError('row has extra data')
+
             # Convert types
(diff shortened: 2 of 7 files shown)


scharch:
> A related question is if `airr.read_rearrangement`, when not validating, should it do something under the circumstance?

Printing a warning sounds reasonable to me. Although I could see failing, too, since there's no guarantee the unexpected field is at the end...

schristley:
> since there's no guarantee the unexpected field is at the end

That's true, and if it's in the middle, that shifts the later fields which could then be misinterpreted as different types, returning garbage data. That would suggest throwing an error because a warning might be ignored.

javh:
Abort with an error sounds best to me too.  This only seems like it could arise from a malformed file. And aborting would force users to fix the file, which is to their benefit. Eg, maybe they have a tab character inside a column value or they wrote the file from R using a matrix format (first column is an unnamed numerical row index). In either case the whole row/file can be wrong, instead of just the last entry.

schristley:
Found a bug with `airr.load_repertoire` when `valid=True`, and fixed tests which weren't handling exceptions properly.

javh (approved):

PR #465, opened 2020-09-28
Improvements for PID usage in the spec.

### LINKED ISSUE #464: Have fields refer to Persistent IDs as often as possible
It seems fairly obvious, but on review we don't actually specify this is the spec. For example, no mention of DOIs/Pubmed ID for papers (other than in the example), no mention of something like ORCID for people (submitted_by, collected_by), and only a brief mention of using an NCBI BioProject ID or similar value in a field like study_id. 

Should we be stronger in suggesting more rigour around this? I think we can do this without being overly prescriptive/restrictive.

I think probably in an ideal world, we might have a separate field for some of these (e.g. submited_by_pid, collected_by_pid, pub_ids_pid) so that using external information would be clear, but at a minimum it seems like we should be saying that for pub_ids you should refer to a known persistent identifier like a DOI or Pubmed ID...

This is related to both #444 and #435 in terms of being explicit about referring to external PIDs like ORCID and SRA run ids or similar.

bcorrie:
Fixes #464

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1363,7 +1363,9 @@ Study:
     properties:
         study_id:
             type: string
-            description: Unique ID assigned by study registry such as one of the International Nucleotide Sequence Database Collaboration (INSDC) repositories.
+            description: >
+                Unique ID assigned by study registry such as one of the International Nucleotide Sequence Database
+                Collaboration (INSDC) repositories.
             title: Study ID
             example: PRJNA001
             x-airr:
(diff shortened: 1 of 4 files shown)


bcorrie:
Fixes #464

bussec:
@bcorrie Do you still want this for v1.4 or do we push to v2.0?

bcorrie:
@bussec I think if we are going to do PIDs more generally, then it makes sense to do this as v2.0 and make sure we hit all fields that make sense to have PIDs for.

bussec:
@bcorrie Ok, then let's move back to #464 to collect thoses fields and PIDs.

bcorrie:
@bussec want to have a quick look at the PID description use in the field names.  A precursor to doing it properly, but worth having in v1.4 I think.

bcorrie:
I don't think this closes #464 any more - that should be done properly as part of v2.0.

PR #466, opened 2020-09-28
Added cell entry points to ADC spec

bcorrie:
Related to #404 and #211 - Added already defined objects to the ADC API spec. 

Note that the Receptor object has not been added as I think there are some things up in the air still???


bussec:
Will handle Receptors via #404

bcorrie:
@schristley do you want to have a look at this...

schristley (approved):

Issue #426, opened 2020-06-29
Define a manifest mechanism


javh:
We need a way to link related files together on the file system. The immediate need is to link Rearrangement TSVs Cell JSON, and expression matrix files (#409), but applies equally to metadata and Clone/Tree/Node files.

Some notes from the 2020-06-29 call:
* The file manifest should live in the metadata file, which will be the "primary" file for the data set.
* We could define a top-level `Manifest` object to store file relations (has the advantage of clarity).
* We could put file links in `DataProcessing`, like the fastq file links (similar to an existing mechanism).
* We could extend the `Info` object with a file manifest (similar to an existing mechanism).

Random thoughts:
* Does the manifest need to define which identifier fields are present in the individual files?
* Do we need URIs and/or checksums to ensure a collection of files is correct/unchanged?
* Do we need metadata for the files themselves (format, creation date, etc)?

franasa:
coming back to the manifest issue, we have some specific questions regarding the properties `filename` and `file_type` that define the file formats for the raw reads or sequences:
 
1) Are these properties to be included on the manifest?
 
both are `miairr: important` but in our case, we don't store unprocessed data files in our db, instead we do import the raw sequences of those files and scireptor performs some pre-processing on them. 

2) Is it expected for such a use case to also export the fastq files together with the files to be defined in the manifest?
 
there is also the property `data_processing_files` which is
an array of file names for data produced by this data processing. 

3) Are these files different to, for example, the TSV, YAML, or the files that we are defining for single cell in #409?

4) Does the definition of the manifest will include both the  RawSequenceData- and DataProcessing-files or would these be part of custom fields that you mention in [#478](https://github.com/airr-community/airr-standards/pull/478#issuecomment-728282251)?

bcorrie:
> coming back to the manifest issue, we have some specific questions regarding the properties `filename` and `file_type` that define the file formats for the raw reads or sequences:
> 
> 1. Are these properties to be included on the manifest?

To be defined as to what goes in the manifest I think - see my comment here: https://github.com/airr-community/airr-standards/pull/478#issuecomment-735263668
> 
> both are `miairr: important` but in our case, we don't store unprocessed data files in our db, instead we do import the raw sequences of those files and scireptor performs some pre-processing on them.

[...]

PR #478, opened 2020-11-10
schema for AIRR data file

schristley:
An explicit schema for an AIRR data(metadata) file. This is an extension of the file format for Repertoire metadata (which just has the `Info` and `Repertoire` keys) to include the additional AIRR objects. The format continues the earlier design where the output from the ADC API exactly matches the file format. Also an `Info` object schema is provided. This was previously only in the ADC API spec.

`Rearrangement` implies rearrangement data in JSON (versus TSV). We might want to consider how to extend this format for #426 so that the data file can refer to TSV files.

This shows how `DataProcessing` can be a "top-level" object.

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -139,6 +139,88 @@ Attributes:
                             type: string
                             description: Ontology name for the top node term
 
+# AIRR Data File
+#
+# A JSON data file that holds Repertoire metadata, data processing
+# analysis objects, or any object in the AIRR Data Model.
+#
+# It is presumed that the objects gathered together in an AIRR Data File are related
+# or relevant to each other, e.g. part of the same study; thus, the ID fields can be
+# internally resolved unless the ID contains an external PID. This implies that AIRR
+# Data Files cannot be merged simply by concatenating arrays; any merge program
+# would need to manage duplicate or conflicting ID values.
+
+DataFile:
+    discriminator: AIRR
+    type: object
+    properties:
+        Info:
+            $ref: '#/InfoObject'
+        Repertoire:
+            type: array
+            description: List of Repertoire metadata
+            items:
+              $ref: '#/Repertoire'
+        Germline:
+            type: array
+            description: Germline sets
+            items:
+              $ref: '#/GermlineSet'
+        DataProcessing:
+            type: array
+            description: List of data processing workflows
+            items:
+              $ref: '#/DataProcessing'
+        Cell:
+            type: array
+            description: List of cells
+            items:
+              $ref: '#/Cell'
+        Rearrangement:
+            type: array
+            description: List of rearrangement records
+            items:
+              $ref: '#/Rearrangement'
+        Clone:
+            type: array
+            description: List of clones
+            items:
+              $ref: '#/Clone'
+        Tree:
+            type: array
+            description: List of trees
+            items:
+              $ref: '#/Tree'
+
+# AIRR Info object, should be similar to openapi
+# should we point to an openapi schema?
+InfoObject:
+    type: object
+    description: Provides information about data and API responses.
+    properties:
+      title:
+        type: string
+      version:
+        type: string
+      description:
+        type: string
+      contact:
+        type: object
+        properties:
+          name:
+            type: string
+          url:
+            type: string
+          email:
+            type: string
+      license:
+        type: object
+        properties:
+          name:
+            type: string
+          url:
+            type: string
+
 #
 # Germline gene schema
 #
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -161,11 +161,6 @@ DataFile:
             description: List of Repertoire metadata
             items:
               $ref: '#/Repertoire'
-        Germline:
-            type: array
-            description: Germline sets
-            items:
-              $ref: '#/GermlineSet'
         DataProcessing:
             type: array
             description: List of data processing workflows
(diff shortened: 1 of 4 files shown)


schristley:
When we define a schema for `RepertoireGroup` or `RepertoireSet`, that can be added in the format as an additional key.

bcorrie:
This kind of feels to me like it should be in a different .yaml file and refer to the other object definitions in airr_schema.yaml like the ADC API schema does???

This seems to need a airr_fileformat.yaml or something like that - do we want to combine the two?

bcorrie:
There is an Info and an InfoObject - can we unify these two?

bcorrie:
Also, if I understand correctly, this would subsume our AIRR Repertoire file format definition, which is essentially:
```
DataFile:
    discriminator: AIRR
    type: object
    properties:
        Info:
            $ref: '#/InfoObject'
        Repertoire:
            type: array
            description: List of Repertoire metadata
            items:
              $ref: '#/Repertoire'
```
Did I get that correct?

schristley:
> Also, if I understand correctly, this would subsume our AIRR Repertoire file format definition, which is essentially:

Yes, exactly. Our current definition is only described in the documentation, so this makes it formal and machine readable.

schristley:
> There is an Info and an InfoObject - can we unify these two?

They are different things. InfoObject is a schema definition (class) while Info is a concrete instance of InfoObject, in this case the Info for the airr-schema.yaml file. Normally, I would have labeled InfoObject as just Info but that obviously conflicts...

scharch:
Do we really want to create a schema where eeeeeverything is in one file? Where you potentially have to load alllll the `Rearrangement`s to access the metadata? Even for JSON, I think I would prefer a manifest that allows more flexibility with putting different pieces in different places...

schristley:
> Do we really want to create a schema where eeeeeverything is in one file?

I don't mean to imply this with the schema, though maybe I suggest with my comment. Each of the keys are optional, so a data file can be any mixture of data. I don't expect all data to be in one file, but in a set of these files. These files might naturally correspond to output from queries to the different endpoints (repertoire, clone, cell) in the ADC API.

> Even for JSON, I think I would prefer a manifest that allows more flexibility with putting different pieces in different places...

Yep, this schema is just one part, a manifest needs to say what files are grouped together.

Rearrangements are a special case because we already have a TSV file format, and I expect rearrangement data to be in their own files. I mostly have the key there for completeness, so that it matches the `/rearrangement` end point in the ADC API. The same may be true for other objects like clones, if we define a TSV format for them.

scharch (approved):

javh:
From the call:
Should we add a "Custom" property that holds a list of non-AIRR files? The intent being to allow users to bundle random files in manifest/directory/tarball with the AIRR schema-defined files.

bcorrie:
Here is an example download "manifest" from the iReceptor Gateway:

```
* Summary *
Total: 21 sequences
covid-19-2.tsv: 0 sequences (1.8 KB)
covid-19-1.tsv: 7 sequences (4.1 KB)
covid-19-3.tsv: 4 sequences (3.3 KB)
covid-19-4.tsv: 10 sequences (5.6 KB)


* Metadata filters *
None

* Sequence filters *
Junction/CDR3 AA: CATSEGQGSQPQHF

* Source *
/sequences?query_id=8653
Downloaded by bcorrie on Nov 16, 2020 at 20:34 UTC
```
The ZIP file contains the following files:
```
covid-19-1.tsv
covid-19-2.tsv
covid-19-3.tsv
covid-19-4.tsv
covid-19-1-metadata.json
covid-19-2-metadata.json
covid-19-3-metadata.json
covid-19-4-metadata.json
```

bcorrie:
A more complicated download manifest whose query hits all of the repositories:

```
* Summary *
Total: 70 sequences
ireceptor-public-archive-3.tsv: 7 sequences (4.9 KB)
ireceptor-public-archive-2.tsv: 41 sequences (17.2 KB)
ireceptor-public-archive-4.tsv: 0 sequences (1.8 KB)
airr-covid-19-2.tsv: 0 sequences (1.8 KB)
ireceptor-public-archive-5.tsv: 8 sequences (5.5 KB)
airr-covid-19-1.tsv: 0 sequences (1.8 KB)
airr-covid-19-4.tsv: 2 sequences (2.6 KB)
airr-covid-19-3.tsv: 1 sequences (2.2 KB)
ireceptor-public-archive-1.tsv: 0 sequences (1.8 KB)
vdjserver.tsv: 11 sequences (20.2 KB)


* Metadata filters *
None

* Sequence filters *
Junction/CDR3 AA: CSARRGLAGSDTQYF

* Source *
/sequences?query_id=24427
Downloaded by bcorrie on Nov 16, 2020 at 20:44 UTC
```

bcorrie:
I have fundamental question about this file format and its intent. This schema addition (either in the same or a different file) seems to be defining a rigorous file format for any combination of object in the AIRR schema. This is important...

My question is, is this really a Manifest file as implied by the title in #426 and the description in that issue. I think the intent of the manifest file that @schristley is talking about is more of a true manifest file (https://en.wikipedia.org/wiki/Manifest_file) while the YAML description we have is a file format for objects in the schema. I actually think we need both...

For example, I don't think the YAML schema in this pull request would be able to describe the set of files that would be included in an iReceptor download. In terms of a manifest file, don't we want something that formalizes a grouping of data files that captures a set of data that is to be used for a specific purpose. In the iReceptor download case the manifest would describe the set of files that correspond to the results of the potentially complicated query that the user issued through the Gateway.

I think that is more along the lines of what we are talking about when we say manifest file, no?

schristley:
> I actually think we need both...

That's correct, we need both. The manifest should be described with a schema. In the last meeting, we discussed if we should put it in `Info` or maybe it should be it's own `Manifest` object. I'm leaning towards the latter.

PR #491, opened 2021-01-28
asynchronous extension for ADC API

### LINKED ISSUE #490: asynchronous API extension for the AIRR Data Commons
The current ADC API (v1.0) is synchronous, which is simple but presents challenges for service implementors, especially as the data increases in scale. For example, it's possible for a malicious client to overload a server if that server doesn't have safeguards in place. Services can make different choices about how to implement the API in order to mitigate such service attacks, but then this may introduce inefficiencies and/or tradeoffs. The CRWG didn't want to "force" any particular design on service implementors, and thus the API allows for services to have safeguards (e.g. limited data transfer, timeouts, etc.)

I think the simplest way to describe the issue is, how to balance interactivity (small, fast queries) versus downloads (large, slow queries)?

This can also be a concern for clients as they 1) might not know if any particular query is fast or slow, and 2) even if they did know, there's no way to inform the server of their intent.

The small, fast queries aren't the problem, it is the large, slow queries. This API extension will provide for asynchronous query capability which allows services to "scale-up" to handle many large, slow queries without sacrificing performance for interactivity. Essentially, services can decide how many of these queries are being performed concurrently (e.g. to manage load on their backend database) without being "forced" to respond as expected with the synchronous API.

schristley:
closes #490 

An open question is getting the result data.

1. stream the data back through an endpoint. This is slower as the data has to pass back through memory (of the API service) only to possibly be written back to an output file by the client.
2. provide file download. This will require authentication which may be onerous for some clients, e.g. some browsers do not allow downloading to disk from AJAX.
3. provide one-time URLs for file download. This is what VDJServer does for all downloads to disk (for the main website), authentication is performed when the one-time URL is created.

Note that 2 and 3 does not prevent streaming the result data directly into the browser, for example if you wanted to visualize the data in a graph. In that case, VDJServer uses 2 to do that.

Another open question is support for event notifications: query is started, data is ready, error occurred, etc.


schristley:
For getting the result data, the current design requires the API service to provide a `download_url`. This should provide download of the data without needing authentication. The `download_url` should be valid so long as the query has not EXPIRED, and should not be limited in the number of uses.

schristley:
It was asked in a recent CRWG meeting, why have a separate API file, with a different base URL, versus adding a `mode` parameter to the standard ADC API which specifies sync or async? There was some previous discussion about such a parameter, but I'm not sure it was captured anywhere so I'll write down my reasoning for posterity.

While the parameter idea initially sounds simple, there are a number of disadvantages, specifically when it gets to the specification and implementation, that makes having a separate API be a "cleaner" solution.

1. The result JSON would be a different object for sync vs async. For sync, it's the actual data, while for async it is `query_response`. The current ADC API is based upon OpenAPI V2, and that version does not allow for differing return schemas, so there there would be no way to represent those responses in the spec. OpenAPI V3 has added the `oneOf` property which allows for this to be represented, but that would force all repositories to upgrade to OpenAPI V3.
2. The async API also requires additional end points, e.g. a status end point to get the status of the async query. Some middleware packages, e.g. `express-openapi`, requires that all end points have a handler otherwise the service will not start. This would mean that repositories, which don't intend to support async, might (depending on their software stack) need to provide stub implementations (which presumably just return a not-implemented code) for those end points.
3. The async API might have additional required/optional parameters on the end points. These parameters would only be valid for the async mode. There's no way with OpenAPI to enforce required/optional parameters based upon the value of another parameter.
4. Lastly, with sync and async combined in a single API, that tends to imply the same API service process must handle both. Now this can be gotten around by writing a front-end proxy service, which interprets the incoming request, then proxies to either a sync or async service, but that's still extra code and another server process that the repository needs to write, deploy and maintain.

Having a separate API file for async eliminates these issues.

* The API is cleaner/simpler and isn't combining both sync and async semantics.
* If a repository did not intend to support async, it will still be required to modify their code to provide stubs/errors for the async mode to be compliant with the ADC API. With a separate API file, those repositories do not need to make any changes at all.
* Repositories that support async can easily have a separate API service to support async, without altering their current sync service.

This decision also shouldn't make clients more difficult to code. If a client is going to utilize sync and async, it's going to have to handle the differences anyways for parameters and return objects, so having some simple code like this to utilize a different base URL shouldn't be onerous.

```
if (async) base_url = '/airr/async/v1'
else base_url = '/airr/v1'
```

schristley:
notifications are to be added, there should be a mechanism to specify which events to be notified about.

PR #493, opened 2021-02-02
add adc_publish_date and adc_update_date

### LINKED ISSUE #472: publish date for a study in an ADC repository
As the ADC gets larger, a nice facility for the ADC to provide is the ability to show the user the newest or latest studies that have been added. Right now there isn't a date for when a study is added to a repository in the ADC. While there are dates associated with publications and/or a BioProject, they could be very different from when the study actually gets annotated and added into an ADC repository. My suggest is to add an `adc_publish_date` field in `Study` that a repository sets when that study has been made available in the repository for query and download. This should be a retro-active field, that is all existing studies should be given a date, even if it's an estimate.

schristley:
close #472

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -460,6 +460,28 @@ Study:
                 subset: study
                 name: Keywords for study
                 format: controlled vocabulary
+        adc_publish_date:
+            type: string
+            format: date
+            description: >
+                Date the study was first published in the AIRR Data Commons.
+            title: ADC Publish Date
+            example: "2021-02-02"
+            x-airr:
+                nullable: true
+                adc-query-support: true
+                name: ADC Publish Date
+        adc_update_date:
+            type: string
+            format: date
+            description: >
+                Date the study data was updated in the AIRR Data Commons.
+            title: ADC Update Date
+            example: "2021-02-02"
+            x-airr:
+                nullable: true
+                adc-query-support: true
+                name: ADC Update Date
 
 
 # 1-to-n relationship between a study and its subjects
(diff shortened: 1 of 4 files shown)


schristley:
@bcorrie are you still okay with this?

bcorrie:
@schristley in https://github.com/airr-community/airr-standards/issues/472#issuecomment-768398018 we talked about having this for the Repertoire object. Where did we land on this? Currently these dates are in Subject.

schristley:
> @schristley in [#472 (comment)](https://github.com/airr-community/airr-standards/issues/472#issuecomment-768398018) we talked about having this for the Repertoire object. Where did we land on this? Currently these dates are in Study.

@bcorrie William was fine on holding off on something more detailed until it was actually needed, and I'm fine with keeping it simple until more is required.

bcorrie:
Ok - and I see I made an error above as well - the dates are in Study, which it looks like you corrected?

Do we define any semantics on this, in particular `adc_update_date`? Or leave it up to the repository?

Also, are we talking about date-time or just date?

schristley:
> Do we define any semantics on this, in particular `adc_update_date`? Or leave it up to the repository?

Nothing beyond what's in the description for the field. Do we need something more precise?

> Also, are we talking about date-time or just date?

The fields have `format: date` which is just the date, do we want the time also?

schristley:
After CRWG, will use `date-time`, as it's reasonable that a study maybe be updated multiple times in a single day (e.g. fix a typo or missing data from an initial push). A client that just happened to read the data between updates (and maybe cached something in the process) wouldn't know the data has actually been updated again unless the time was available.

schristley:
oops, forgot to change to `date-time`

PR #494, opened 2021-02-02
clean up the SampleProcessing object

### LINKED ISSUE #325: SampleProcessing object incomplete??
Is it just me or is there stuff missing here? https://github.com/airr-community/airr-standards/blob/75456c32b3f967877045c8d07b5dfd0df1d79ed9/specs/airr-schema.yaml#L1259 --there is only a `sample_processing_id` field, even though the description makes it sound like there should be fields for `sample_id`, `cell_processing_id`, etc....

schristley:
closes #325


schristley:
Our schema python code doesn't handle `allOf` at the top-level of an object... sigh

schristley:
Hello @javh @imkeller @ustervbo , I was able to modify the python library to support `allOf`. The change is that `allOf` is at the top-level of an object (`SampleProcessing` specifically), versus within an array. Also, it is not all `$ref`s. I attempted to do the same thing for R. I was able to modify the `load_schema` function in `Schema.R`, though it tends to duplicate some code in `extract_field_content`. However, the validation doesn't work in `Interface.R`, the test suite gives an error:

```
Error (test-Interface.R:156:5): read_repertoire loads a list
Error: missing value where TRUE/FALSE needed
Backtrace:
 1. airr::read_repertoire(repertoire_file) test-Interface.R:156:4
 2. airr::read_airr_yaml(file, schema = RepertoireSchema) /work/lang/R/R/Interface.R:391:4
 3. airr::validate_airr_yaml_json(data = data, schema = schema) /work/lang/R/R/Interface.R:207:2
 4. base::sapply(...) /work/lang/R/R/Interface.R:273:2
 5. base::lapply(X = X, FUN = FUN, ...)
 6. airr:::FUN(X[[i]], ...)
 7. airr::validate_airr_entry(yaml_list[[X]], schema = schema) /work/lang/R/R/Interface.R:273:35
 8. airr::validate_airr_entry(definition_list[[f]][[n_array]], schema = reference_schemes[[n_ref]]) /work/lang/R/R/Interface.R:316:10
```

When I look at `Interface.R`, it seems to expect an array of references? I wasn't really sure how to modify because with `allOf` at the top-level of an object, all of the properties are combined together (at least how I changed `Schema.R`) instead of staying as references. Maybe you have an idea how to change?

javh (approved):
The R error is fixed. I'm not 100% sure it's _correct_, because I haven't walked through all the new code, but the tests are working now. Also, I copied the top level schema down assuming it was the correct one - I think it was only indentation mismatches.

javh:
Seeing this error in the schema consistency checks:
```
SampleProcessing object is different between V2 and V3 spec
{'discriminator': 'AIRR', 'allOf': {0: {'properties': {'sample_processing_id': {'x-airr': {delete: ['nullable']}, 'nullable': True}}}}}
```
I assume this is default of `allOf`?

schristley:
> Seeing this error in the schema consistency checks:
> 
> ```
> SampleProcessing object is different between V2 and V3 spec
> {'discriminator': 'AIRR', 'allOf': {0: {'properties': {'sample_processing_id': {'x-airr': {delete: ['nullable']}, 'nullable': True}}}}}
> ```
> 
> I assume this is default of `allOf`?

The check script may need to be updated. It attempts to modify the V2 into V3 to perform the check. It probably isn't handling the `allOf` properly.

javh:
Okay, I made a change to checks that resolves the issue. I'm not sure how generalizable it is, but everything passes.

Is this ready to merge?

schristley:
> Is this ready to merge?

There might be extra `print` statements in `interface.R` otherwise looks good.

javh:
Doh! Thanks. Fixed.

Issue #314, opened 2020-01-09
RearrangementReader only reads TSV - could we modify to include JSON? 


lfunderburk:
Have been using `airr.load_rearrangement` and `airr.load_repertoire` to load JSON and TSV results. 

There seem to be differences between the two modules - in particular the file format input each allows. 

We were able to successfully test that JSON response for repertoire metadata and flag missing fields. We want to do the same for JSON response for rearrangement data but noticed it only accepts TSV format using the `RearrangementReader` class.

Is airr-standards open to including loading and validating JSON format when dealing with rearrangements and if so, what is the best way to modify the code?  

Refer to python code 

https://github.com/airr-community/airr-standards/blob/master/lang/python/airr/io.py

for  `RearrangementReader`   class and

https://github.com/airr-community/airr-standards/blob/master/lang/python/airr/interface.py

for `load_repertoire` and `load_rearrangement` functions.

javh:
IIRC, this is something we had planned to do at some point, if/when the need arose.  So I don't think there would be an objections to adding this functionality (someone please chime in if this is an issue).

I suspect the easiest way to add support would be to copy-paste-modify the read/write code from `airr.load_repertoire`/`airr_write_repertoire` into `airr.load_rearrangement`/`airr.dump_rearrangement`, with an if-else block to perform the appropriate operation based on the input/output type. This would require adding another argument to these functions for the input/output file type. For example, `format="auto"` to autodetect from the file extension, format="tsv" to force TSV, and format="yaml" and format="json" for the other two.

That's my thoughts anyway. Pinging @schristley, as he may have more insight (he wrote the Repertoire I/O code).

We're pushing to v1.3 release, so if you were to put in a PR we'd have to push it back to a later release. I tagged it as v2.0, but it might not be that long. (We just don't have a schedule for v1.4 right now.)

bcorrie:
@javh was this addressed and can it be closed? Or if not, do we think it is important for v2.0?

javh:
There have been a lot of changes to how schemas are loaded in the packages since this was opened. I think we can let this one go.

PR #502, opened 2021-02-24
R package repertoire

### LINKED ISSUE #316: R package doesn't support Repertoire
The python package has support for loading/writing Repertoire objects as json/yaml, but the R package is missing this functionality.

This might be too late to get into v1.3, but it'd be nice if we could.

imkeller:
The R package can read and write and validate Repertoires now. The unit tests + vignette need to be further improved when making R and python packages consistent.
This should close #316.


schristley:
Thanks @imkeller ! I only browsed the code so far, but it got me thinking. Does R need its own `good_repertoire.airr.yaml` and `bad_repertoire.airr.yaml`? I'd really prefer not to create more copies, that's just more maintenance. I guess we need to find a way to share them between languages... The obvious place is under the top-level `tests` directory, and they are copied into the lang directories.

Hey @javh , the solution I'm thinking of requires being able to run a script on commit (to copy the files), or to copy them during the release and testing processes. You know of a way to automate that?

javh:
> Hey @javh , the solution I'm thinking of requires being able to run a script on commit (to copy the files), or to copy them during the release and testing processes. You know of a way to automate that?

Hrm. I'm not sure. For R, I don't think references files up the tree will work as a general solution, because everything needs to sit in the package directory. Maybe we can get GitHub actions to do some sort of auto-commit of the newer file to the other directories (not sure), but that wouldn't fix local testing. For both python and R, we could use a url for the test data source location. That has the same issue with local tests not working if you haven't pushed the change to the test data.

I'm kind of inclined to do the same thing we've been doing for the schema files and just have Travis check them for consistency. Maybe we can write a little prebuild script to check dates and sync files.

imkeller:
Can python handle test files outside the package directory? If yes, we could put the test files in the R package directory and make python read them from there. Not a very elegant solution though.

Otherwise I think a solution similar to the one for schema files would be good.

javh:
> Can python handle test files outside the package directory?

It can. We can use relative paths with `../` for the test data location... But, I'm not sure that we should.

schristley:
Does this code handle JSON format too?

imkeller:
> Does this code handle JSON format too?

No, I don't think so. But I can implement JSON reading/writing if this is needed. Is there a test file in JSON format somewhere?

schristley:
> No, I don't think so. But I can implement JSON reading/writing if this is needed. Is there a test file in JSON format somewhere?

No, I guess not. There are a couple of quick ways though, one is to use the python library to read in the yaml and write out the json like so:

```
import airr
data = airr.load_repertoire('input.airr.yaml')
airr.write_repertoire('input.airr.json', data['Repertoire'])
```

Another is to query the ADC, as it provides JSON, though that file is big and not formatted for easy viewing. Maybe restrict the data for one study or just one repertoire like here:

```
curl -k -o data.json --data '{}' https://vdjserver.org/airr/v1/repertoire

curl -k -o data.json --data '{"filters":{"op":"=","content":{"field":"study.study_id","value":"PRJNA593622"}}}' https://vdjserver.org/airr/v1/repertoire

curl -k -o data.json --data '{"filters":{"op":"=","content":{"field":"repertoire_id","value":"1329892364225474070-242ac113-0001-012"}}}' https://vdjserver.org/airr/v1/repertoire
```

If you have an iReceptor Gateway account, you can download a JSON file from the GUI.

ssnn-airr:
Thanks @imkeller for moving things forward.
A question for all. I have noticed `read_airr` has become `read_airr_tsv` and `read_airr_yaml`. Do we want to keep the original function `read_airr`, that takes as input the schema, and use the schema to do the right reading? Similar applies to validate_airr_tsv and validate_airr_yaml. `read_airr` is an exported function, so this change could will break things.

ssnn-airr:
Thanks @imkeller for moving things forward.
A question for all. I have noticed `read_airr` has become `read_airr_tsv` and `read_airr_yaml`. Do we want to keep the original function `read_airr`, that takes as input the schema, and use the schema to do the right reading? Similar applies to validate_airr_tsv and validate_airr_yaml. `read_airr` is an exported function, so this change will break things.

imkeller:
Thanks @ssnn-airr !

> Do we want to keep the original function `read_airr`, that takes as input the schema, and use the schema to do the right reading? 

Can you specify what you mean? The schema is read by `load_schema()`, I don't understand.

> `read_airr` is an exported function, so this change will break things.

If an external program uses `read_airr`, yes, it will break. But I thought that the functions the user should use are `read_rearrangement`, `read_repertoire` and `read_alignment`. These are the functions used in the vignette.

ssnn-airr:
> I thought that the functions the user should use 

Ah, users! :raised_back_of_hand:  What they should do, and what they end up doing :smile: I may or may have not been using `read_airr` because providing the same functionality, it makes more explicit that data is using the AIRR format...

ssnn-airr:
> Can you specify what you mean?

Only matters if we want to keep the `read_airr` function. Which is not for me to decide, and I am ok with start being a good user... although I like `read_airr` better that `read_rearrangement`, which is a long and ugly word to type. 

The definition is `read_airr <- function(file, base=c("1", "0"), schema=RearrangementSchema, ...) `. The function knows if `data` is a repertoire or a rearrangement, based on `schema`, and then inside it can use the right reading function (repertoire, rearrangement).

imkeller:
The original `read_airr` function has an argument base, which is not applicable to Repertoires. This is why I decided to have 2 functions: one for tsv (rearrangement) and one for yaml (repertorie and others to come).
It is unclear to me whether we also expect Repertoires in tsv format or Rerrangements in YAML format. I also wasn't aware of the JSON format (only Repertoires? or also rearrangements?).

ssnn-airr:
Maybe it would be useful to use this opportunity to make a systematic review of the R and python packages, and see what functions each of them have, what names and arguments they use, and try to harmonize the code and the tests. When I say systematic, I basically mean create a spreadsheet/document where we can compare both (R and python), and the tests and expectations. We can use this document to help decide what needs to be changed and where should we start.

imkeller:
Yes, this makes sense. Is this a general idea, or did you take a look at the new code and felt like there is a need for renaming and changing arguments?

@franasa and I have started to set up a R/python document to compare the unit tests and if possible I would like to change this in a separate pull request.

ssnn-airr:
It is something general. As this pull request includes function name changes, it seems a good time to do this. Is the document you mention available for me and @schristley @javh ?

javh:
Sorry I've been slow on the code review. Hoping to get to it this weekend. 

But, briefly, it seems like there might be a third option of keeping the existing `read_airr` interface, but abstracting the format in addition to the schema. For example, adding a `format` argument, and making private functions that are specific to the file type, where `format="auto"` would figure it out from the file extension (.json, .yaml, .tsv) and/or schema default, `format="json"` would read json files, etc.  This would be a way to address #314 as well, because you would decouple the format from the schema in the interface.

Personally, I don't like doing this sort of thing if the return object differs, but if only the input differs, then I think it's a bit cleaner.  (Analogous to class dispatching in R.)

imkeller:
> But, briefly, it seems like there might be a third option of keeping the existing `read_airr` interface, but abstracting the format in addition to the schema. For example, adding a `format` argument, and making private functions that are specific to the file type to do the reads, where `format="auto"` would figure it out from the file extension (.json, .yaml, .tsv) and/or schema default, `format="json"` would read json files, etc. This would be a way to address #314 as well, because you would decouple the format from the schema in the interface.

Yes, this sounds like a good idea. This is also what the python package does when reading JSON or YAML repertoire files.
The python package does not have a function similar to `read_airr`, it has all reading and writing functions specific for each type of data.

Do we expect all kind of data (repertoire, rearrangement, alignment, and others to come) in all three file formats (tsv, json, yaml)?

javh:
Yeah, it doesn't, but we could probably generalize `io.RearrangementReader` pretty easily by making the `schema` variable in `__init__` an argument,  then do an inheritance thing for `RearrangementReader`, `AlignmentReader`, etc, to maintain backwards compatibility.  We wrote the python library first, when we only had Rearrangement and Alignment - looks like we thought about abstracting it, but didn't actually do it.

The python library is setup to use iterators for I/O and the object oriented structure is very different. If there's some differences in how the R/python library work, then I think that's natural, even desirable. I think we should be as consistent as we can, but not at the expense of trying to make the R library "pythonic" or vice versa.

I don't know if we expect all schema to come in all formats. I'm not sure that a Repertoire tsv will work. But, if it's not extra work, then we could try to future proof - we don't have to enable everything for every format.

javh:
Sorry for the delay... I finally dug into this a bit more.  I have a general question, what do we want to do for the return object of `read_airr_yaml`?  Right now, it returns a length=1 list with "Repertoire" as the key with individual repertoires nested inside with numerical keys.

I'm fine with a named list. But, I'm inclined to de-nest this and have the return object be the list of Repertoire objects. And key (name) each repertoire list item by `repertoire_id`.

Thoughts?

ssnn-airr:
> de-nest this and have the return object be the list of Repertoire objects. And key (name) each repertoire list item by repertoire_id. Thoughts?

My first thoughts are. 1) It would look nicer. 2) If I recall correctly, and I could be wrong, the python package also uses this nesting. Do we want to keep the user experience similar in both packages, or not? 3) If we go the de-nest route, validation (check for uniqueness) of `repertoire_id` should be done always when loading a repertoire (even if we use a validate=false in the loading function). 4) Also, as `repertoire_id` is nullable, this should be handled. And sometimes the list would be named after `repertoire_id`, sometimes it would be just a number.

schristley:
Be sure to handle `repertoire_id` being blank for both reading and writing of repertoire files. There is [brief discussion](https://github.com/airr-community/airr-standards/tree/master/lang/python#reading-airr-repertoire-metadata-files) about this with the python library. That's why the python library doesn't automatically convert the list to a dictionary. I would be careful of any auto-assignment of `repertoire_id` unless the user explicitly asks for it.

javh:
From the call:
1. Leave the nesting of `Repertoire` to accommodate other top level metadata objects in the same file (eg, #478).
2. We should sync function names across python/R. Some discussion about whether the repertoire reading function should be referred to as a metadata reading function instead (not resolved, but we can worry about that later).

imkeller:
image

imkeller:
This table summarizes the relevant function names in python and R packages (old V1.3.0 vs. new V1.4.0). The colors indicate inconsistencies. Here is what I plan to do/ my questions:

- Red: I am going to name the R functions `validate_repertoire` and `validate_rearrangement`. The `validate_airr_tsv` and `validate_airr_yaml` will be kept as internal functions to be used by future public validation functions.
- Blue: I am not sure why there is the distinction between read an load in python. Should I rename the R functiond to "load"? This would also mean renaming the `read_rearrangement` function that was there is the V1.3.0.
- Green: I think here the inconsistency is in the python package. Will keep the R functions as they are. Should we rename to `write_rearrangement` in python or can we live with it?

ustervbo:
My 2c: In R I would keep 'read_*' since this is inline with the base read.*
functions.

javh:
Thanks, @imkeller.  Regarding `load`/`dump`/`create` in the python package, these are a consequence of having two different behaviors in the python package: iterative read/write and operating on data that's all in memory as pandas data.frames.

So `write` ends up being a method on the writer object generated by `create`:
```
writer = airr.create_rearrangement('output.tsv')
for row in reader:  writer.write(row)
```

It's been a long time, but I think that's why we went with `dump_rearrangement` over `write_rearrangement` - to avoid confusion with the `write` methods. However, I can see a good argument for renaming `dump_rearrangement` to `write_rearrangement`.

Even though `load_rearrangement` behaves more similarly to the R `read` function, I'm inclined to leave them as is both R and python, because the `read` function is the default behavior in both environments. And @ustervbo says, there's already an informal convention for what `read` means in R.

We could also consider merging `load_rearrangement` and `read_rearrangement` in python and have a flag that controls whether the function returns an iterator or a data frame. I don't know if that's worth the price of breaking backwards-compatibility. I suppose we could do that as a stealth change by adding the flag, making the default behavior of `read_rearrangement` unchanged, and keeping `load_rearrangement` as a wrapper. Which is basically what is happening right now anyway. But, that doesn't feel very pythonic to me.

ustervbo:
It occurs to me that we've discussed this before, and the conclusion was to
keep the list structure, because this matches the behaviour in python
library.

Best wishes,
Ulrik

imkeller:
Yes, I think I remember the same. So the list structure returned now is ok?
I have the impression that the repertoire.yaml and the repertoire.json files are not 100% comparable, because the json file that I obtained using curl and the above mentioned code does include an additional section "info". As far as I remember this is the advanced version of how airr files can look like.

schristley:
> Hi @schristley , I implemented the JSON reading and just want to make sure it is correct. So when I get the json file from one of the commands you suggested earlier:
> `curl -k -o data.json --data '{"filters":{"op":"=","content":{"field":"repertoire_id","value":"1329892364225474070-242ac113-0001-012"}}}' https://vdjserver.org/airr/v1/repertoire `
> ...
> I get a lot of missing AIRR mandatory fields (age_event, ancestry_population, ethnicity, template_quality, template_amount,...). As far as I see this is correct, so everything fine with my implementation. Is this ok?

Hi @imkeller , sorry I'm so late in responding! Yes, that is correct behavior, I forgot about that. You need to use the `include_fields` parameter; here is command which will "fill out" the repertoire object with all the fields in the AIRR schema.

```
curl -k -o data.json --data '{"filters":{"op":"=","content":{"field":"repertoire_id","value":"1329892364225474070-242ac113-0001-012"}},"include_fields":"airr-schema"}' https://vdjserver.org/airr/v1/repertoire
```

I guess you could also just directly request that repertoire, which automatically fills out the repertoire object.

```
curl -k -o data.json https://vdjserver.org/airr/v1/repertoire/1329892364225474070-242ac113-0001-012
```

schristley:
Hi @imkeller , there are conflicts in some files which need to be resolved. The conflicts in DESCRIPTION and NEWS are easy, but I'm not sure about Interface.R because the conflicts are extensive. Could you resolve them? Even though you can resolve them on the webpage here, you might want to do it locally so that you can test the result. Let me know if you have any questions when doing that.

imkeller:
@schristley I have merged master into r_package_repertoire and fixed the merge conflicts + some things in the tests. It looks like it's ok now.

imkeller:
@schristley @javh @ssnn-airr @ustervbo 
I am not sure if there are still conflicts when merging this branch. I have merged master into r_package_repertoire. Does this work or do I need to merge r_package_repertoire into master (I am not sure if I would have the permissions to do this)?

schristley (approved):

schristley:
> I am not sure if there are still conflicts when merging this branch. I have merged master into r_package_repertoire. Does this work or do I need to merge r_package_repertoire into master (I am not sure if I would have the permissions to do this)?

Nope, no conflicts. Github just gets confused sometimes; I've manually merged.

Issue #433, opened 2020-07-03
Specify types for custom columns in Rearrangements TSV?


scharch:
Ideally, I would be able to do this with a simple dictionary like { 'columnX':int, 'columnY':boolean }, but I could settle for passing in a full auxiliary schema if that's considered the proper way to go about it.

javh:
To you mean as arguments to read/write/validate in the python library?

scharch:
Yeah, so I can do something like `if row['columnX'] > 0`. Obviously I can just cast it explicitly, but it seems like it might be a useful feature generally.

schristley:
Is the field type sufficient, or would we see a circumstance where other attributes might be desirable? `nullable` is one example. If we don't really know but want to allow for the possibility in the future, we could just alter that simple dictionary slightly like this. Then we could support additional properties easily.

```
{ 'columnX': { type: int }, 'columnY': { type: boolean }}
```

scharch:
Yeah, I could see how that could be desirable. Flexible is good, anyway, and it doesn't seem to cost anything in this case.

bcorrie:
This would be a nice extension - to make sure I understand, this is for the validate libraries, correct? So you can say to the python validate functions, here is an AIRR TSV file with custom columns, and here is the schema for those custom columns, please validate the file against the AIRR schema and the custom schema... Correct?

schristley:
Yes, but also the read/write functions so it does automatic type conversion.

imkeller:
Could someone please share an example file for which there are custom columns or columns where the type cannot be guessed? I didn't find any in the test data.
The R package performs as expected for good_data.tsv and bad_data.tsv.
What is the role of extra_data.tsv? There is a parsing error, but I think it does not reflect the column type problem.

scharch:
Leaving this open as the functionality has only been implemented in R, not Python

bcorrie:
@scharch still v2.0 issue?

scharch:
I mean, it would be nice for the python and R functionalities to match...

javh:
This was implemented via the `aux_types`  argument to `read_trabular`.

scharch:
@javh I believe this should be reopened as there is not yet a corresponding functionality in the python `RearrangmentReader`

javh:
Okiee. Reopened. I missed that this was a python library issue and not an R library issue (for no good reason, just not paying attention).

PR #505, opened 2021-03-10
Issue 475 and 433

### LINKED ISSUE #475: read_rearrangement in the R package is vulnerable to confusion about column types from readr::read_tsv
I have a file that is giving the following warning from `airr:read_rearrangement`, which is from `readr::read_tsv`:

```
Warning: 84610 parsing failures.
row    col             expected               actual
7570   d_call_10x      1/0/T/F/TRUE/FALSE     IGHD1-1
7591   vjl_group       1/0/T/F/TRUE/FALSE     G21 
```
Leading to all `NA` values in those columns.  Using `guess_max=10000` with `readr::read_tsv` avoids the error.
Those aren't AIRR columns, but we should still be loading them honestly somehow. Maybe by adding a `guess_max` argument to pass through to `read_tsv` or by defaulting unrecognized fields to type `character`.

ustervbo:
Add argument to `airr::read_airr()` to define the type of non-schema columns, with character as default.

Fitting unit test is added. Test data created on the fly from toy_data.tsv

Closes #475 and #433

### DIFF
--- a/lang/R/DESCRIPTION
+++ b/lang/R/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: airr
 Type: Package
-Version: 1.4.0-dev
+Version: 1.4.0
 Date: 2021-03-21
 Authors@R: c(person("Jason", "Vander Heiden", role=c("aut", "cre"), 
                     email="jason.vanderheiden@gmail.com"),
--- a/lang/R/DESCRIPTION
+++ b/lang/R/DESCRIPTION
@@ -1,13 +1,15 @@
 Package: airr
 Type: Package
 Version: 1.3.0
-Date: 2020-05-26
+Date: 2021-03-21
 Authors@R: c(person("Jason", "Vander Heiden", role=c("aut", "cre"), 
                     email="jason.vanderheiden@gmail.com"),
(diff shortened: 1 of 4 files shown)


scharch:
For #433 this should be done for the python library, as well. I might have some time next week...

javh:
Thanks, @ustervbo!  I made some minor change: shorted the argument name, extra docs, added the change to the news, and stuck you in the author list.

Can you confirm you want to be in the package author list and confirm the email address I put is correct, please?

javh (approved):

schristley:
It looks like these types are very R-specific. Should we be more generic and use OpenAPI types? I like the idea in #433 of passing in an OpenAPI field structure. Is there any reason not to do that?

javh:
> It looks like these types are very R-specific. Should we be more generic and use OpenAPI types? I like the idea in #433 of passing in an OpenAPI field structure. Is there any reason not to do that?

Meaning:
string -> character
number -> double
integer -> integer
boolean -> logical

javh:
From the call:
* Make the type name change from @schristley and merge in.

PR #506, opened 2021-03-15
Germline fields merge

bcorrie:


bcorrie:
Hmm, @schristley you are correct in #438. It is odd though, as both the Network view here (https://github.com/airr-community/airr-standards/network) and my desktop app (Sourcetree) show merge from master into germline-fields but not the merge of germline-fields into master. 

Any insights as to why that would be? 

That is why I created this pull request - I thought it had not been merged yet...

schristley:
maybe the type of commit? #438 says it was a merge commit. That's actually a commit type that we have turned off, so my guess is that you did the merge into master locally and pushed it. I know I occasionally have to do that because there are conflicts that need to be manually resolved. Or not the commit type but due to the local merge, and GitHub didn't recognize it for some reason? Just guesses, not that familiar with the different git merges.

bcorrie:
OK, I am going to close this pull request...

PR #523, opened 2021-05-04
Update rearrangement support for scirpy

grst:
Since v0.7, scirpy can write AIRR data back to the rearrangement format. It can therefore also serve as converter from 10x VDJ, Tracer, and Bracer to AIRR format.


schristley:
thanks for the PR @grst !

PR #529, opened 2021-06-03
AIRR Germline schema

schristley:
initial drop in of germline schema

some tasks

- [ ] determine appropriate normal/de-normal representation
- [ ] review `nullable` for each field
- [ ] review field names
- [ ] hook schema into documentation
- [ ] ADC API
- [ ] "fix" germline set references in `Repertoire`


schristley:
replaced by #530

Issue #518, opened 2021-03-29
characterize the custom schema attributes in OGRDB


schristley:
Decide which can/should be made part of `x-airr`.

williamdlees:
In 517 you are asking which objects in the OGRDB schema are relevant to a germline gene reference set. Here I think you must be asking what other objects in the OGRDB schema should go into the z-airr schema. I don't think there are any.

schristley:
Not objects per se, but additional attributes on the objects and/or properties. From a quick skim of the [schema](https://github.com/airr-community/gldb-wg/blob/master/germline_schema/schema/receptor_germline_schema.yaml), I see `occurrence`, `rows`, `width`. Maybe they are not needed and can be removed. The `type` of many of the properties will need to be cleaned up too.

williamdlees:
I thought I'd rmoved all the row and width attributes: they are gone now. The preamble (discriminator, type, occurrence) was modelled on the AIRR schema at the time this one was first put together, it can be updated to whatever is used now. The types, I think, are best updated when we complete a review of the content of the schema, because they contain useful information that will need to be documented elsewhere, should we choose to keep the attributes.

schristley:
Hi @williamdlees , created this issue in case there were OGRDB specific attributes that you might want in the schema as optional attributes. I think you indicated there probably won't need to be any? If you still believe so then we can close this issue.

williamdlees:
Thanks, I think we should close this now.

Issue #483, opened 2020-11-18
Document germline capabilities more clearly


bcorrie:
Need to capture the current state and what is expected of users of the standard in this regard - see pull request #438 for the spec changes that need to be better documented.

scharch:
done

Issue #482, opened 2020-11-18
Next steps for Germline developments


bcorrie:
Further to the discussion in #416 and the pull request in #438 where we created the first germline implementation, I wanted to capture a new issue with the outstanding "next steps" as discussed here: https://github.com/airr-community/airr-standards/pull/438#issuecomment-716825784. 

The list from the comment above includes:

- Finalize GeneDescription
- Finalize how we will normalize germline gene sets
- Discuss if we want/need repositories that store germline gene sets
- Come up with an API to query germline gene sets

bcorrie:
Related information:

- Using CURIEs for germline annotation: https://github.com/airr-community/airr-standards/issues/416#issuecomment-639611583
- Germline set pervious work: #157 
- GermlineSet object (previously added and then removed) that we want to develop further: https://github.com/airr-community/airr-standards/pull/438/commits/ef4bb0bbcaac9f1e91768ea9ae2d76addb4afb67

williamdlees:
Closing this following the pull request and merge

PR #530, opened 2021-06-10
Germline

### LINKED ISSUE #517: characterize the schema objects in OGRDB
Determine which objects are "core" to an interoperability standard for germline gene reference sets.

williamdlees:
closes #517 #518 #483 #157 #482 

Addition of GermlineSet and GermlineDescription definitions. Revision of genotype definitions and some renaming to conform to fir with general usage of terms: GermlineSet for a set of alleles found in a species or strain, and Genotype for the list found in an individual. This request supercedes https://github.com/airr-community/airr-standards/pull/529 : apologies, I did not notice this request earlier.

some tasks:

- [x] determine appropriate normal/de-normal representation
- [x] review `nullable` for each field
- [x] review field names
- [x] hook schema into documentation
- [x] ADC API
- [x] "fix" germline set references in `Repertoire`
- [ ] Acknowledgement? Is it time to create a `Person` object to use throughout the schema? Link to JSON-LD?


bcorrie:
I had a quick look at this... The definition of `ReceptorGenotype` and `MHCGenotype` look reasonable - you have added more detail, but fundamentally they work the same as the former `GermlineClassSet` and `MHCClassSet`.

The way these are referred to in the `Subject` and `DataProcessing` objects no longer work the same though. You have essentially redefined `GermlineSet` to be a completely new type of object (which already existed) but are referencing it in `Subject` and `DataProcessing` in the same way as it was used originally. I don't think this works. 

In addition, the `ReceptorGenotype` and `MHCGenotype` objects are actually not referred to at all in the spec. This definitely doesn't work 8-)

There used to be an array of `GermlineSet` objects which were light weight container objects that referred to an external ID for a "fully defined germline set" (which is what you refer to as a GermlineSet) and each one had a locus and a list of alleles (the old `GermlineClassSet` . So for a subject you essentially had one `GermlineSet` per locus for which you had germline information (a `GermlineClassSet`. I think we still need this structure, with the ability to refer to an external object that is your `GermlineSet`.

Basically I think what happened is you redefined `GermlineSet` but didn't replace the functionality of being able to store `ReceptorGenotype` with subjects.

schristley:
> I had a quick look at this... The definition of `ReceptorGenotype` and `MHCGenotype` look reasonable - you have added more detail, but fundamentally they work the same as the former `GermlineClassSet` and `MHCClassSet`.

Actually, this is an important point, they aren't fundamentally the same. With this schema work, GLDB is making a clear distinction between `GermlineSet` and `Genotype`. While a `Genotype` can be complex (look at [definition](https://github.com/airr-community/ogrdb/blob/42a046ce9892961c8fc31d666b4c1a574f981ad6/schema/inferred_gene_submission_schema.yaml#L604) in OGRDB), the GLDB group decided to initially hold off on expanding `Genotype` in the AIRR schema for now, and just define a basic `Genotype` object. Both have commonality though in that both have a "list of genes". I've changed `Subject` to reference the new objects, but there is still more work to be done on the definitions.

`data_processing_germline` is different in that I think this is a `GermlineSet` though it isn't clear to me how it matches up with `germline_database` which is also a `GermlineSet`. I've removed the `MHCGermlineSet` as we aren't defining it.

schristley:
@williamdlees One comment I had regards the statement, "All genes in a GermlineSet should be from a single locus." This has the implication that a `DataProcessing` can only process a single locus, because there is only a single `GermlineSet` object attached to `DataProcessing`. This makes it unclear how to curate 10X single cell studies where CellRanger annotates all loci in a single processing step.

One solution is to allow multiple `GermlineSet` objects to be referenced in `DataProcessing`, need to think some more how this might work.

Another solution is to re-introduce `GermlineClassSet` which is all the genes for a single locus, while `GermlineSet` is a set of `GermlineClassSet` objects which covers all loci for the species. This adds complexity that I'm not sure is needed.

Ultimately, there needs to be a simple way to know exactly which genes were used for an annotation. I'll note that even the current `DataProcessing` is ambiguous in this regards. For example, `germline_database` may indicate "VDJServer X.X.X" but that's the full set and doesn't indicate that (say) only the TRB genes were used.

schristley:
I've enabled this branch on RTD so can see the documentation changes, it's available [here](https://docs.airr-community.org/en/germline/index.html)

williamdlees:
Re. the use of genotype with Subject - I think this is problematic, because the genotype inferred from one repertoire could disagree with the genotype inferred from another, for example if some alleles are expressed in one repertoire but are undetectable in another. I've moved the field to Repertoire and also turned it into a list, so that we can acommodate genotypes inferred from multiple loci - and, potentially, genotypes of the same locus derived by different methods.

williamdlees:
Re. the use of multiple GermlineSets in a DataProcessing - I think this is best achieved by making a list in DataProcessing, which I have added. Together with the requirement that a GermlineSet is confined to a single Locus, I think this aalso addresses the point of potentially not knowing which subset of genes was used. The field germline_database seems redundant to me - do we need to keep it?

williamdlees:
@bcorrie  @schristley  I'd be in favour of scrapping MHCGenotype for the time being - unless you feel that you're ready to start importing existing haplotyping information derived from other (non repertoire) sources. If not, let's postpone defining the field until we're ready to use it.

schristley:
> @bcorrie Re. the use of genotype with Subject - I think this is problematic, because the genotype inferred from one repertoire could disagree with the genotype inferred from another, for example if some alleles are expressed in one repertoire but are undetectable in another. I've moved the field to Repertoire and also turned it into a list, so that we can acommodate genotypes inferred from multiple loci - and, potentially, genotypes of the same locus derived by different methods.

@williamdlees Hmm, no, I don't think your change is correct. A Subject only has one (true) genotype right? That's what suppose to be stored in the Subject. I think you are assuming that genotype is just for that one repertoire, and that those genotypes can be different in the Subject from one repertoire to the next, but that's invalid. If there are 5 repertoires for the same subject, that Subject object MUST be identical across all of those repertoires, otherwise the whole Repertoire metadata structure fails. That Subject is a singular object, it's just duplicated across those repertoires.

We also don't want it in Repertoire, that is essentially sticking "analysis output" in the Repertoire object, which is something we don't want to do. The situation you are referring to isn't directly supported by the AIRR schema, those genotypes are output from a data processing run, and are not directly accessible. If we wanted to do that, we would need to formalize the genotype just like we have for other analysis output (clones, trees, etc.) and make it a first class object with it's own file format and ADC API query end point, but I don't think we want to do that now. Right now that genotype is just some extra data attached to the Subject.

schristley:
> @schristley Re. the use of multiple GermlineSets in a DataProcessing - I think this is best achieved by making a list in DataProcessing, which I have added. Together with the requirement that a GermlineSet is confined to a single Locus, I think this aalso addresses the point of potentially not knowing which subset of genes was used. The field germline_database seems redundant to me - do we need to keep it?

@williamdlees Yeah after thinking about it more, I agree that multiple GermlineSets in DataProcessing seems the way to go. Note that the current DataProcessing object is "broken", there is an ongoing redesign that will resolve numerous issues with the current structure. Unfortunately, yes, `germline_database` is a MiAIRR field, but it was always an open text field without a specific format so I think it will just become a simple display name/label versus an identifier.

williamdlees:
> @williamdlees Hmm, no, I don't think your change is correct. A Subject only has one (true) genotype right? That's what suppose to be stored in the Subject. I think you are assuming that genotype is just for that one repertoire, and that those genotypes can be different in the Subject from one repertoire to the next, but that's invalid. If there are 5 repertoires for the same subject, that Subject object MUST be identical across all of those repertoires, otherwise the whole Repertoire metadata structure fails. That Subject is a singular object, it's just duplicated across those repertoires.
> 
> We also don't want it in Repertoire, that is essentially sticking "analysis output" in the Repertoire object, which is something we don't want to do. The situation you are referring to isn't directly supported by the AIRR schema, those genotypes are output from a data processing run, and are not directly accessible. If we wanted to do that, we would need to formalize the genotype just like we have for other analysis output (clones, trees, etc.) and make it a first class object with it's own file format and ADC API query end point, but I don't think we want to do that now. Right now that genotype is just some extra data attached to the Subject.

Unfortunately we do see cases of variation in inferred genotype between repertoires from the same individual. In VDJbase, where there are several repertoires for an individual, we nominate one representative to be used in cases where we only want one from the subject. We could follow that approach here, but I think it's important to record which repertoire a genotype is derived from, and as far as I can see that's tricky to denormalise if it is stored in Subject. Is there a solution to that?

schristley:
> Unfortunately we do see cases of variation in inferred genotype between repertoires from the same individual. In VDJbase, where there are several repertoires for an individual, we nominate one representative to be used in cases where we only want one from the subject. We could follow that approach here, but I think it's important to record which repertoire a genotype is derived from, and as far as I can see that's tricky to denormalise if it is stored in Subject. Is there a solution to that?

It shouldn't be problem adding a field/array to Subject that points to repertoires. Those repertoires wouldn't be denormalized though as it could be a circular reference, but it can contain identifiers. For example:

```
inferred_from:
    type: array
    description: List of repertoire identifiers used to infer this genotype
    example:
        - rep123
        - rep456
    items:
        type: string
```

Even so, I see one crinkle in that a Repertoire can have multiple DataProcessing. Imagine the user ran two different DataProcessing workflows to generate the genotype (e.g. VDJbase and IgDiscover) then picked one. The above info doesn't say which DataProcessing was used. I'm gonna delay on dealing with that though until we make some progress with the upcoming DataProcessing sprint.

bcorrie:
> @bcorrie @schristley I'd be in favour of scrapping MHCGenotype for the time being - unless you feel that you're ready to start importing existing haplotyping information derived from other (non repertoire) sources. If not, let's postpone defining the field until we're ready to use it.

@williamdlees I am not sure about this... I believe there are cases when a study may have MHC information about a subject, and this provides a mechanism to store that in a repository. So I don't think it is premature to go down this path. @schristley you have some studies on VDJServer that have HLA for some subjects, and currently you store that as a custom field in the repertoire, no? Is this not where you would like to be able to store that information?

bcorrie:
> > @bcorrie Re. the use of genotype with Subject - I think this is problematic, because the genotype inferred from one repertoire could disagree with the genotype inferred from another, for example if some alleles are expressed in one repertoire but are undetectable in another. I've moved the field to Repertoire and also turned it into a list, so that we can acommodate genotypes inferred from multiple loci - and, potentially, genotypes of the same locus derived by different methods.
> 
> @williamdlees Hmm, no, I don't think your change is correct. A Subject only has one (true) genotype right? That's what suppose to be stored in the Subject. I think you are assuming that genotype is just for that one repertoire, and that those genotypes can be different in the Subject from one repertoire to the next, but that's invalid. If there are 5 repertoires for the same subject, that Subject object MUST be identical across all of those repertoires, otherwise the whole Repertoire metadata structure fails. That Subject is a singular object, it's just duplicated across those repertoires.
> 
> We also don't want it in Repertoire, that is essentially sticking "analysis output" in the Repertoire object, which is something we don't want to do. The situation you are referring to isn't directly supported by the AIRR schema, those genotypes are output from a data processing run, and are not directly accessible. If we wanted to do that, we would need to formalize the genotype just like we have for other analysis output (clones, trees, etc.) and make it a first class object with it's own file format and ADC API query end point, but I don't think we want to do that now. Right now that genotype is just some extra data attached to the Subject.

I think I agree with @schristley here. What we are trying to store in this case is **the** Genotype of the Subject (as best we know it). In this case it is "metadata" about the subject that is "known". It might be acquired through sequencing, it might be acquired through inference, but it is metadata. So I think this belongs in the `Subject` object in the spec.

Now it is true that one might generate lots of different genotypes for a single subject, with different techniques and at different times, but those are analyses. I don't think we want to store the output of analyses in our metadata objects. As @schristley says, if we want to store such analyses in the ADC, we should treat them like our other analysis entities (`Rearrangement`, `Clone`, `Cell`, `Tree`) including having an API for querying them.

I think we should think carefully about the use cases for that - as my understanding is that the goal for this from the AIRR perspective is to store the Genotype for a Subject as AIRR metadata. We then want to be able to link things in that Genotype to entities stored elsewhere (e.g. OGRDB) to get full definitions for those entities.

schristley:
> @schristley you have some studies on VDJServer that have HLA for some subjects, and currently you store that as a custom field in the repertoire, no? Is this not where you would like to be able to store that information?

That's a good point, I had forgotten about that study. Yes, it would be good to have a simple way to store that information. I guess the question becomes how deep down that rabbit hole we want to go. I know just enough of the HLA community to know that they have their own naming issues, as in the common names one sees only represent part of the gene, and there are more detailed proposals that cover the whole gene, etc etc. So I see a list of names as useful, I'm not sure if the other fields are needed.

williamdlees:
> 
> 
> > Unfortunately we do see cases of variation in inferred genotype between repertoires from the same individual. In VDJbase, where there are several repertoires for an individual, we nominate one representative to be used in cases where we only want one from the subject. We could follow that approach here, but I think it's important to record which repertoire a genotype is derived from, and as far as I can see that's tricky to denormalise if it is stored in Subject. Is there a solution to that?
> 
> It shouldn't be problem adding a field/array to Subject that points to repertoires. Those repertoires wouldn't be denormalized though as it could be a circular reference, but it can contain identifiers. For example:
> 
> ```
> inferred_from:
>     type: array
>     description: List of repertoire identifiers used to infer this genotype
>     example:
>         - rep123
>         - rep456
>     items:
>         type: string
> ```
> 
> Even so, I see one crinkle in that a Repertoire can have multiple DataProcessing. Imagine the user ran two different DataProcessing workflows to generate the genotype (e.g. VDJbase and IgDiscover) then picked one. The above info doesn't say which DataProcessing was used. I'm gonna delay on dealing with that though until we make some progress with the upcoming DataProcessing sprint.

I think it's worth going back to this point. The genotype is determined from an annotation, uising a selected workflow. So it really needs to be a first-class queryable object, I think, referring back to a Rearrangement, and inclluding some of its own DataProcessing. Given the arguments above, that says to me that we should stop beating around the bush and make Genotype a first-class queryable object. We could still pick a 'representative' Genotype and link to it from Subject - perhaps, in the same way, you might intend to link to a 'representative' earrangement from each repertoire? Otherwise I am not sure you simple queries such as 'does repertoire x contain a CDR3 with this sequence' will be resolved.

bcorrie:
I think the top down use case we are trying to support is to help someone find data where subjects have known gene's in their genotype. For example, give me all of the subjects that have this V gene in their receptor genotype. Give me all subjects that have this HLA gene in their MHC genotype. Combine that with gender specific, disease specific queries, etc., and you have a powerful mechanism for looking for genotype specific features in the AIRR-seq data. But this is metadata about the subject, and the capability that we are adding to the AIRR data model is to allow researchers to capture and store that information in their study metadata if they so choose. To me this is part of the curation process for a study and having genotype information as part of the Subject object means that this has been curated as part of the study by the study curator/data steward.

Being able to store genotypes as the output of an analysis/annotation step and treating that output as an annotation data set that is related to, and produced from, a specific set of annotations, is a different and separate issue. This is substantially different than what we are talking about above and implies as @williamdlees says that they need to be treated as an annotation output that is stored in each ADC repository (I of course look at this from the ADC perspective - but the same applies to files on disk). 

We can choose to do one, the other, or both. The original solution with genotype in Subject addressed the first issue.

schristley:
I'm not opposed to going first-class object with Genotype, but this is a complicated task. In one of the earlier GLDB meetings, it was discussed and the general consensus (as I remember) was not to tackle at this time. Though the OGRDB [Genotype](https://github.com/airr-community/ogrdb/blob/42a046ce9892961c8fc31d666b4c1a574f981ad6/schema/inferred_gene_submission_schema.yaml#L604) is quite extensive, we likely need to design that in the larger context of the haplotype design... Plus how that relates with [HapMap](https://www.genome.gov/10001688/international-hapmap-project) and [1000genomes](https://www.internationalgenome.org/data) open variation data.

Anyways, that Genotype work should be decoupled from this PR so we're not solving two big problems at the same time. So is it reasonable that we go with the simpler change now, i.e. a single representative Genotype in Subject, and have a separate issue/PR for dealing with Genotype more extensively?

williamdlees:
Scott, 

I don’t think there would be GLDB objection to making Genotype a first class object, and I think most of us would argue for making progress on genotypes without waiting for a larger context. But it’s not a primary motivation for me personally right now, I would  rather make progress with germline sets. 

Really, I don’t understand why putting one genotype in Subject is ok, but putting multiple genotypes there breaks a fundamental principle. There’s a good argument for putting no genotypes in Subject, which you and Brian have made and I have tried to respond to, but once that is broken I don’t see that the number matters. Why don’t we allow multiple genotypes to be in Subject - with a reference to the repertoire and technology with which they were derived - and if ADC wish to restrict themselves to just one, that’s an implementation decision they can make. 

William

bcorrie:
> Scott, I don’t think there would be GLDB objection to making Genotype a first class object, and I think most of us would argue for making progress on genotypes without waiting for a larger context. But it’s not a primary motivation for me personally right now, I would rather make progress with germline sets. Really, I don’t understand why putting one genotype in Subject is ok, but putting multiple genotypes there breaks a fundamental principle. There’s a good argument for putting no genotypes in Subject, which you and Brian have made and I have tried to respond to, but once that is broken I don’t see that the number matters. Why don’t we allow multiple genotypes to be in Subject - with a reference to the repertoire and technology with which they were derived - and if ADC wish to restrict themselves to just one, that’s an implementation decision they can make. William

I think the difference is conceptual - and goes towards what MiAIRR is (of which `Subject` is an object). The MiAIRR fields in the `Repertoire` object are designed to capture and describe metadata about a study - not the output of analysis. If a study gathers information and knows something about the genotype of a subject, the genotype field in the `Subject` object is the place to store that. It is exactly what it says it is, `Subject.genotype` stores the subject genotype metadata as documented by the curator of the study metadata. It might be "measured" with genomic sequencing or "inferred" with repertoire sequencing, but it is something that is part of the study metadata that the curator who is the data steward for that study in that specific repository is responsible for.

This is very different conceptually than a genotype that is a computational output of an analysis step, carried out by researcher X starting from Repertoire R for subject S. As you say, there might be many of these, using many algorithms, and many parameters, resulting in mana different genotypes. I don't think it makes sense to store those as part of the `Subject` metadata - mainly because they aren't 8-) They aren't metadata about the study, they are data produced by an analysis step by a researcher for a specific purpose...

Note that if we want to store the output of such analyses as object in the AIRR Data Commons, we can. Then can become objects like `Rearrangement`, `Clone`, and `Cell`. Each can be associated back to a subject, or even a sample (if the genotype is inferred from a specific sample), just like `Rearrangement`, `Clone`, and `Cell` are. And you can have many Genotypes, processed by many different algorithms, stored for the same `Sample` if you like.

But I agree with @schristley - this is a very different thing than storing a genotype with a `Subject` as metadata and I don't think we should confuse the two... I think is a mistake is for us to be changing the genotype field of the `Subject` object, which is metadata about the study itself, based on analyses that are performed by researchers. If you are the curator/data steward for the study in a repository (e.g VDJBase), and you re-compute genotype base on a new germline, then sure, `Subject.genotype` might change in that repository - but that is an action of the data steward for VDJBase.

We certainly can/should consider adding the ability to store genotypes - but lets not confuse that with providing the ability of a data curator to store genotype about a subject if it is known as part of their study metadata.

bcorrie:
> . There’s a good argument for putting no genotypes in Subject, which you and Brian have made and I have tried to respond to, but once that is broken I don’t see that the number matters

I have not, at least intentionally, argued that there should be no genotypes in subject. If I did or was interpreted that way, that was a mistake 8-)

It seems to me that if the correct biological concept is that `Subjects` have a genotype then there should be a mechanism in the AIRR Standard to store a subject's genotype if it is known. For me, that is really all the Germline extension that we are talking about now is about... 8-)

williamdlees:
Brian,

I could hardly disagree with that. But we don’t have the genotype, we have analyses that suggest the genotype. What do we do with them?

bcorrie:
> Brian, I could hardly disagree with that. But we don’t have the genotype, we have analyses that suggest the genotype. What do we do with them?

If I processed this as part of my curation of the study, I would store the genotype as inferred with the flag `subject.genotype.inference_process = repertoire_sequencing` 8-) Like with most MiAIRR study metadata fields, the curator needs to provide enough information to the user so that they know whether they can use/trust the data. Some people might not trust repertoire_sequencing for genotype inference, others may not trust genomic_sequencing.

The main thing for me is that the genotype that is stored in `Subject` is the one that the curator of the study metadata in the repository thinks is "correct|best|closest" to being "the" genotype for that subject. In @ctwatson case, this would be what, in my understanding, would be the sequencing he does per subject (with `inference_process = genomic_sequencing`). For VDJBase, you would presumably use `inference_process = repertoire_sequencing`. In our case, if a paper reported having subject genotype data, either immune receptor or MHC, we would store what the paper said the genotype for each subject was, indicating how the study authors generated this data. All of this is part of the study description through the study metadata.

This is what others use to find data that is of specific interest to them, using the curated study based genotype data. For example, give me all the rheumatic heart disease patients with `IGHV4-61*02` in their genotype, then give me all that don't have it, and do a comparison...

Note if we stored the genotypes from many analyses (using our genotype analysis output storage scenario), you could also ask the question in which data from subjects with the same disease did some genotype inference analysis tool find `IGHV4-61*02` - which might return different results than when searching the genotype data curated with the study. Finally, you could search for `IGHV4-61*02` in the rearrangements, and use that to infer that the allele was in the subject's genotype because an annotation tool found a sequence that was close to it (this is the only thing you can do today!).

All are useful questions and slightly different ways of looking at usage of `IGHV4-61*02` in the data in the ADC.

bcorrie:
> Thanks Brian, It doesn’t sound as though we are that far apart in terms of concept. I do however have an issue with introducing the "correct|best|closest" field in Subject *before* we introduce “our genotype analysis output storage scenario” because it would hide the information that was available to make the curational decision, which, to me, goes against the principle of reproducibility, and because it will cause problems later when we want to go back and collect the information to populate the genotype analysis output storage scenario when it is introduced.

Warning - a bit of a rant...

For me, having a field that stores some metadata (in this case subject genotype) is a necessary first step. We want to store something as a starting point so we can move forward - this was agreed months ago, with the acknowledgement that what we define can be expanded to include better documentation of curator decisions and pointers to germline sets from which the genotype was extracted. But that doesn't change the fact that we need a subject genotype field. We seem to be back to not being able to add subject genotype until we have a Germline Spec and the ability to define, store, and look up Germline Sets. 

Both are important, but they are complimentary and quite different. We don't need both to move forward...

>Is this not just a matter of tabbing the definition of Genotype to the left a bit?

I think you need to elaborate a bit on that - not sure what you mean.

>Forgive me but I think I could have implemented the query in VDJbase in the time we have spent discussing it! 

Yes, and we could have had a subject genotype field in the AIRR Spec two years ago if we didn't wait to ensure we had a germline spec and way to store germline sets and genotype analysis output first. Apologies for trying to push `subject.germline` through to get something in the standard without germline and germline set specs and implementation, but it has been two years... 8-)

BTW, in terms of implementation - we have the `subject.germline` part of the schema implemented in our repositories and APIs and it is in beta testing. The `subject.germline` part of the spec was added to the master branch eight months ago after *extensive* (https://github.com/airr-community/airr-standards/pull/438) discussion. It hasn't really undergone much change since then. We have based our implementation on this part of the spec. 

https://github.com/airr-community/airr-standards/blob/f48591e282f52e8943b6c8fe7570a22eb537b589/specs/airr-schema.yaml#L708

The germline branch was created on May 18 (six weeks ago). The suggestion on this branch as it currently stands is there is no ability at all to store `subject.germline` at all. I believe this is a problem... If nothing else, I suggest we put `subject.germline` back in the spec as it was pre May 18 and discuss how it should be changed before changing functionality. 

Again, I believe the need to store `subject.germline` is for the most part independent of whether we decide to store genotype analysis objects and/or Germline Sets. They should be linked, sure, but they do not rely on each other for existence...

bcorrie:
I think a good summary of the distinctive nature of the two different problems being discussed here is as follows:

- We have the `subject.germline` merge that happened in November, this has been completed with much discussion: https://github.com/airr-community/airr-standards/pull/438
- We have the germline and germline set additions which this pull request is about, closing this issue: https://github.com/airr-community/airr-standards/issues/482 (and others).

When I closed #438 I new there was more work to do on the germline spec and germline definitions. 

Unfortunately, working on this pull request to add a germline spec and germline sets has broken the `subject.germline` part of the spec (removed it in fact). It makes sense for the definition of germline and germline sets to change the `subject.germline` definition, but not remove it 8-)

williamdlees:
I’m sorry Brian, I have no wish to impact your implementation and didn’t realise you were that well ahead. Also accept that this request has become too big and drifted away from its real purpose. The changes are big anyway, because of the need to separate germline set and genotype, which none of us anticipated 8 months ago. 

 

I’ve put genotype back into Subject. Please have a look and let me know what you think.

 

William

bcorrie:
> I’m sorry Brian, I have no wish to impact your implementation and didn’t realise you were that well ahead. Also accept that this request has become too big and drifted away from its real purpose. The changes are big anyway, because of the need to separate germline set and genotype, which none of us anticipated 8 months ago. I’ve put genotype back into Subject. Please have a look and let me know what you think. 

No worries - sorry for the rant 8-) We have been around and around on this topic so often and we want/need to move forward with some sort of initial implementation...

I will have a look at the spec... 

FYI, here is where we are at using the `subject.germline` spec on a toy data set:

An ADC API query that says give me all of the repertoires whose subject have `IGHV1-1*01` in their genotype...
```
curl -d '{"filters": {"op":"=","content": {"field":"subject.germline.receptor_germline.germline_class_list.germline_alleles","value":"IGHV1-1*01"}},"fields":["subject.germline","subject.subject_id"]}' http://turnkey-test2.ireceptor.org/airr/v1/repertoire
```
An ADC API query that says give me all of the repertoires that contain HLA-C in their genotype.
```
curl -d '{"filters": {"op":"contains","content": {"field":"subject.germline.mhc_germline.germline_class_list.germline_alleles","value":"HLA-C"}},"fields":["subject.germline","subject.subject_id"]}' http://turnkey-test2.ireceptor.org/airr/v1/repertoire
```
We are close to being able to put this into production. Need to test queries on the Gateway, but the API is working so that should be a simple step. 

Of course we don't have any data with germline per subject yet.

Note that this is my experimental repository, so I add and remove data all the time. There is some toy germline data loaded now (so the queries work) but it may disappear the next time I need to use the repository for testing some other feature 8-)

williamdlees:
@bcorrie @schristley just wondering how you felt about this now - do you think we might be able to move it forward? Thanks.

schristley:
@williamdlees I'm still catching up after my vacation but I'll review this in more detail this week.

bcorrie:
@williamdlees I think that looks good. Compared to what we had before, we would populate the `gene_symbol` fields only. Once we had the repository/ID stuff figured out, we could then go back to populate the other fields `coding_sequence_identifier`, `germline_set_ref`.

bcorrie:
One clarification in terms of use... Given that there is an `undocumented_alleles` and `deleted_genes`, when you infer a genotype, what would you expect in these fields. Lets say the annotation was done with the "standard" IMGT reference set. I am assuming you would expect:

- documented_alleles would contain the alleles from the reference set that were inferred that this subject indeed had. 
- undocumented_alleles would contain any alleles that were inferred for this subject, but NOT in the reference set used

What does `deleted_genes` contain? Is it:

- The list of all alleles in the reference set that this subject's genotype does not contain?
- The list of alleles that the original annotation made an allele call for and the genotype inference determined were incorrect
- Or some other subtlety that I am missing??? 8-)

williamdlees:
>     * `documented_alleles` would contain the alleles from the reference set that were inferred that this subject indeed had. 
>     * `undocumented_alleles` would contain any alleles that were inferred for this subject, but NOT in the reference set used

Yes exactly. The documented alleles could be sourced from more than one reference set if necessary.

> What does `deleted_genes` contain? Is it:
> 
>     * The list of all alleles in the reference set that this subject's genotype does not contain?

Actually a list of *genes* for which no alleles were called (if haplotyping is possible, the chromosomal phasing can be inferred, for example by using a program like Rabhit. Otherwise just the genes for which no alleles were called).

williamdlees:
> 
> 
> @williamdlees I think that looks good. Compared to what we had before, we would populate the `gene_symbol` fields only. Once we had the repository/ID stuff figured out, we could then go back to populate the other fields `coding_sequence_identifier`, `germline_set_ref`.

Thanks Brian. Yes that would be good, though I suggest we imntroduce a convention for IMGT reference sets as soon as possible, so that you can include the date it was downloaded.

bussec:
> Actually a list of genes for which no alleles were called (if haplotyping is possible, the chromosomal phasing can be inferred, for example by using a program like Rabhit. Otherwise just the genes for which no alleles were called).

@williamdlees I have to admit that I am struggling to understand the use case for this field. If a gene is not on the list, it is not on the list, isn't that enough? Or is this again a problem with parsing gene names out of an allele list?

williamdlees:
If we take the case of homozygous deletions, If people are interested in searching for repertoires with specific gene deletions or looking at population-level statistics, which I think they might be, it wouldn’t be practical to service a query by running through all the repertoires on the fly, looking for gaps. Even more the case for heterozygous deletions, where an inference tool has to scan the annotates sequences in the repertoire.

bcorrie:
> > @williamdlees I think that looks good. Compared to what we had before, we would populate the `gene_symbol` fields only. Once we had the repository/ID stuff figured out, we could then go back to populate the other fields `coding_sequence_identifier`, `germline_set_ref`.
> 
> Thanks Brian. Yes that would be good, though I suggest we imntroduce a convention for IMGT reference sets as soon as possible, so that you can include the date it was downloaded.

Would it make sense to use something like this: 

```
"germline_set_ref":"IMGT:202118-1"
```
To coincide to a download identifier from here: http://www.imgt.org/download/V-QUEST/archives/

This almost makes it a CURIE in that you can use the CURIE to resolve to a URI:

http://www.imgt.org/download/V-QUEST/archives/IMGT_vquest_release202118-1.zip

This is not exactly how we use CURIEs, but close... @bussec any thoughts?

schristley:
@williamdlees I've added nullable tags to all of the fields that didn't have them. I mostly relied upon `required` but made a few guesses here and there. Also brought the V2 and V3 specs inline.

williamdlees:
Thanks Scott. I’ve made a few updates (to both versions), the fields look good to me now.

bussec:
> If we take the case of homozygous deletions, If people are interested in searching for repertoires with specific gene deletions or looking at population-level statistics, which I think they might be, it wouldn’t be practical to service a query by running through all the repertoires on the fly, looking for gaps.

@williamdlees Ok, I agree that this is good use case. The question that remains is whether `deleted_genes` contains **all** genes of the reference set that do not have an entry in `documented_alleles` or only **some** of them:

* In the former case (**all**), there is exactly one possible set for `deleted_genes` (given input data and reference set). Therefore I would be wondering whether this is something that could better be generated by the repository, as it serves mainly search optimization.
* In the latter case (**some**), the field would become an annotation generated by the original data provider to point out specific genes of interest. While I agree that this can be meaningful, it would lead to a situation in which `deleted_genes` is an unreliable source from the user perspective (the provider might not have been interested in the gene).

Which of these two options did you have in mind?

bussec:
> This is not exactly how we use CURIEs, but close... @bussec any thoughts?

While we are used to IRIs also acting as URLs, this is not a prerequisite. With the split of the `CURIEResolution` section into `CURIEMap` and `InformationProvider` (in #524), when could handle this (in this case the `.zip` suffix, which would not be supported by normal CURIE prefix expansion).

There is however another issue that is harder to solve, because it  requires cooperation from the `germline_set_ref` provider: In the absence of a clear commitment of the provider to **not** break links, the IRI in your example is effectively a URL, i.e., if the link breaks, we right now would not have any way to ensure that a new location with a similar name still holds the same file. There are some contingency measures that we could take (e.g., keeping/publishing the checksums of the files), so this should not be a show stopper to move ahead with this. But ultimately this needs to be solved on the provider side (which would be simple, they could e.g. put copies onto Zenodo and get DOIs for them).

The final point is whether about the specific prefix you proposed. As IMGT has more resource than VQUEST, we should check which ones might be of potential use and thus require prefixes on their own.

williamdlees:
To me it is not simply a list of all the genes in the reference set that were not observed. it is an annotation generated by whoever processes the repertoire from sequences to annotation (which might not necessarily be the original data provider,  VDJServer does this, as does VDJbase). With all that implies.

 

There are two use cases in particular that highlight this for me:

 

1.	The original sequences might not include all genes, for example the experimentalist might have decided only to amplify a subset with a restricted set of primers. There are cases of this I believe.
2.	The person doing the annotation may have been able to use an anchor J-gene to determine a haplotype (VDJbase does this wherever possible). In this case we clearly have an annotation rather than something that a downstream database can work out for itself.

bussec:
Thanks for clarifying, this all makes sense. I guess we can add the "complement of `documented_alleles` via the API at a later point.

bussec:
Two more points on consistency in naming objects/properties:
* The `gene_symbol` property expects a gene designation in `deleted_genes`, but *seems* (my interpretation) to 
expect the allele designation in `GeneDescription`, `ReceptorGenotype.documented_alleles`, `MHCGenotype.germline_alleles`. This would be different specifications within the schema as well as being at variance with the definition of a "gene symbol" by HGNC and IMNC (which does not include the allele).
* The `ReceptorGenotype*` objects refer to loci, not to receptors. A receptor -- both biologically and based on our `Receptor` object -- is a protein complex with (typically) two distinct components/chains. Users may expect that `receptor_id` and `receptor_genotype_id` refer to more closely related object than they actually do.

williamdlees:
Thanks Christian. Would it be better, do you think, to change gene_symbol to allele_symbol in those contexts, and to change ReceptorGenotype to ReceptorLocusGenotype?

schristley:
> * The `ReceptorGenotype*` objects refer to loci, not to receptors. A receptor -- both biologically and based on our `Receptor` object -- is a protein complex with (typically) two distinct components/chains. Users may expect that `receptor_id` and `receptor_genotype_id` refer to more closely related object than they actually do.

I think we can just call it `Genotype`?

williamdlees:
I thought we would want to distinguish it from the MHCGenotype

bussec:
@williamdlees I would have to check but IMO the term "allele symbol" is typically not used. Do you want to keep is distinct from the `allele_name` in `undocumented_alleles` as the latter is not a canonical designation?

williamdlees:
Let’s make it allele_name, then. Sorry.

williamdlees:
We should probably add the species designation to the IMGT IRI, as the file contains all species.

bcorrie:
@williamdlees @schristley I notice that William changed `Subject` to have a `ReceptorGenotypeSet` (which is looking good), but I noticed the `DataProcessing` object still has a  `data_processing_germline_sets`.

Is that what we want? My original concept for this was to enable the ability to store the set of alleles that were used in the annotation, essentially a list much like `documented_alleles` from `ReceptorGenotype`. That way it would be possible to curate the actual alleles from multiple germline sets that were used in an annotation.

The way it is now, we are storing an entire `GermlineSet` with the full `GeneDescription`, which I don't think is what we want? `DataProcessing` objects shouldn't be storing `GeneDescriptions` should they?

schristley:
> Is that what we want?

My suggestion is that we "punt" on DataProcessing in this PR. As DataProcessing is being reworked, we can add in the appropriate linkages later (e.g. DataProcessing PR that goes across the whole schema)

schristley:
> I thought we would want to distinguish it from the MHCGenotype

Well the word `Genotype` is different from `MHCGenotype` ;-D

Because these schemas are within the `AIRR` context (the `discriminator`), we've gone with simple names like `Repertoire` vs `AdaptiveImmuneReceptorRepertoire`, `Study` vs `AIRRSeqStudy` and `Receptor` vs something longer (even though these are a small subset of all biological receptors).

But I'm fine with a longer name if `Genotype` isn't specific enough within an AIRR context.

bcorrie:
> My suggestion is that we "punt" on DataProcessing in this PR. As DataProcessing is being reworked, we can add in the appropriate linkages later (e.g. DataProcessing PR that goes across the whole schema)

Can I remove it from DataProcessing for now? 8-)

williamdlees:
Scott, that works for me.

 

William

bcorrie:
@schristley I notice that you added a `repertoire_id` field to the `Genotype` object:

https://github.com/airr-community/airr-standards/blob/b989c7d27dc08c0350e63d25e6355efdd512f5fc/specs/airr-schema.yaml#L931

Is this something we want/need. The Genotype object is associated with a subject, so it doesn't make sense to link it to a single Repertoire does it?

I would think linking the Genotype data that was generated from genotype inference from a specific Repertoire would be handled by the `DataProcessing` object - which we are yet to define...

schristley:
> @schristley I notice that you added a `repertoire_id` field to the `Genotype` object:
> 
> https://github.com/airr-community/airr-standards/blob/b989c7d27dc08c0350e63d25e6355efdd512f5fc/specs/airr-schema.yaml#L931
> 
> Is this something we want/need. The Genotype object is associated with a subject, so it doesn't make sense to link it to a single Repertoire does it?
> 
> I would think linking the Genotype data that was generated from genotype inference from a specific Repertoire would be handled by the `DataProcessing` object - which we are yet to define...

Yes, @williamdlees wanted to know which repertoire was used to generate each genotype, under the idea that a different genotype might be produced from each repertoire. I suppose it might just point to a DataProcessing, but that might be ambiguous. So, Genotype is similar to Clone/Rearrangement in that it points to both its Repertoire and to its DataProcessing.

bcorrie:
> > @schristley I notice that you added a `repertoire_id` field to the `Genotype` object:
> > https://github.com/airr-community/airr-standards/blob/b989c7d27dc08c0350e63d25e6355efdd512f5fc/specs/airr-schema.yaml#L931
> > 
> > Is this something we want/need. The Genotype object is associated with a subject, so it doesn't make sense to link it to a single Repertoire does it?
> > I would think linking the Genotype data that was generated from genotype inference from a specific Repertoire would be handled by the `DataProcessing` object - which we are yet to define...
> 
> Yes, @williamdlees wanted to know which repertoire was used to generate each genotype, under the idea that a different genotype might be produced from each repertoire. I suppose it might just point to a DataProcessing, but that might be ambiguous. So, Genotype is similar to Clone/Rearrangement in that it points to both its Repertoire and to its DataProcessing.

I am not sure that makes sense though. Again, I think we are confusing the biological concept of a `Subject` having a `Genotype` (which is what the field in the `Subject` captures) versus how a `Subject`'s genotype is derived (which might be from inferences from one or more repertoires through "data processing"). 

The `Genotype` object is specifically designed to capture `defined_alleles` from multiple `GermlineSets` which implicitly implies that they are possibly inferred from different `DataProcessings`. In addition, it also seems likely that a `Subjects` `Genotype` would often be inferred from different `Repertoires` if a subject has multiple samples from different tissues etc. 

I think what we want is for a `DataProcessing` to link a `Repertoire` to a `Genotype` through the `DataProcessing` pointing to the `receptor_genotype_id`. This accomplishes the same link (a `Genotype` is produce from a `Repertoire`) but removes the 1-1 relationship as well as helps us track the provenance of the tool that was used to produce the `Genotype`.

Bottom line is that I don't think it makes sense for a `subject.genotype.receptor_genotype_set` to be derived from a single `Repertoire`.

williamdlees:
To me it feels dangerous to have a singleton subject.genotype.receptor_genotype_set, given that, today, we don’t have a method for determining what it might be. Perhaps we’ll reach a point in the future where there is an accepted method, and conflicts in results can be treated as errors, but we’re a long way from that right now. If a genotype set in that position would, by its nature, have to be a singleton, I think it would be better not to have the field at all, because there is no valid use of the field at the moment, it will only lead to confusion. On the other hand, it wouldn’t seem unreasonable to me to offer a ‘best approximation’ in this position, if we wanted to. Ultimately, that’s surely what the other subject fields offer: we seldom know absolute truth. But in that case, given the current state of the art, I think it would be important to give an indication of how the best approximation was arrived at, and that ‘s where we seem to get into representational issues.

 

Being able to link an inferred genotype to a repertoire is a useful thing to do. This is what we do in VDJbase today, and if we’re to start offering an AIRR-standards-compliant query, we need a way to represent that. We don’t need a ‘best approximation’ field, but could populate it if it existed.

bcorrie:
> To me it feels dangerous to have a singleton subject.genotype.receptor_genotype_set, given that, today, we don’t have a method for determining what it might be. Perhaps we’ll reach a point in the future where there is an accepted method, and conflicts in results can be treated as errors, but we’re a long way from that right now. If a genotype set in that position would, by its nature, have to be a singleton, I think it would be better not to have the field at all, because there is no valid use of the field at the moment, it will only lead to confusion. On the other hand, it wouldn’t seem unreasonable to me to offer a ‘best approximation’ in this position, if we wanted to. Ultimately, that’s surely what the other subject fields offer: we seldom know absolute truth. But in that case, given the current state of the art, I think it would be important to give an indication of how the best approximation was arrived at, and that ‘s where we seem to get into representational issues. Being able to link an inferred genotype to a repertoire is a useful thing to do. This is what we do in VDJbase today, and if we’re to start offering an AIRR-standards-compliant query, we need a way to represent that. We don’t need a ‘best approximation’ field, but could populate it if it existed. 

I am no expert by a long stretch, but isn't this a feature of the technique being used to derive the genotype? The thing we are trying to accomplish with this pull request is the ability to represent the `Subject`'s genotype? It is in fact "metadata" about the `Subject`. Like most metadata, it captures what the data curator believes to be accurate for the 

It is my understanding that when one produces a genotype using genomic sequencing one is capturing the `Subject`'s genotype (it is my understanding that this is what @ctwatson does). This has nothing to do with repertoires and we need to be able to capture this. 

When one infers genotype using repertoire sequencing, you are estimating the subjects genotype based on a repertoire. Having the ability to use an inferred genotype for a subject from repertoire sequencing is what this is trying to capture. In many cases, if you have one repertoire per subject (one time point, one tissue, etc), this inferred genotype is a good proxy for the subject's genotype. The fact that it is marked as inferred can help the researcher decide how they want to treat that (whether they trust the inference process used). That is why I suggest we capture the link between `Genotype` and `Repertoire` with a `DataProcessing`. If a genotype is inferred, the researcher is going to want to know more than just which repertoire the inference came from, they are going to want to know which `DataProcessing` steps were used to infer the genotype. 

It sounds to me like your concern is mostly about having multiple genotype inferences for a single reprertoire

bcorrie:
@schristley says:
> So, Genotype is similar to Clone/Rearrangement in that it points to both its Repertoire and to its DataProcessing.

Storing Genotype as an analysis output like Clone/Rearrangement is very different from setting the metadata for a `Subject`'s genotype. That is not what this pull request is talking about... We are going in circles 8-)

https://github.com/airr-community/airr-standards/pull/530#issuecomment-868838028

schristley:
> It is my understanding that when one produces a genotype using genomic sequencing one is capturing the `Subject`'s genotype (it is my understanding that this is what @ctwatson does). This has nothing to do with repertoires and we need to be able to capture this.

This is out-of-scope for this PR. This PR is fundamentally about GermlineSets and GeneDescriptions. The Genotype is a complicated [topic](https://github.com/airr-community/airr-standards/pull/530#issuecomment-859082637), and I don't think we should be dragging down this PR with trying to perfect genotypes. We are only getting into it because what's in the AIRR schema on `master` is [wrong](https://github.com/airr-community/airr-standards/blob/d5857a47bee7e250281bdd70d12b8e1c3a08c127/specs/airr-schema-openapi3.yaml#L886) from a terminology perspective, it's not a germline for a subject, it's a genotype. That needs to be cleaned up, and we cannot just leave the old stuff as is because it conflicts.

As far as I can see, the Genotype object has everything in it from before, maybe there is some extra stuff, but it's optional, ignore it if it doesn't apply to you. Then repeat the mantra three times, "this is only an experimental draft", and realize there will be adjustments. If that's not good enough, then modify until you are satisfied. But again, I think any serious genotype work needs to be it's own PR.

schristley:
> @schristley says:
> 
> > So, Genotype is similar to Clone/Rearrangement in that it points to both its Repertoire and to its DataProcessing.
> 
> Storing Genotype as an analysis output like Clone/Rearrangement is very different from setting the metadata for a `Subject`'s genotype. That is not what this pull request is talking about... We are going in circles 8-)

We're not. I'm trying to mediate between the two viewpoints. I understand your viewpoint as wanting to store a genotype in the subject as metadata, with the details about how that genotype got decided as a fuzzy blob off in DataProcessing land or wherever, blah blah I don't care, I just need to store it so I can query upon it. The other viewpoint is thinking about the practical aspects of generating genotypes, and wanting to record those genotypes in a well-defined format, with at least some preliminary provenance regarding the process. That will surely be refined later.

Both viewpoints are valid. Is there some fundamental reason why the same `Genotype` object cannot be used for both cases??? At least for this PR which isn't suppose to be about genotypes anyways...

bcorrie:
@bcorrie says:
> It is my understanding that when one produces a genotype using genomic sequencing one is capturing the Subject's genotype (it is my understanding that this is what @ctwatson does). This has nothing to do with repertoires and we need to be able to capture this.

@schristley says:
> This is out-of-scope for this PR. This PR is fundamentally about GermlineSets and GeneDescriptions. 

OK, this has gotten horribly MESSY...  And we are going around in circles. https://github.com/airr-community/airr-standards/pull/530#issuecomment-869923375

My main point is that if this PR isn't about `Subject` germline/genotype, then it can't break `Subject` germline/genotype and it shouldn't be changing it.

>The Genotype is a complicated [topic](https://github.com/airr-community/airr-standards/pull/530#issuecomment-859082637), and I don't think we should be dragging down this PR with trying to perfect genotypes. We are only getting into it because what's in the AIRR schema on `master` is [wrong](https://github.com/airr-community/airr-standards/blob/d5857a47bee7e250281bdd70d12b8e1c3a08c127/specs/airr-schema-openapi3.yaml#L886) from a terminology perspective, it's not a germline for a subject, it's a genotype. That needs to be cleaned up, and we cannot just leave the old stuff as is because it conflicts.

`Subject` level "Germline" (as it was called at the time) was approved and merged into Master months ago. It was at the `Subject` level and stored a list of Germline alleles that were metadata about a `Subject` - that is, they were the germline alleles, either acquired through genomic sequencing or from inference from repertoire sequencing, that were associated with a `Subject`.  See #438.

Changes to the fields added by #438 were included in this PR because it was related to being able to store and reference germline sets. Rather than store a simple gene name as a string, we are better served by storing entities from `GermlineSets`. Unfortunately, as mentioned in https://github.com/airr-community/airr-standards/pull/530#issuecomment-869923375 this pull request and the GermlineSet changes broke that capability, which we have now fixed. There is now a Genotype object (rather than Germline) associated with a `Subject` and we can now store a more complete level of information (reference objects in a GermlineSet) about `Subject` genotype. So this PR is changing `Subject` genotype and it is in scope. Maybe it shouldn't be, but it currently is because those changes were made.

So if this PR isn't about `Subject` genotype, then it CAN NOT break the `Subject` germline/genotype from #438... 8-) If germline is the wrong name for the field, then changing the field name is trivial.

It appears to me that the scope creep came in because it was decided that to fix the naming conflict for `GermlineSet` and `GeneDescription` that `GermlineSet` should be added to `Subject` genotype. 

So we either need to revert back to `Subject` having a list of allele names as it did in #438 (possibly with a field name change from germline to genotype) and not refer to GermlineSets at all as part of `Subject` genotype

Or

Solve this problem through this PR.

I agree that we can't solve the ultimate Genotype definition through this PR, but we also can't let this PR break being able to store at least a basic `Subject` genotype as defined in #438.

bcorrie:
> > @schristley says:
> > > So, Genotype is similar to Clone/Rearrangement in that it points to both its Repertoire and to its DataProcessing.
> > 
> > 
> > Storing Genotype as an analysis output like Clone/Rearrangement is very different from setting the metadata for a `Subject`'s genotype. That is not what this pull request is talking about... We are going in circles 8-)
> 
> We're not. 

Yeah, I think we are 8-)

>I'm trying to mediate between the two viewpoints. I understand your viewpoint as wanting to store a genotype in the subject as metadata, with the details about how that genotype got decided as a fuzzy blob off in DataProcessing land or wherever, blah blah I don't care, I just need to store it so I can query upon it. The other viewpoint is thinking about the practical aspects of generating genotypes, and wanting to record those genotypes in a well-defined format, with at least some preliminary provenance regarding the process. That will surely be refined later.
> 
> Both viewpoints are valid. Is there some fundamental reason why the same `Genotype` object cannot be used for both cases??? At least for this PR which isn't suppose to be about genotypes anyways...

Sure, as long as this PR doesn't break `Subject` genotype...

Which is where this all started... Having a `Genotype` object being associated with a `repertoire_id` doesn't make sense in the general sense, in particular at the `Subject` level - which is why I raised the question in the first place...

schristley:
> OK, this has gotten horribly MESSY... And we are going around in circles. [#530 (comment)](https://github.com/airr-community/airr-standards/pull/530#issuecomment-869923375)

Well I don't know what you want then. I'm unable to parse all of your comments into actionable changes, so you need to make the changes that are going to satisfy your requirements, and then we can review. My primary concern is the name change, as my understanding is it's the "genotype" to be stored in Subject and not the "germline set used to analyze the subject's repertoires".

bcorrie:
> > OK, this has gotten horribly MESSY... And we are going around in circles. [#530 (comment)](https://github.com/airr-community/airr-standards/pull/530#issuecomment-869923375)
> 
> Well I don't know what you want then. I'm unable to parse all of your comments into actionable changes, so you need to make the changes that are going to satisfy your requirements, and then we can review. My primary concern is the name change, as my understanding is it's the "genotype" to be stored in Subject and not the "germline set used to analyze the subject's repertoires".

I don't think it makes sense to have a `Genotype` object refer to a single `repertoire_id`. My original question is simple - and is listed here: https://github.com/airr-community/airr-standards/pull/530#issuecomment-892226364

williamdlees:
Brian,

 

Let’s suppose there’s a technique that gives a near-perfect fully haplotyped genomic sequence of the locus. There’s a lot of structural variation across the population – regions that are copied and regions that are deleted – so it isn’t straightforward to map the near-perfect sequence to positions of known genes and alleles. It’s also non-trivial to parse out the functional genes from the sequence – we don’t have a comprehensive understanding of the motifs that make a gene functional or non-functional. For humans we have a reasonable understanding of what genes are there and broadly what variation exists, but even for humans, the best genotype we come up with for a subject today could be different to one we come up with in a  few years’ time, as we understand what variation is possible, what gene duplications are seen, sort out the naming and so on. For other species the position is much worse.

 

I’ve been thinking about this a bit more and I think it would be reasonable to have either genomically- or repertoire- derived results in the subject.genotype.receptor_genotype_set singleton, provided that there’s some statement of the data used to derive it. As long as the assembly or the repertoire is referenced, someone who cares can go to the reference and find some details, attempt to reproduce the result and so on. But without that reference, in either the genomic or the repertoire case, I think it’s an issue. For example someone might decide to publish genotypes derived from a less-than-perfect genomic sequence, 1000 genome data for example. At least if the reference is there, a user can see what it is, and make up their minds whether or not to trust it. If the source isn’t made clear, the user is dependent on the judgement of the curator and can’t attempt  to reproduce.

 

You could say that the same issue exists with any subject metadata, even age or sex for example. But those examples are a lot more straightforward, and there are generally accepted standards. We’re not there yet with genotypes. Maybe this is an argument for not putting the genotype in the subject metadata at all, but then we come back to practicality. It’s useful to have a genotype listed there. And the DataProcessing schema isn’t sorted out yet. So while it would be ideal to have full details of methods and so on, maybe we could just settle on a singleton with reference to the repertoire or genomic sequences used to derive it?

schristley:
> > > OK, this has gotten horribly MESSY... And we are going around in circles. [#530 (comment)](https://github.com/airr-community/airr-standards/pull/530#issuecomment-869923375)
> > 
> > 
> > Well I don't know what you want then. I'm unable to parse all of your comments into actionable changes, so you need to make the changes that are going to satisfy your requirements, and then we can review. My primary concern is the name change, as my understanding is it's the "genotype" to be stored in Subject and not the "germline set used to analyze the subject's repertoires".
> 
> I don't think it makes sense to have a `Genotype` object refer to a single `repertoire_id`. My original question is simple - and is listed here: [#530 (comment)](https://github.com/airr-community/airr-standards/pull/530#issuecomment-892226364)

And I thought I [answered that question](https://github.com/airr-community/airr-standards/pull/530#issuecomment-892235294), so if you're not satisfied then make the changes necessary so that "it makes sense" to you, and we can review.

bcorrie:
I thought it was worth discussing before I removed it, but I have gone ahead and made the change as you suggested...

bcorrie:
I think we should leave the information about how a Genotype was produced to be described through the DataProcessing object - since that is what is happening when Genotypes are inferred. They are the outcome of applying some processing to a set of Rearrangements.

We can take this use case into account in our DataProcessing sprint...

williamdlees:
I’m fine with this, as long as we are ok with the principle that it will be applied to the subject metadata in due course/

bcorrie:
> I’m fine with this, as long as we are ok with the principle that it will be applied to the subject metadata in due course.

I definitely agree that we want to know how an inferred genotype was computed for the `Subject` genotype. I would think something along the lines of having a `DataProcessing` object that contains a link to the source repertoire (through `repertoire_id`) and the output object, in this case the `Genotype` (through `receptor_genotype_set_id`). 

So if you want to know how the `subject.genotype` was produces you get its `receptor_genotype_set_id` and search for a `DataProcessing` that has such an ID.

Or alternately, if the DataProcessing doesn't track the output entities, we could have a `subject.genotype` point to the `data_processing_id` that was used to produce it.

Note that this is more general than having `subject.genotype` pointing to a repertoire_id because a DataProcessing could take many repertoires as input. So assuming `DataProcessing` is general enough to do this, this should work well???

schristley:
Not sure if an extension API file is the right way to go, but if we add the end points to the core ADC API file then repositories will be required to implement. I put them under the same `/airr/v1` base url, but we can consider making it different.

bcorrie:
> Not sure if an extension API file is the right way to go, but if we add the end points to the core ADC API file then repositories will be required to implement. I put them under the same `/airr/v1` base url, but we can consider making it different. I'm thinking of the three scenarios:
> 
> * a repository only supports the germline end points
> * a repository only supports the core ADC API end points
> * a repository supports both

I think I would lean to be different end points. No reason a repository can't implement both, but separating them from /airr/v1 seems to make sense.

bcorrie:
Hey, check this out:

curl --data '{"fields":["study.study_title","study.study_id","sample.sample_id","subject.subject_id","subject.genotype"]}' http://turnkey-test2.ireceptor.org/airr/v1/repertoire

schristley:
> I think I would lean to be different end points. No reason a repository can't implement both, but separating them from /airr/v1 seems to make sense.

@bcorrie I thought you'd appreciate the symmetry :-D  Here's all the end points with the current proposal:

```
/airr/v1/repertoire
/airr/v1/rearrangement
/airr/v1/clone
/airr/v1/germline
/airr/v1/gene
```

if you want a different base url, then these would be the end points where we have to decide what `SOMETHING` is

```
/airr/v1/repertoire
/airr/v1/rearrangement
/airr/v1/clone
/airr/SOMETHING/v1/germline
/airr/SOMETHING/v1/gene
```

I prefer the current as adding `SOMETHING` doesn't feel like its adding anything ;-D

Thinking longer term, presumably we are going to add Cell, Tree, Receptor and maybe other end points (any top-level AIRR object I guess). Should they be extension end points? Under a different base url? And looking back, what about Clone, should we have added it to the core ADC API, and thus essentially requiring all repositories to implement, why isn't it an extension as well?

bcorrie:
> > I think I would lean to be different end points. No reason a repository can't implement both, but separating them from /airr/v1 seems to make sense.
> 
> @bcorrie I thought you'd appreciate the symmetry :-D Here's all the end points with the current proposal:

Good point 8-)  

I think my main concern is the growing set of API entry points that are "required" to be ADC compliant and those that are optional. And I also wonder about mixing entry points that serve different purposes.

Maybe one question to answer is does the /airr part of the API mean it is an AIRR Community API or does it mean that the API returns AIRR-seq and related data? I was thinking of it as AIRR-seq data, but maybe that is wrong?

bcorrie:
> Thinking longer term, presumably we are going to add Cell, Tree, Receptor

There is already a /airr/v1/cell

schristley:
> I think my main concern is the growing set of API entry points that are "required" to be ADC compliant and those that are optional. And I also wonder about mixing entry points that serve different purposes.

We have two (or four depending on how you count) required end points. These are published and documented. Everything else is draft/experimental, but it would be good to decide on guidelines, for example, if you only intend to host bulk repertoire data why need `/cell`?

```
/repertoire
/rearrangement
```

schristley:
> Maybe one question to answer is does the /airr part of the API mean it is an AIRR Community API or does it mean that the API returns AIRR-seq and related data? I was thinking of it as AIRR-seq data, but maybe that is wrong?

The AIRR Community API, in particular V1, `/airr/v1`. It is our "namespace". One thing we can do when adding the new endpoints is to bump up the version, `/airr/v2`. Then that creates a clear differentiation between the current published and documented, and all the new stuff.

bcorrie:
> We have two (or four depending on how you count) required end points. These are published and documented. Everything else is draft/experimental, but it would be good to decide on guidelines, for example, if you only intend to host bulk repertoire data why need `/cell`?

I think I would expect an API for an object in the AIRR spec to be required/implemented but return nothing if you don't have any Cell data... No much of a burden on the API implementer but gives consistency of use from the client perspective.

I think of experimental/draft VS approved API entry points as being different that optional VS required entry points. Experimental/draft entry points will either become approved official entry points or be dropped. Of all the entry points (draft/experimental or approved) I would expect many (most) would be required but some would be optional. 

Essentially the two classes of API are independent.

- An API is either required or optional
- An API is either draft/experimental or approved

Does that make sense? 8-)

Using the above argument /airr/v1/germline makes sense, and then the question becomes is it optional or required?

schristley:
> Essentially the two classes of API are independent.
> 
> * An API is either required or optional
> * An API is either draft/experimental or approved
> 
> Does that make sense? 8-)

I think that makes sense. In my mind then, `/airr/v1` is required whiled `/airr/async/v1` is optional, because supporting async requires a significant investment by the repository.

> Using the above argument /airr/v1/germline makes sense, and then the question becomes is it optional or required?

I would then argue that all of the "data" end points under `/airr/v1` be required, because as you say:

> I think I would expect an API for an object in the AIRR spec to be required/implemented but return nothing if you don't have any Cell data... No much of a burden on the API implementer but gives consistency of use from the client perspective.

That makes it easy, and we don't have to get into categorizing the different end points, then explaining why one is optional versus required, and etc. They just return empty results if they don't have that data.

bcorrie:
I added this topic to be discussed at the next CRWG

schristley:
@bcorrie So the result of the CRWG call was that we leave `/airr/v1` as it was published, and put all the new endpoints in `/airr/v2`?

bcorrie:
Yes, I believe so... Although we are developing these (/clone and /cell) on /airr/v1 at the moment so we will have to change that...

schristley (approved):

schristley:
@williamdlees I think this is ready to merge, the API changes will be put all together in a separate PR.

williamdlees:
Scott, I’m sorry, I’ve been reviewing the germline stuff with Christian and Corey, and there’s a change in there that isn’t ready.

 

We would like to change the name of the GeneDescription object, because it’s not really describing a gene. At the moment in the germline branch it’s called Sequence, but I think this gives us issues in the larger schema because there are other kinds of sequences, for example in repertoires. We haven’t concluded on a name as yet.


William

williamdlees:
Given that this part of the schema will be a work in progress for a while, we could change it back to GeneDescription for this release and fix it later. Don’t want to hold you up, and would very much like to get the germline branch merged.

 

From: ***@***.*** ***@***.***> 
Sent: 06 September 2021 12:16
To: 'airr-community/airr-standards' ***@***.***>; 'airr-community/airr-standards' ***@***.***>
Cc: 'Mention' ***@***.***>; ***@***.***' ***@***.***>; 'Watson,Corey T' ***@***.***>
Subject: RE: [airr-community/airr-standards] Germline (#530)

 

Scott, I’m sorry, I’ve been reviewing the germline stuff with Christian and Corey, and there’s a change in there that isn’t ready.

 

We would like to change the name of the GeneDescription object, because it’s not really describing a gene. At the moment in the germline branch it’s called Sequence, but I think this gives us issues in the larger schema because there are other kinds of sequences, for example in repertoires. We haven’t concluded on a name as yet.


William

schristley:
> Scott, I’m sorry, I’ve been reviewing the germline stuff with Christian and Corey, and there’s a change in there that isn’t ready. We would like to change the name of the GeneDescription object, because it’s not really describing a gene. At the moment in the germline branch it’s called Sequence, but I think this gives us issues in the larger schema because there are other kinds of sequences, for example in repertoires. We haven’t concluded on a name as yet.

Hmm, ok, it seemed to me that it described many annotations for a germline gene. Is it an allele description instead? I'm assuming we need something in the schema that says "Gene" or "Allele" because those are concepts that researchers would expect. I agree that the name Sequence by itself is probably too ambiguous.

> Given that this part of the schema will be a work in progress for a while, we could change it back to GeneDescription for this release and fix it later. Don’t want to hold you up, and would very much like to get the germline branch merged.

We can do that if you think the name won't be quickly resolved.

williamdlees:
I've reviewed the naming again with Christian and Corey and I think we are good to go now. In particular we have renamed GeneDescription to AlleleDescription, which is a closer fit, and still specific enough not to cause problems eleswhere.

bcorrie:
@schristley `repertoire_id` was removed purposefully based on discussions here: https://github.com/airr-community/airr-standards/pull/530#issuecomment-892992427

I see it has been added back in on your "put back accidentally deleted stuff" commit above. Is that a mistake? If not we should discuss...

schristley:
FYI @williamdlees , be careful with your commits. I think you might have copied an old file for the schema instead of using the latest from github because it deleted out some other changes brought in from the master branch (e.g. RepertoireGroup). No harm done, I noticed the mistake and was able to easy fix.

schristley:
> @schristley `repertoire_id` was removed purposefully based on discussions here: [#530 (comment)](https://github.com/airr-community/airr-standards/pull/530#issuecomment-892992427)
> 
> I see it has been added back in on your "put back accidentally deleted stuff" commit above. Is that a mistake? If not we should discuss...

Patience my young padawan ;-D Yes, it was a mistake in William's commit but I think I've got them all cleaned up.

bcorrie:
@schristley another question on the ADC API germline additions... 

Having a /germline and /gene entry point makes sense to me.

I am not so sure about having the /repertoire end point returning a array of `GermlineSet` and `GeneDescription` objects. 

For the most part, don't we want `Repertoire` to point to `GermlineSet` identifiers when and where they are used and then look them up using the /germline API. What does it mean when for the /repertoire end point to return a `GermlineSet` (or `GeneDescription`) at the same level as the Repertoire object? I am unclear of the semantics of what that means?

schristley:
> @schristley another question on the ADC API germline additions...

I want to resolve all of the API changes in a [separate PR. ](https://github.com/airr-community/airr-standards/pull/530#issuecomment-912873763)Because we are talking about '/airr/v1' versus '/airr/v2' and other stuff that's more than just the germline additions. And yes, we need to discuss some of the semantics.

Would you prefer I remove all API changes from this PR, and move them over to a new one?

bcorrie:
> > @schristley another question on the ADC API germline additions...
> 
> I want to resolve all of the API changes in a [separate PR. ](https://github.com/airr-community/airr-standards/pull/530#issuecomment-912873763)Because we are talking about '/airr/v1' versus '/airr/v2' and other stuff that's more than just the germline additions. And yes, we need to discuss some of the semantics.
> 
> Would you prefer I remove all API changes from this PR, and move them over to a new one?

I put the /airr/v1 and /airr/v2 on the agenda for tomorrows CRWG meeting, to make sure everyone is on the same page. I agree probably good to make that change in a separate PR (the v1 v2 discussion) but I don't think there is much controversy there.

I am OK with adding the /germline and /gene stuff to this PR, just like we have with /cell and /clone. These are returning the objects as defined in the spec with the "elegant" 1:1 mapping of API entry point to a set of our top level Objects. 8-)

But we could move that and the discussion for /repertoire returning these objects to a separate PR. If we want to close this pull request off and merge into master in a timely manner with just the spec changes, we could move the API stuff to a separate PR. I am OK with either but the API stuff might slow down finishing this PR - which maybe we don't want to do 8-)

williamdlees:
Sorry Scott.

schristley:
had to manually merge

bcorrie:
Woo hoo!!! I would buy a round of beer if I could 8-)

schristley:
> Woo hoo!!! I would buy a round of beer if I could 8-)

and no beer emoji on github, how sad

schristley:
Gotta do it the old school way, I guess :-D

```
                              .sssssssss.
                        .sssssssssssssssssss
                      sssssssssssssssssssssssss
                     ssssssssssssssssssssssssssss
                      @@sssssssssssssssssssssss@ss
                      |s@@@@sssssssssssssss@@@@s|s
               _______|sssss@@@@@sssss@@@@@sssss|s
             /         sssssssss@sssss@sssssssss|s
            /  .------+.ssssssss@sssss@ssssssss.|
           /  /       |...sssssss@sss@sssssss...|
          |  |        |.......sss@sss@ssss......|
          |  |        |..........s@ss@sss.......|
          |  |        |...........@ss@..........|
           \  \       |............ss@..........|
            \  '------+...........ss@...........|
             \________ .........................|
                      |.........................|
                     /...........................\
                    |.............................|
                       |.......................|
                           |...............|
```

bcorrie:
@bussec since this pull request is merged and there is an outstanding task, can we delete the task from this pull request. We have an issue for it, but it seems wrong to have it as an unfinished task in this pull request since the pull request is merged. Suggesting we delete the task from this pull request here: https://github.com/airr-community/airr-standards/pull/530#issue-917241743

schristley:
@bcorrie all done :-D

bussec:
Thanks @schristley !

Issue #486, opened 2020-12-11
Where would one annotate metadata with an adjuvant 


bcorrie:
In the case of an immunization study, where would be the right place to make note of any adjuvants used in a study? A question from a user looking for such information in the AIRR Data Commons. The only place I could see that would make sense would be in the field `immunogen` with the adjuvant listed as part of describing the immunogen???

Their suggestion was that maybe we would want to have such a field to encourage its curation while describing a study?

bussec:
Arrrg! :pirate_flag: I am just realizing that the parent object of this field -- like `intervention` -- is in only called `Diagnosis`, which doesn't make much sense...

Anyhow: `immunogen` should only describe the antigen that the immune system is expected to react against. If this is mixed with an adjuvant, the latter is described as part of the `intervention`, i.e. as free text.

Nevertheless,  I agree that there is a use case in which a research wants to see all studies using a certain adjuvant. So having a separate field would make sense. And as this seems to be a quite central use case for AIRR-seq studies, maybe we could solve it for good with a separate `Immunization` Object (instead of a quick fix within `Diagnosis`):

```
Immunization:
    discriminator: AIRR
    type: object
    required:
        - schedule
        - schedule_unit
        - procedure
    properties:
        schedule:
            type: array
            description: Immunization schedule, starting at timepoint 0.
            items:
                type: float
            title: Immunization schedule
            example:
                - 0
                - 7
                - 21
            x-airr:
                miairr: important
                nullable: true
                adc-query-support: true
                set: 1
                subset: diagnosis and intervention
                name: Immunization schedule
        schedule_unit:
            $ref: '#/Ontology'
            description: Unit of immunization schedule
            title: Unit of immunization schedule
            example:
                id: UO:0000033
                label: day
            x-airr:
                miairr: important
                nullable: true
                adc-query-support: true
                set: 1
                subset: diagnosis and intervention
                name: Immunization schedule unit
                format: ontology
                ontology:
                    draft: false
                    top_node:
                        id: UO:0000003
                        label: time unit
        procedure:
            type: array
            description: Information for immunization procedure at each scheduled time point
            items:
                $ref: '#/Procedure'
            x-airr:
                nullable: false
                adc-query-support: true    

Procedure:
    discriminator: AIRR
    type: object
    required:
        - immunogen
        - immunogen_amount
        - immunogen_amount_unit
        - adjuvant
        - adjuvant_amount
        - adjuvant_amount_unit
        - application_route
        - application_site
[...]

PR #533, opened 2021-06-12
draft of repertoire groups

### LINKED ISSUE #374: a simple proposal for repertoire groups and time course
Comparative analysis between repertoire groups (e.g. Treatment vs. Control) is a common procedure, and time course is also an increasingly common experimental design, so it would be nice to have a simple way to define and document them in the metadata.

I have two proposals, both relatively simple. The first proposal is to have an explicit `RepertoireGroup` object that defines the group.

```
RepertoireGroup:
    discriminator: AIRR
    type: object
    properties:
        group_id:
            type: string
        group_name:
            type: string
        group_description:
            type: string
        repertoires:
            type: array
            items:
                type: string
        time_course:
            type: array
            items:
                type: string
```

I'm not sure if we need `group_id` or whether `group_name` is sufficient; it might depend a bit on how we want to manage provenance.

The `repertoires` array is a list of `repertoire_ids` (or PIDs), thus defining the group.

For time course, as `repertoires` is an array, it defines an explicit ordering of those repertoires. Then the `time_course` array, which should be the same length as `repertoires`, provides user-defined labels (e.g. ["5w", "6m", "1y"] or ["1", "2", "3"] or ["pre", "post"]) for each time point.

In terms of the file format, the metadata file has an `Info` object and a `Repertoire` array, we could just add a `RepertoireGroup` array and store it in the same file.

The other proposal is an implicit definition. Here we add a couple of fields to the `Repertoire` object.

```
Repertoire:
    discriminator: AIRR
    type: object
    properties:
        groups:
            type: array
            items:
                type: string
        time_course:
            type: string
```

This acts as a simple tagging scheme, you add user-defined "tags" to the `groups` array (e.g. ["treatment", "another"]).

Then to operate on a group, you would iterate over all of your repertoires and collect the ones with the same tag like "treatment".

For time course, I'm assuming the repertoire has just one time course label. I suppose `time_course` could also be an array, and then `groups` have to be kept in sync (order wise) with `time_course`.

This implicit definition requires that you know the set of repertoires, and there is the chance for tag conflict/confusion if repertoires from different studies are used together. There is also the question of provenance if a new study groups repertoires in a different way, i.e. adds new tags to `groups`, how would one record that? Therefore, because of these issues, I'm less inclined towards this implicit definition but I offer it as it seems even simpler.

VDJServer currently support repertoire groups, but it goes a step further by allowing the groups to be dynamically defined. For example, you can provide a "group by" variable such as `subject.subject_id` and it will create groups for each value in the variable, in this case repertoires grouped by each subject. Another method is a simple logical statement, `subject.sex == male`, to group repertoires where the statement true. By the time of execution, these all essentially get translated into a static list of repertoires like in `RepertoireGroup` above. However, we could consider something like this if we want to go beyond the simple proposal.

schristley:
closes #374 #445


schristley:
@scharch I put in the simple version as a draft. There are still a number of outstanding issues though

- How does this work with the ADC API? Do we need a query end point?
- How this ties into `DataProcessing` is TBD.
- Does not support the advanced filtering suggested in #445

scharch (commented):
I think this structure leaves enough room to add filtering later that I'm ok to proceed on that front. However, I'd like the `time_course` array to be more general --`RepertoireGroup`s will also be used for case-control studies, population studies, etc etc. So a `repertoire_descriptions` array seems more appropriate, and maybe we can add a `study_type` field (perhaps with a controlled vocabulary?) to identify when those descriptions should be interpreted as timepoints.

schristley:
> I think this structure leaves enough room to add filtering later that I'm ok to proceed on that front. However, I'd like the `time_course` array to be more general --`RepertoireGroup`s will also be used for case-control studies, population studies, etc etc. So a `repertoire_descriptions` array seems more appropriate, and maybe we can add a `study_type` field (perhaps with a controlled vocabulary?) to identify when those descriptions should be interpreted as timepoints.

I don't understand how an array of repertoire_descriptions would be used in those cases, can you give an example?

scharch:
Imagine a use case comparing responses in aged vs young individuals. There are 20 repertoires in this `RepertoireGroup`, 10 from each cohort. The `repertoire_description` array might look something like `['Aged-1', 'Aged-2', .... 'Aged-10', 'Youth-1', ... 'Youth-10']`.

Or maybe it's a longitudinal COVID study with `repertoire_description=['Severe-A-1_month', 'Severe-A-6_months', ... 'Mild-C-1_year', ... 'Prepandemic-A']`, which wouldn't fit neatly into a `time_course`, despite being a longitudinal study.

schristley:
> Imagine a use case comparing responses in aged vs young individuals. There are 20 repertoires in this `RepertoireGroup`, 10 from each cohort. The `repertoire_description` array might look something like `['Aged-1', 'Aged-2', .... 'Aged-10', 'Youth-1', ... 'Youth-10']`.

Ok, I see what you mean. Though I think this example should really be two groups, one for aged and one for youth, and then you're doing comparative analysis between the two groups. I don't think we want to encourage tools to "interpret" the description, that sounds how to reproduce...

Part of what you mention for the description, I was actually thinking that would be in the Repertoire versus in the RepertoireGroup. Each [Repertoire](https://github.com/airr-community/airr-standards/blob/4d13b8afcac74b869823b1367a623cf9c578788a/specs/airr-schema.yaml#L1903) has its own `repertoire_name` and `repertoire_description`. However, thinking about it, I can see that maybe you want to create a RepertoireGroup for a Repertoire you don't control (e.g. its from a different study) so you cannot change the fields...

Ok, I'm going to restructure then to allow having a description attached to each Repertoire in the RepertoireGroup.

schristley:
@scharch I've introduced a more detailed TimePoint object that's based on these [discussion](https://github.com/airr-community/airr-standards/issues/486) and re-structured RepertoireGroup.

scharch (approved):
Looks great (though I guess `description: stuff` is a placeholder?)

PR #537, opened 2021-07-09
Error in doc

ssnn-airr:
Here, and should be or. Looks like a copy/paste error when creating the examples.


schristley:
thanks @ssnn-airr !

PR #539, opened 2021-07-21
Add ADC specific gene and subgroup query fields

bcorrie:
When we closed https://github.com/airr-community/airr-standards/pull/295 we agreed that these fields should be added to the ADC API but not be part of the actual file format standard. They are a query convenience. When we closed the other issue we didn't create something to capture this need.

@schristley not sure how to do this, but I copied the mechanism you used for adc_update_date.

Lets add this to discussion at CRWG tomorrow.

### DIFF
--- a/specs/adc-api-openapi3.yaml
+++ b/specs/adc-api-openapi3.yaml
@@ -121,39 +121,40 @@ components:
             type: integer
   
     # list of rearrangement extension fields
-    rearrangement_extension_fields:
-      type: object
-      properties:
-        adc_v_subgroup:
-          type: string
-          description: >
-            V subgroup of the V gene. If referring to a known reference sequence in a database, such as IMGT/GENE-DB,
-            the relevant gene nomenclature should be followed (e.g., IGHV4).
-          example: IGHV4
-          x-airr:
-            nullable: true
-            adc-api-optional: false
-            adc-query_support: true
-            name: V gene subgroup
-        adc_v_gene:
-          type: string
-          description: >
-            V gene. If referring to a known reference sequence in a database, such as IMGT/GENE-DB,
-            the relevant gene nomenclature should be followed (e.g., IGHV4-59).
-          example: IGHV4-59
-          x-airr:
-            nullable: true
-            adc-api-optional: false
-            adc-query_support: true
-            name: V gene  
+    rearrangement_extension:
+      description: The extended Rearrangement object with additional query fields for the ADC.
+      allOf:
+        - $ref: https://raw.githubusercontent.com/airr-community/airr-standards/master/specs/airr-schema-openapi3.yaml#/Rearrangement
+        - type: object
+          properties:
+            adc_v_subgroup:
+              type: string
+              description: >
+                V subgroup of the V gene. If referring to a known reference sequence in a database, such as IMGT/GENE-DB,
+                the relevant gene nomenclature should be followed (e.g., IGHV4).
+              example: IGHV4
+              x-airr:
+                nullable: true
+                adc-api-optional: false
+                adc-query_support: true
+                name: V gene subgroup
+            adc_v_gene:
+              type: string
+              description: >
+                V gene. If referring to a known reference sequence in a database, such as IMGT/GENE-DB,
+                the relevant gene nomenclature should be followed (e.g., IGHV4-59).
+              example: IGHV4-59
+              x-airr:
+                nullable: true
+                adc-api-optional: false
+                adc-query_support: true
+                name: V gene
   
   # list of rearrangement annotations
     rearrangement_list:
       type: array
       items:
-        allOf:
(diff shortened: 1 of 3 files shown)


bcorrie:
See this comment for more info... https://github.com/airr-community/airr-standards/pull/295#issuecomment-811137033

bcorrie:
We probably want these for clones as well???

schristley:
Based on discussion in CRWG, we should put these fields into extended objects in the ADC API spec versus in the AIRR schema spec. Also add some doc generation code.

bcorrie:
> Based on discussion in CRWG, we should put these fields into extended objects in the ADC API spec versus in the AIRR schema spec. Also add some doc generation code.

@schristley I think we have some challenges here... We are essentially saying that we have some extra query-able fields (e.g.  `adc_v_subgroup`) and that we want to add them to the spec. There are two components to this:

- I know we want to add them to the query spec, so that users know they can use them in queries. This is our main use case, as we want to provide an optimized query for subgroup and family. The issue here is that we don't have a spec for queries, our `filters` object is just that, an object, in the spec, and it is only defined in the docs. The docs say that you need to provide a fully qualified query object (e.g. `subject.subject_id`) from the spec, but there is no spec definition for this. So I don't think there is any way to add these fields for queries other than add them to the docs. So really to do this we simply have to add a paragraph in the docs that says in addition to query on AIRR fields, you can also query on some special fields. Correct me if I am wrong.
- Do we want to add them to the query response? You can query for `v_subgroup` and you get all the rearrangements (or clones) that meet your query criteria, and this would return v_call. Do we want to return a `v_subgroup` field? I am not sure that is necessary or desirable? The main thing I think is important is the query convenience... If we do want this, then we can indeed add these fields to the spec as the query response is well defined. But the question is do we want this.

Any thoughts?

bcorrie:
As discussed at CRWG, we do want to return these in the API response, so they need to be added to the API spec.

We need to figure out a good way to document these special fields in the AIRR docs.

bcorrie:
@schristley I am trying to add the gene query and response extensions - and what I think should work is not... I am trying to use the allOf attribute to include two objects in the list of fields that are being returned.

When I look at the spec (try the OpenAPI 3) in the swagger editor, it says there are no errors, but the response according to the swagger editor is just a string. So it isn't clear to me if this is correct or not? Any thoughts... 

The swagger editor says the response is this:

```
{
  "Info": {
    "title": "string",
    "version": "string",
    "description": "string",
    "contact": {
      "name": "string",
      "url": "string",
      "email": "string"
    },
    "license": {
      "name": "string",
      "url": "string"
    }
  },
  "Rearrangement": [
    "string"
  ],
  "Facet": [
    {
      "count": 0
    }
  ]
}
```

schristley:
@bcorrie `allOf` may be valid within array's items but maybe not. The spec is unclear on that point, and I only see examples of `allOf` in objects. Anyways, I think it looks a little cleaner to construct an extended rearrangement object. This seems to work properly with the swagger editor.

bcorrie:
I still get the odd behaviour in that the swagger editor (and other API preview engines e.g. the one in JIRA) report the API response as being a string (see the screen grab below). So no errors, but the API response is not shown as a list of fields like it normally is when allOf is used. Now maybe that is a limit of the Swagger editor???

image

Do we need to worry about this?

bcorrie:
This is what it looks like without the allOf - it gives a list of the fields...

image

schristley:
Yeah, it's odd. It was working for me last night but I'm getting the same as you now. Even though the response looks odd, if you look at the models in the editor, they are correct, so yeah maybe some weird bug with the editor.

schristley:
Try this, cut out these lines from the spec in the editor (which should cause an error), then paste them back in. When I do that, the response looks correct.

```
    rearrangement_list:
      type: array
      items:
        $ref: '#/components/schemas/rearrangement_extension'
```

bcorrie:
Hmm, I can't get that to work, but if you are OK with it in terms of format, then I am OK.

bcorrie:
Yes model looks correct. I will add in D/J/C fields...

bcorrie:
@williamdlees a question for you... We have added some "helper" ADC query fields, so you can specify a gene and subgroup for VDJC calls.  So we have v_call, which is an AIRR field, and the adc_v_subgroup and adc_v_gene that are "helper" fields for query ease of use and optimization. These exist in the API only, not in the spec (as per this pull request),

We were wondering what your opinion was on having an adc_j_subgroup and adc_c_subgroup? In IMGT nomenclature, J and C genes have no subgroups. Is that a biological thing or is that a feature of the IMGT nomenclature as it is today. Should we remove these fields to avoid confusion? Would OGRDB ever have a subgroup for J and C?

Any suggestions?

bcorrie:
For the record, the problem with swagger is odd, the extension object models are correct, and the correct response is rendered correctly sometimes (for me the clone_extension always seems to work but not the rearrangement_extension). We are assuming that these are correct even though swagger gets confused.

schristley (approved):

PR #541, opened 2021-08-09
Reduced run time of R-CMD-check in Travis

javh:
Tweaks to Travis R rules to attempt to reduce check time.


javh:
R package tests down from ~9:30 to ~3:45. I'm not seeing an obvious way to reduce the run time further.

bussec (approved):
LGTM

PR #546, opened 2021-09-13
Test R-CMD-check using GitHub Actions

javh:
A test PR using GitHub actions.

### DIFF
--- a/.github/workflows/py-unittest.yaml
+++ b/.github/workflows/py-unittest.yaml
@@ -1,0 +1,35 @@
+# Workflow derived from https://github.com/r-lib/actions/tree/master/examples
+# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
+on:
+  push:
+    branches: [main, master]
+  pull_request:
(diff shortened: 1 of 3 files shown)


javh:
Check runs in 3:43 using the rocker/verse image.  ~2:30 of that is `R CMD check` itself, which isn't going to get any faster. 

~50 sec is setting up the container - not sure yet if that can be reduced via caching

bussec (approved):
LGTM, thanks!

javh:
Coolness. I'll add the python tests and see if I can use a smaller container before merging.

javh:
First draft of GitHub actions for the schema consistency checks, R package, and python package are now in this PR. They are in separate configs in the `.github/workflows` directory.

I also added a condition to whether the tests will be run for the R and python packages that requires a change in either the GitHub actions config or `lang/R` / `lang/python` as appropriate. Which should save a lot of time, because then we don't need to run the R tests every time we change the schema.  It's a bit annoying to test though, because the workflow is skipped, so I can disable it for now if we want.

PR #548, opened 2021-09-14
Addition of a manifest object.

### LINKED ISSUE #426: Define a manifest mechanism

bcorrie:
For discussion about format and changes.

Closes #426


bcorrie:
Basic iReceptor Gateway manifest for files in a download would look like this:

```
{
"Info":{},
"DataSets":
[
    {"repertoire_file":"covid-19-1-metadata.json", "rearrangement_file":"covid-19-1.tsv"},
    {"repertoire_file":"covid-19-2-metadata.json", "rearrangement_file":"covid-19-2.tsv"},
    {"repertoire_file":"covid-19-3-metadata.json", "rearrangement_file":"covid-19-3.tsv"},
    {"repertoire_file":"covid-19-4-metadata.json", "rearrangement_file":"covid-19-4.tsv"},
    {"repertoire_file":"ireceptor-public-archive-3-metadata.json", "rearrangement_file":"ireceptor-public-archive-3.tsv",
    {"repertoire_file":"vdjserver-metadata.json", "rearrangement_file":"vdjserver.tsv"}
]
}

**Comment by bcorrie** [comment]
This currently has one file of each time per data set. I think the main question is do we need an array of each file type per dataset?

**Comment by bcorrie** [comment]
```
{
"Info":{},
"DataSets":
[
    {"repertoire_file":["covid-19-1-metadata.json"], "rearrangement_file":["covid-19-1.tsv"]},
    {"repertoire_file":["covid-19-2-metadata.json"], "rearrangement_file":["covid-19-2.tsv"]},
    {"repertoire_file":["covid-19-3-metadata.json"], "rearrangement_file":["covid-19-3.tsv"]},
    {"repertoire_file":["covid-19-4-metadata.json"], "rearrangement_file":["covid-19-4.tsv"]},
    {"repertoire_file":["ireceptor-public-archive-3-metadata.json"], "rearrangement_file":["ireceptor-public-archive-3.tsv]",
    {"repertoire_file":["vdjserver-metadata.json"], "rearrangement_file":["vdjserver.tsv"]}
]
}
```

**Comment by bcorrie** [comment]
A typical study as a single data set might look like:

```
{
"Info":{},
"DataSets":
[
  {    
        "repertoire_file":["PRJNA42.json"],
        "rearrangement_file":
        [
            "subject1.tsv","subject2.tsv","subject3.tsv","subject4.tsv","subject5.tsv"
        ]
  }
]
}
```

**Comment by schristley** [comment]
> This currently has one file of each time per data set. I think the main question is do we need an array of each file type per dataset?
> 
> I suspect we do. So a manifest is an array of data sets, and each data set has an array of files of each type.

Hmm, yeah, that's an interesting question if we want the manifest to represent a single AIRR data set, or if it should represent multiple data sets. I guess I was imagining it to represent just a single one, but multiple data sets shouldn't be that hard to support, as you say, it's just an array of arrays.

**Comment by bcorrie** [comment]
> > This currently has one file of each time per data set. I think the main question is do we need an array of each file type per dataset?
> > I suspect we do. So a manifest is an array of data sets, and each data set has an array of files of each type.
> 
> Hmm, yeah, that's an interesting question if we want the manifest to represent a single AIRR data set, or if it should represent multiple data sets. I guess I was imagining it to represent just a single one, but multiple data sets shouldn't be that hard to support, as you say, it's just an array of arrays.

I think I would definitely prefer that - as without it we can't really represent an iReceptor Gateway download as a single manifest file.

A data set in this context is something where the files are expected to have any cross referencing ID fields to be unique in the context. It is a grouping of files that can be processed together without further manipulation. That is in a data set, if there is a repertoire_id in a repertoire file, any of the other files (rearrangement, clone, cell) that have the same repertoire_id would be considered to be talking about the same Repertoire. I kind of thing of the Manifest mapping to having a similar uniqueness criteria to what we have for a repository.

From an ADC context, I think we want a manifest to be able to represent more than one data set where that uniqueness criteria is true but it isn't necessarily true across all data sets within the manifest. For example, we want to have a data set from N repositories, where the uniqueness criteria is valid within each of the N data sets but the N data sets might have a repertoire_id in common. But that is OK, because it is an error to process a rearrangement file with a repertoire file that is not part of the same data set... Similarly for processing files in a manifest.

**Comment by schristley** [comment]
> A data set in this context is something where the files are expected to have any cross referencing ID fields to be unique in the context. It is a grouping of files that can be processed together without further manipulation. That is in a data set, if there is a repertoire_id in a repertoire file, any of the other files (rearrangement, clone, cell) that have the same repertoire_id would be considered to be talking about the same Repertoire. I kind of thing of the Manifest mapping to having a similar uniqueness criteria to what we have for a repository.

That's right, you never really were on board with making repertoire_ids be globally unique... Sigh, tool developers are going to be annoyed with us. If I do a query for all cancer data, now I (the tool) am going to have manage all of the conflicting IDs if I try creating repertoire groups and do analysis that crosses "data sets", what a pain...

**Comment by scharch** [comment]
I don't understand the reason for having `DataSetManifest` and `Manifest` be separate objects...

Anyway, iReceptor should be returning everything as part of a single `RepertoireGroup`. But the slots for file names still have to be arrays, because we have no requirements on keeping things together (or how to organize them if broken up.

**Comment by bcorrie** [comment]
> That's right, you never really were on board with making repertoire_ids be globally unique... Sigh, tool developers are going to be annoyed with us. If I do a query for all cancer data, now I (the tool) am going to have manage all of the conflicting IDs if I try creating repertoire groups and do analysis that crosses "data sets", what a pain...

I'm not against it, we just have to figure out how to do it.

In https://github.com/airr-community/airr-standards/issues/246#issuecomment-611218230 we kind of came to the conclusion that we would probably use another field to get uniqueness globally and that repertoire_id would be unique in the context of a data set (repository, study, etc) being considered (eg. a DataSetManifest above). In #347 we talk about global uniqueness and persistence. So this is still an open question.

**Comment by bcorrie** [comment]
> I don't understand the reason for having `DataSetManifest` and `Manifest` be separate objects...

Are you concerned about having separate YAML objects or the structure of having an array of DataSetManfests. They certainly don't have to be separate, I did that for clarity of definition.

My rationale for the structure is that:

A `DataSet` consists of a group of logically related files and the `DataSetManifest` describes their relationship. It groups a set of `repertoire_files` that describes the metadata about the `rearrangement_files` and `clone_files`.

A `Manifest` simply holds an array of `DataSetManifests`. This allows a manifest to describe N data sets that might have different file groupings.

**Comment by bcorrie** [comment]
> Anyway, iReceptor should be returning everything as part of a single `RepertoireGroup`. But the slots for file names still have to be arrays, because we have no requirements on keeping things together (or how to organize them if broken up.

I don't think it can, and I am not sure it wants to. 

It can't because of the issue raised by Scott above. The iReceptor Gateway pulls data from many repositories, so repertoire_id is not unique across repositories. So we can't have a single RepertoireGroup with a list of repertoire_ids because there could be conflicts/collisions.

It doesn't want to because it doesn't know how to group the Repertoires. The data that is eventually downloaded from a query can be generated from a very complex process of data refinement. So the Repertoires can be grouped in a widely varying set of ways and the Gateway has no clue as to how the researcher might want to group the Repertoires from that query. The only grouping that really makes sense is to group at the repository level. Now one could create a RepertoireGroup at the repository level, but that is pretty well identical to the list of Repertoires that you get in Repertoire JSON file for the repository, so it is redundant.

So a manifest for an iReceptor Gateway download would look like this: https://github.com/airr-community/airr-standards/pull/548#issuecomment-919567744

**Comment by scharch** [comment]
> The iReceptor Gateway pulls data from many repositories, so repertoire_id is not unique across repositories.

OK good point. So I guess one `RepertoireGroup` from each repository (each with its own `Manifest`) then?



> It doesn't want to because it doesn't know how to group the Repertoires. The data that is eventually downloaded from a query can be generated from a very complex process of data refinement.

Well, ok for now, but the long term goal is precisely for `RepertoireGroup` to capture that refinement.

**Comment by schristley** [comment]
> > The iReceptor Gateway pulls data from many repositories, so repertoire_id is not unique across repositories.
> 
> OK good point. So I guess one `RepertoireGroup` from each repository (each with its own `Manifest`) then?

I don't see how those groups are particularly useful.

> > It doesn't want to because it doesn't know how to group the Repertoires. The data that is eventually downloaded from a query can be generated from a very complex process of data refinement.
> 
> Well, ok for now, but the long term goal is precisely for `RepertoireGroup` to capture that refinement.

My understanding is that you want the `RepertoireGroup` to reflect the query results, i.e. here's all the repertoires that were returned from the query? I guess that's primarily a convenience because you can also walk through the list of repertoires returned and construct that group yourself.

Regardless, that is an interesting enhancement for API V2. The `repertoire` end point would return repertoires as normal from a query but also return a `RepertoireGroup` with that query. We could also have a simple function in the reference libraries that construct a `RepertoireGroup` given a metadata file.

**Comment by bcorrie** [comment]
> > > The iReceptor Gateway pulls data from many repositories, so repertoire_id is not unique across repositories.
> > 
> > 
> > OK good point. So I guess one `RepertoireGroup` from each repository (each with its own `Manifest`) then?
> 
> I don't see how those groups are particularly useful.

For me, I think a RepertoireGroup file would be redundant for the iReceptor Gateway. Essentially an iReceptor Gateway download comes from a /repertoire query generating a set of repertoire_ids, which is then used to download all the AIRR rearrangement data for those repertoire_ids. You get a repertoire JSON file that contains the list of Repertoires and an AIRR TSV file that contains the rearrangements from those repertoires (unless you filter at the sequence/rearrangement level). 

The manifest is simple: 

```
{
"Info":{},
"DataSets":
[
    {"repertoire_file":["covid-19-1-metadata.json"], "rearrangement_file":["covid-19-1.tsv"]}
]
}
```

Recall that the Repertoire JSON file is simply an array of Repertoire information, one per repertoire_id. So this is essentially the same as the RepertoireGroup file that would be used to represent the data. The RepertoireGroup would be the same array of repertoire_ids as that in the Repertoire JSON file, without all the metadata for the Repertoire. So in this case, it would be redundant. A RepertoireGroup file for us would essentially be a subset of the full Repertoire JSON file with just the fields `repertoire_id`, `repertoire_description`, and `subject.sample.collection_time_point_relative`.

Now if the user wanted to split the data that they downloaded to do comparative analyses (say they downloaded all Homo sapiens COVID-19 IGH data) and wanted to split the data to look at gender differences. For that analysis I could see how you would want to slice the data based on gender and generate two RepertoireGroup files, one for each gender. You can slice that data in hundreds of different ways, but that is part of the downstream analysis, where RepertoireGroup files would be extremely important. But that isn't the role of the Gateway.

**Comment by bcorrie** [comment]
> Regardless, that is an interesting enhancement for API V2. The `repertoire` end point would return repertoires as normal from a query but also return a `RepertoireGroup` with that query. We could also have a simple function in the reference libraries that construct a `RepertoireGroup` given a metadata file.

A function that would split a Repertoire JSON file based on a specific Repertoire field into a set of RepertoireGroup files would be VERY useful IMHO. 

To do the gender split I mention above, you just give the utility a JSON file and a field (e.g. subject.sex) and it would go through the JSON file, gather the possible field values from the file, construct a set of repertoire_ids for each, and then generate a RepertoireGroup file for each possible value.

```
repertoire_split subject.sex covid-19-1-metadata.json
```

Would generate N files where N is the number of distinct values that existed in the subject.sex field. For example, if the file contained "male", "female", and "not collected" you would get three RepertoireGroup files as follows:

covid-19-1-metadata_male.json
covid-19-1-metadata_female.json
covid-19-1-metadata_not_collected.json

You could then use the repertoire_id fields to easily analyze rearrangement data from male and female subjects from covid-19-1.tsv

**Comment by javh** [comment]
From the call:
* We still want to do this.
* @scharch will meditate on this concept and the strategy.

**Comment by bcorrie** [comment]
FYI we went ahead and implemented something like this for downloads from the iReceptor Gateway. If you do a download from the Gateway you get a manifest.json file that describes the download from each repository. It follows the idea of the manifest object here but it might not match the spec as we have it - it is our own internal implementation. We use it to explicitly describe how Repertoire files are linked to Rearrangement/Clone/Cell/GEX data you get from a repository.

Our manifest file looks like this:

```
{
    "Info": {
        "title": "AIRR Manifest",
        "version": "3.0",
        "description": "List of files for each repository",
        "contact": {
            "name": "iReceptor Gateway",
            "url": "https://gateway-staging.ireceptor.org",
            "email": "support@ireceptor.org"
        }
    },
    "DataSets": [
        {
            "repository": "IPA 6",
            "repository_url": "https://ipa6.ireceptor.org/airr/v1/",
            "repertoire_file": "ireceptor-public-archive-metadata.json",
            "rearrangement_file": "ireceptor-public-archive-rearrangement.tsv",
            "cell_file": "ireceptor-public-archive-cell.json",
            "expression_file": "ireceptor-public-archive-gex.json"
        }
    ]
}
```

**Comment by scharch** [comment]
> FYI we went ahead and implemented something like this for downloads from the iReceptor Gateway. 

Thanks @bcorrie, this is definitely a useful starting point. However, in rereading the discussion above, it still feels like we might need to resolve `RepertoireGroup` before coming back around to this.

Two comments on what you have, though:
1. It looks like it should be relatively easy to adapt


>             "repository": "IPA 6",
>             "repository_url": "https://ipa6.ireceptor.org/airr/v1/",
>             "repertoire_file": "ireceptor-public-archive-metadata.json",

to a local repertoire. Any complications that you see?

2. I feel like we might want to make the schema a little more flexible/less pre-specified. In the current PR, there's no slot for GEX, because it wasn't on our radar two years ago. Or raw fastqs (https://github.com/airr-community/airr-standards/issues/426#issuecomment-735000321_). Or splitting different parts of the schema across multiple files (https://github.com/airr-community/airr-standards/issues/426#issuecomment-876707931_)

**Comment by bcorrie** [comment]
Thoughts after the call today, and based on the above:

> Two comments on what you have, though:
> 
> 1. It looks like it should be relatively easy to adapt
> 
> > ```
> >         "repository": "IPA 6",
> >         "repository_url": "https://ipa6.ireceptor.org/airr/v1/",
> >         "repertoire_file": "ireceptor-public-archive-metadata.json",
> > ```
> 
> to a local repertoire. Any complications that you see?

If I understand correctly, I don't think so. If you had a local data set you were describing, you would just have a null `repository_url` and `repository` field. In fact, `repository` as a field should probably be renamed `dataset_name` or something like that.

> 2. I feel like we might want to make the schema a little more flexible/less pre-specified. In the current PR, there's no slot for GEX, because it wasn't on our radar two years ago. Or raw fastqs ([Define a manifest mechanism #426 (comment)](https://github.com/airr-community/airr-standards/issues/426#issuecomment-735000321)_). Or splitting different parts of the schema across multiple files ([Define a manifest mechanism #426 (comment)](https://github.com/airr-community/airr-standards/issues/426#issuecomment-876707931)_)

Yes, no objections to other files per se, the ones I listed above map directly to the AIRR Objects such as `Repertoire`, `Rearrangement`, `Cell` and `CellExpression` but the manifest format should probably be more flexible/extensible. Indeed, the iReceptor Gateway downloads now also provide query JSON files that I would like to be able to add to the manifest.

**Comment by bcorrie** [comment]
Also, I am pretty sure we want to go back to having an array of file names for each file type, more like what we had here:

https://github.com/airr-community/airr-standards/pull/548#issuecomment-923109956

I would want to be able to do something like this:

```
{
"Info":{},
"DataSets":
[
    {"repertoire_file":["study-metadata.json"], 
     "rearrangement_file":["subject1.tsv","subject2.tsv", ... "subjectN.tsv"]}
]
}
```
I think that is actually a bug in our manifest generation code - I think the intent was always to have an array of files, but our code I think drops the array when there is only one file (and in our downloads there is always only one file) per dataset.

**Comment by schristley** [comment]
Another comment, we have `DataFile` defined which specifies the format of the AIRR JSON. It isn't restricted to `Repertoire` but can hold many of the AIRR objects (rearrangements excluded). This presents a slight conundrum because that data might be put all in one AIRR JSON file, or it might be spread across multiple AIRR JSON files. For example, germline sets coming from one source, repertoire and repertoire groups from another, and a third providing clone data, etc. The `repertoire_file` tag/type points to the AIRR JSON that hold the `Repertoire` data but then that means you need tags for all the other ones as well. If you want to be a bit self-reflective, you can use the same name for the tag/type as the object itself, essentially how `DataFile` is.

This also follows into have an array of file names, as all of the `Repertoires` might not be in the same file either but spread across multiple.

Another thing if you are considering ease of use and computational efficiency in your design. Without some link between the repertoires and the rearrangements (or any objects actually), one might have to search all of the rearrangement data, to find the entries for a specific repertoire. It might be nice if the structure kept inter-related objects together.

**Comment by bcorrie** [comment]
> The `repertoire_file` tag/type points to the AIRR JSON that hold the `Repertoire` data but then that means you need tags for all the other ones as well. If you want to be a bit self-reflective, you can use the same name for the tag/type as the object itself, essentially how `DataFile` is.

Yes, we could use the AIRR Object name as the tag (I think that is what you are suggesting). We could have:
```
{
"Info":{},
"DataSets":
[ {
     "Repertoire":["study-metadata.json"], 
     "Rearrangement":["subject1.tsv","subject2.tsv", ... "subjectN.tsv"]
    }
] }
```
Two potential issues:

- I used an explicit tag with `_file` because I wanted to make it clear that the field was a file name since not all fields in a `Manifest` are files. In our schemas we use tags like "Repertoire" when we are describing the full object (like in Datafile) but this is different.
- If we are considering having files with non-AIRR objects in them that might be a bit complicated/confusing???

**Comment by bcorrie** [comment]
> Another thing if you are considering ease of use and computational efficiency in your design. Without some link between the repertoires and the rearrangements (or any objects actually), one might have to search all of the rearrangement data, to find the entries for a specific repertoire. It might be nice if the structure kept inter-related objects together.

To me that is what the `DataSets` in a `Manifest` are for. In the iReceptor Gateway case, if a download comes from 4 different repositories then there are 4 `DataSets` in the `Manifest` for that download, and it would look like this:

```
{
    "Info": {
        "title": "AIRR Manifest",
        "version": "3.0",
        "description": "List of files for each repository",
        "contact": {
            "name": "iReceptor Gateway",
            "url": "https://gateway.ireceptor.org",
            "email": "support@ireceptor.org"
        }
    },
    "DataSets": [
        {
            "repository": "VDJServer",
            "repository_url": "https://vdjserver.org/airr/v1/",
            "repertoire_file": "vdjserver-metadata.json",
            "rearrangement_file": ["vdjserver.tsv"]
        },
        {
            "repository": "VDJbase",
            "repository_url": "https://airr-seq.vdjbase.org/airr/v1/",
            "repertoire_file": "vdjbase-metadata.json",
            "rearrangement_file": ["vdjbase.tsv"]
        },
        {
            "repository": "COVID 19-1",
            "repository_url": "https://covid19-1.ireceptor.org/airr/v1/",
            "repertoire_file": "airr-covid-19-metadata.json",
            "rearrangement_file": ["airr-covid-19.tsv"]
        }
    ]
}
```
All data from a repository is in a single `DataSet` and all files for that data set are listed so all of the related objects are linked.

But you could easily represent that same download grouped as DataSets in other ways.

The download from VDJServer in this download has two different Repertoires, so you could change the VDJServer DataSet to look like:

```
[Stuff deleted]
    "DataSets": [
        {
            "repository": "VDJServer",
            "repository_url": "https://vdjserver.org/airr/v1/",
            "repertoire_file": "vdjserver-metadata.json",
            "rearrangement_file": ["repertoire1.tsv", "repertoire2.tsv"]
        },
[Stuff deleted]
```
In this case all of the Repertoire data is in one file but the rearrangements are spread across two files.

Or you can:
```
[Stuff deleted]
    "DataSets": [
        {
            "repository": "VDJServer",
            "repository_url": "https://vdjserver.org/airr/v1/",
            "repertoire_file": "repertoire1.json",
            "rearrangement_file": ["repertoire1.tsv"]
        },
        {
            "repository": "VDJServer",
            "repository_url": "https://vdjserver.org/airr/v1/",
            "repertoire_file": "repertoire2.json",
            "rearrangement_file": ["repertoire2.tsv"]
        },
[Stuff deleted]
```
In this case all of the Repertoire and Rearrangement data for each Repertoire is described by a single DataSet.

All are valid manifests for the same data, just organized differently, but in all cases all related data is grouped together. The granularity of how the data is split is just different.

In the case above, the download has 2 Repertoires from VDJServer, 28 from VDJBase, and 44 from covid19-1. So the Gateway chooses to represent this data as 3 DataSets (one per repository), but it could be represented as 74 DataSets, one per Repertoire. The Manifest here supports any of those.

**Comment by schristley** [comment]
> > Another thing if you are considering ease of use and computational efficiency in your design. Without some link between the repertoires and the rearrangements (or any objects actually), one might have to search all of the rearrangement data, to find the entries for a specific repertoire. It might be nice if the structure kept inter-related objects together.
> 
> To me that is what the `DataSets` in a `Manifest` are for. In the iReceptor Gateway case, if a download comes from 3 different repositories then there are 3 `DataSets` in the `Manifest` for that download, and it would look like this:

Ok, perfect!

**Comment by schristley** [comment]
> > The `repertoire_file` tag/type points to the AIRR JSON that hold the `Repertoire` data but then that means you need tags for all the other ones as well. If you want to be a bit self-reflective, you can use the same name for the tag/type as the object itself, essentially how `DataFile` is.
> 
> Yes, we could use the AIRR Object name as the tag (I think that is what you are suggesting). We could have:

Yeah, that's what I meant. I agree using them muddies the semantics.


> Two potential issues:
> 
> * I used an explicit tag with `_file` because I wanted to make it clear that the field was a file name since not all fields in a `Manifest` are files. In our schemas we use tags like "Repertoire" when we are describing the full object (like in Datafile) but this is different.
> * If we are considering having files with non-AIRR objects in them that might be a bit complicated/confusing???

I've been looking at other JSON schemas, in particular PROV, and in some cases, they use a type as the name where the type is CURIE-like, i.e. nested namespaces separate with colons, `prov:type`, `prov:file`, and so on. It's a nice technique to manage a shared name space, which sounds like it might be useful for your 2nd bullet and for a manifest. Something like this with an `airr` namespace:

```
[Stuff deleted]
    "DataSets": [
        {
            "airr:repository:name": "VDJServer",
            "airr:repository:url": "https://vdjserver.org/airr/v1/",
            "airr:file:Repertoire": "vdjserver-metadata.json",
            "airr:file:Rearrangement": ["repertoire1.tsv", "repertoire2.tsv"]
        },
[Stuff deleted]
```

Maybe having fixed names is fine, but might be worth considering if the implied namespace hierarchy from the type can be useful for the manifest in some way. Might be useful for automation.

**Comment by bcorrie** [comment]
Here is a new, more extensible manifest attempt where each set of files is of a specific type, and the types are the AIRR Schema Objects (`Repertoire`, `Rearrangement` etc). This means that 1) when the AIRR schema is extended with new Schema objects, values change, not the keys in the schema and 2) you can have non-AIRR files in your dataset.

A simple manifest would look like this:

```
{
    "Info": {
        "title": "AIRR Manifest",
        "version": "3.0",
        "description": "List of files for each repository",
        "contact": {
            "name": "iReceptor Gateway",
            "url": "https://gateway.ireceptor.org",
            "email": "support@ireceptor.org"
        }
    },
    "DataSets": [
    {
        "dataset_name": "VDJServer Repertoire 1",
        "repository_url": "https://vdjserver.org/airr/v1/",
        "dataset": [
            { "data_type" : "Repertoire", "data_files" : [ "repertoire1.json"] },
            { "data_type" : "Rearrangement", "data_files" : ["repertoire1.tsv"] },
            { "data_type" : "fastq", "data_files" : ["repertoire1.fastq"] }
        ]
    }
}
```

**Comment by bcorrie** [comment]
A typical iReceptor Gateway Rearrangement data download from three repositories would look like this:

```
{
    "Info": {
        "title": "AIRR Manifest",
        "version": "3.0",
        "description": "List of files for each repository",
        "contact": {
            "name": "iReceptor Gateway",
            "url": "https://gateway.ireceptor.org",
            "email": "support@ireceptor.org"
        }
    },
    "DataSets": [
        {
            "dataset_name": "VDJServer",
            "repository_url": "https://vdjserver.org/airr/v1/",
            "dataset": [
              { "data_type" : "Repertoire", "data_files" : [ "vdjserver-metadata.json"] },
              { "data_type" : "Rearrangement", "data_files" : ["vdjserver.tsv"] }
            ]
        },
        {
            "repository": "VDJbase",
            "repository_url": "https://airr-seq.vdjbase.org/airr/v1/",
            "dataset": [
              { "data_type" : "Repertoire", "data_files" : [ "vdjbase-metadata.json"] },
              { "data_type" : "Rearrangement", "data_files" : ["vdjbase.tsv"] }
            ]
        },
        {
            "repository": "COVID 19-1",
            "repository_url": "https://covid19-1.ireceptor.org/airr/v1/",
            "dataset": [
              { "data_type" : "Repertoire", "data_files" : [ "airr-covid-19-metadata.json"] },
              { "data_type" : "Rearrangement", "data_files" : ["airr-covid-19.tsv"] }
            ]
        }
    ]
}
```

**Comment by bcorrie** [comment]
And a more complex download from a single repository that contains many types of AIRR Data (a download of Cell/Gex data from the iReceptor Gateway) would look like this:
```
{
    "Info": {
        "title": "AIRR Manifest",
        "version": "3.0",
        "description": "List of files for each repository",
        "contact": {
            "name": "iReceptor Gateway",
            "url": "https://gateway-staging.ireceptor.org",
            "email": "support@ireceptor.org"
        }
    },
    "DataSets": [
        {
            "dataset_name": "IPA 6",
            "repository_url": "https://ipa6.ireceptor.org/airr/v1/",
            "dataset": [
              { "data_type" : "Repertoire", "data_files" : [ "ireceptor-public-archive-metadata.json"] },
              { "data_type" : "Rearrangement", "data_files" : ["ireceptor-public-archive-rearrangement.tsv"] },
              { "data_type" : "Cell", "data_files" : ["ireceptor-public-archive-cell.json"] },
              { "data_type" : "CellExpression", "data_files" : ["ireceptor-public-archive-gex.json"] }
            ]
        }
    ]
}
```

**Comment by schristley** [comment]
Hi @bcorrie , I'm good with the information in the manifest, but one thing I don't like is the openness of  `data_type`. I know the description says what it should be but nothing is enforcing it, and as I expect this manifest will be most useful for automated machine usage. One way is to make `data_type` an enum or controlled vocabulary, but what do you think about this as alternative for `dataset`? This is like `DataFile` in that the keys are the types.

```
{
    "Info": {
        "title": "AIRR Manifest",
        "version": "3.0",
        "description": "List of files for each repository",
        "contact": {
            "name": "iReceptor Gateway",
            "url": "https://gateway-staging.ireceptor.org",
            "email": "support@ireceptor.org"
        }
    },
    "DataSets": [
        {
            "dataset_name": "IPA 6",
            "repository_url": "https://ipa6.ireceptor.org/airr/v1/",
            "dataset": {
              "Repertoire" : [ "ireceptor-public-archive-metadata.json"],
              "Rearrangement" : ["ireceptor-public-archive-rearrangement.tsv"],
              "Cell" : ["ireceptor-public-archive-cell.json"],
              "CellExpression" : ["ireceptor-public-archive-gex.json"]
            }
        }
    ]
}
```

**Comment by bcorrie** [comment]
Yes, I had that at one point... https://github.com/airr-community/airr-standards/pull/548#issuecomment-1809406977

The issues, as listed there:
- In our schemas we use keys like "Repertoire" when we are describing the full object (like in Datafile) but this is different. These aren't the object, they are list of files that contain data. I think that is potentially confusing. Anywhere we have a `Repertoire` key in the spec, the value is always a list of `Repertoire` objects. This would break that...
- I think we want to be able to include files with non-AIRR objects in them. So having a fixed set of keys would preclude that I think, no?

I think this is still computable in the sense that you can get the expected object names that you would be able to use as a type from the AIRR Spec. It is a bit harder to validate, as you can't rely on a JSON validator to do it, and you have to check the key/value pairs to make sure the value is valid. But if we do want to be able to include other files in the manifest, then something like this is important.

If we want to make this an AIRR Object only manifest, I would agree your mechanism would be better. 

What do others think? Is this an AIRR Manifest or is this a data manifest with AIRR semantics.

**Comment by bcorrie** [comment]
BTW, here is a use case for other types of files. In the iReceptor Gateway downloads you get query files for the data files that in principle should allow you to reproduce each of the files in your data set.

```
{
    "Info": {
        "title": "AIRR Manifest",
        "version": "3.0",
        "description": "List of files for each repository",
        "contact": {
            "name": "iReceptor Gateway",
            "url": "https://gateway.ireceptor.org",
            "email": "support@ireceptor.org"
        }
    },
    "DataSets": [
        {
            "dataset_name": "VDJServer",
            "repository_url": "https://vdjserver.org/airr/v1/",
            "dataset": [
              { "data_type" : "Repertoire", "data_files" : [ "vdjserver-metadata.json"] },
              { "data_type" : "Rearrangement", "data_files" : ["vdjserver.tsv"] },
              { "data_type" : "RepertoireQuery", "data_files" : ["vdjserver_repertoire_query.json"] },
              { "data_type" : "RearrangementQuery", "data_files" : ["vdjserver_rearrangement_query.json"] }
            ]
        }
]
}
```
This set of files is actually what you get when you do an iReceptor Gateway download.

You can reproduce `vdjserver-metadata.json` with

```
curl -d @vdjserver_repertoire_query.json https://vdjserver.org/airr/v1/repertoire
```

You can reproduce `vdjserver.tsv` with:

```
curl -d @vdjserver_rearrangement_query.json https://vdjserver.org/airr/v1/rearrangement
```
All of the above information is in the manifest!

From a reproducibility perspective, this is quite important. You know exactly where the data came from that is in your data set, and if need be you could reproduce it. As such, the query files are pretty important aspects of the data set and in my mind would be ideally part of the manifest.

Admittedly this is a specific use case for the iReceptor Gateway, but this flexibility is pretty powerful.

**Comment by schristley** [comment]
> Yes, I had that at one point... [#548 (comment)](https://github.com/airr-community/airr-standards/pull/548#issuecomment-1809406977)

> * In our schemas we use keys like "Repertoire" when we are describing the full object (like in Datafile) but this is different. These aren't the object, they are list of files that contain data. I think that is potentially confusing. Anywhere we have a `Repertoire` key in the spec, the value is always a list of `Repertoire` objects. This would break that...

Ok, I can see that point. What about having the keys be `RepertoireFile`, `RearrangementFile`, and so on? That avoids the conflict and also says what they are.

> * I think we want to be able to include files with non-AIRR objects in them. So having a fixed set of keys would preclude that I think, no?

Nope, having a set of defined keys doesn't disallow the object to have additional keys. It's not so different from as with the `Repertoire` object that iReceptor adds some additional `ir_` fields in the ADC. At least from an AIRR perspective, our validation routines ignore extra stuff that is outside of the spec.

I also think we can be expansive with our keys, so your idea of `RepertoireQuery` and `RearrangementQuery` are good things to add.

**Comment by schristley** [comment]
> I also think we can be expansive with our keys, so your idea of `RepertoireQuery` and `RearrangementQuery` are good things to add.

I guess what I mean by this is that we don't have to restrict ourselves to just AIRR things. For example, we could add `10XCLoupeFile` and `10XVLoupeFile` for the Loupe output files from 10X's CellRanger, and so on.

In a way, I do this for VDJServer's analysis jobs. All output files must be "tagged and typed" in the metadata otherwise it is an error and they cannot be archived. This has two effects, 1) I can validate that an analysis jobs produced all of the output files that are expected and 2) downstream tools and GUI can say "I want that output file" and can look it up in the metadata and map it an actual file on disk.

**Comment by bcorrie** [comment]
> Nope, having a set of defined keys doesn't disallow the object to have additional keys. It's not so different from as with the `Repertoire` object that iReceptor adds some additional `ir_` fields in the ADC. At least from an AIRR perspective, our validation routines ignore extra stuff that is outside of the spec.
> 
> I also think we can be expansive with our keys, so your idea of `RepertoireQuery` and `RearrangementQuery` are good things to add.

If that is the case, then I wouldn't object to that. Strict rules that say these and only these keys are allowed are restrictive. Rules that say these keys are supported but other keys are allowed would work I think.

I will wait a bit for other comments before making any more changes. @scharch any thoughts/comments on where this has landed? I haven't considered this with respect to RepertoireGroup yet, maybe that can be next on my AIRR v2.0 mission 8-)

**Comment by bcorrie** [comment]
Also need to consider which fields in a `Manifest` would be `required`.

**Comment by javh** [comment]
I prefer the simplicity of @schristley's type-keyed version. But, it's not clear to me how we would add properties like explicit file types if we wanted to (eg, tsv, json, etc). 

For example, something like this:

```
"DataSets": [
    { "dataset_name": "VDJServer Thingy",
      "repository_url": "https://vdjserver.org/airr/v1/",
      "accession": "abc123",
      "data_class": "Repertoire", 
      "data_type": "json",
      "data_file": "vdjserver-metadata.json" },
    {...}]
```

I guess the question is, do we care? For the simple version, we're assuming everything that the user needs to know to read the file is within the key (eg, "Repertoire") or filename (eg, `*.json`).

**Comment by scharch** [comment]
Oh, golly, I've missed a lot here.

> To me that is what the `DataSets` in a `Manifest` are for. In the iReceptor Gateway case, if a download comes from 3 different repositories then there are 3 `DataSets` in the `Manifest` for that download, and it would look like this:

AARGH - this should be (the absolute simplest usecase of a) `RepertoireGroup`!

And individual `dataset`s should really be 1 repertoire each (as in the final example here https://github.com/airr-community/airr-standards/pull/548#issuecomment-1809430438) though I think it would be ok for them to all point the same JSON file. But the naming quickly gets tortuous, and I don't have a good idea how to solve it.

> I guess the question is, do we care?

I don't think so, but I don't see the harm in the solution @bcorrie proposed (nor do I have much preference between either version).

Otherwise, I like what is here

**Comment by javh** [comment]
I think I have a slight preference for the second option above (one entry per file), just because it seems more extensible and it would be easier to add properties specific to individual files (checksums, accessions, timestamps, whatever).

**Comment by scharch** [comment]
Is there any reason to add `contains_` tags like in `Study`?

**Comment by schristley** [comment]
I'm just catching up here myself. What I wanted to add, after going back and looking at what I did with VDJServer is that yes, you want individual files listed so that you can add properties to them. This doesn't preclude using type-keyed version, but at same time, it doesn't add much either, it mostly shifts where the arrays versus dictionaries are in the structure. For example, in the current proposal `data_files` is an array 

```
"DataSets": [
    { "name": "VDJServer",
      "repository_url": "https://vdjserver.org/airr/v1/",
      "accession": "abc123",
      "files":
      [
        {"data_type": "Repertoire", "file_type": "json", "filename": "vdjserver-metadata.json" },
        {"data_type": "Rearrangement", "file_type": "tsv", "filename": "repertoire1.tsv" },
        {"data_type": "Rearrangement", "file_type": "tsv", "filename": "repertoire2.tsv" }
      ]
     },
    {...}
]
```

while in the type-keyed version, `data_files` is an object

```
"DataSets": [
    { "name": "VDJServer",
      "repository_url": "https://vdjserver.org/airr/v1/",
      "accession": "abc123",
      "files": {
         "Repertoire": [ { "file_type": "json", "filename": "vdjserver-metadata.json" } ],
         "Rearrangement": [ { "file_type": "tsv", "filename": "repertoire1.tsv" },
                            { "file_type": "tsv", "compress": "gzip", "filename": "repertoire2.tsv.gz" } ]
       }
    },
    {...}
]
```

I would say the slight advantage to the type-keyed version is that code can quickly grab the list of files for a type because of the key (thus O(logn) complexity), while in the array version, code has to enumerate over the array entries and check the `file_type` for each entry (thus O(n) complexity).

Please do not name the field `data_type`, it is a `file_type` or `file_format`. The key is the data type. Using `filename` instead of `data_file` seems a bit more descriptive.

A `compress` attribute would be useful, as rearrangement (and other large files) would commonly be compressed. We don't want to rely upon parsing the `file_type`.

**Comment by schristley** [comment]
> > To me that is what the `DataSets` in a `Manifest` are for. In the iReceptor Gateway case, if a download comes from 3 different repositories then there are 3 `DataSets` in the `Manifest` for that download, and it would look like this:
> 
> AARGH - this should be (the absolute simplest usecase of a) `RepertoireGroup`!

@scharch I would disagree. The manifest should be limited to metadata about files without necessarily saying anything about analysis, and if we use `RepertoireGroup` for both then it can cause confusion. For example, one of the data types might be `RepertoireGroup` which references the file containing the group used for analysis, like below.

```
"DataSets": [
    { "name": "VDJServer",
      "repository_url": "https://vdjserver.org/airr/v1/",
      "accession": "abc123",
      "files": {
         "Repertoire": [ { "file_type": "json", "filename": "vdjserver-metadata.json" } ],
         "Rearrangement": [ { "file_type": "tsv", "filename": "repertoire1.tsv" },
                            { "file_type": "tsv", "compress": "gzip", "filename": "repertoire2.tsv.gz" } ],
         "RepertoireGroup": [ { "file_type": "json", "filename": "group-metadata.json" } ]
       }
    },
    {...}
]
```

The manifest isn't meant to be a provenance about how analysis was done, or should be done, on the files. That falls more in the domain of `DataProcessing`

**Comment by scharch** [comment]
> if we use `RepertoireGroup` for both then it can cause confusion.

OK, fair enough. I guess when use iReceptor, it is to download a dataset I want to immediately start analyzing (ie a `RepertoireGroup`) and failed to appreciate broader use cases.

**Comment by schristley** [comment]
> > if we use `RepertoireGroup` for both then it can cause confusion.
> 
> OK, fair enough. I guess when use iReceptor, it is to download a dataset I want to immediately start analyzing (ie a `RepertoireGroup`) and failed to appreciate broader use cases.

While @bcorrie shows using the manifest to describe an iReceptor download for ADC (which is capability outside of the ADC API spec), I can see doing the same thing for VDJServer download for ADC as VDJServer provides the similar capability to download a whole dataset. However, that shouldn't be the sole purpose, nor even the primary purpose in my opinion.

The primary purpose should be that the manifest can be provided to an analysis tool/workflow and it can perform some analysis on that dataset without you having to do anything extra. In VDJServer's analysis, I'm using something that is very similar to this manifest. It has much the same info, the repertoire metadata file to use, the list of rearrangement files, the file describing the repertoire groups, and so on. It also has some extra stuff like what germline set to use and other things related to the process flow of the analysis. So if AIRR adopts a manifest, I very much expect to alter VDJServer's custom file to shift to AIRR.

Now the current design with how @bcorrie describes for an iReceptor download does generate some ambiguity about datasets. He is creating a dataset for each ADC repository, but you as the user might not care or know about those, you might be thinking about all of that data as a single dataset, because it is a single query in your mind. So if you input the manifest into an analysis tool, how does it handle those datasets, does it process them separately, merge them, something else? Do you merge those datasets together before running analysis? We could certainly provide functionality with `airr-tools` to do that for you automatically (in theory all IDs and filenames are unique so the merge is simple). Or we could ask ADC GUIs like iReceptor and VDJServer to also always provide an (additional) manifest file that has everything combined.

**Comment by javh** [comment]
> Please do not name the field `data_type`, it is a `file_type` or `file_format`. The key is the data type. Using `filename` instead of `data_file` seems a bit more descriptive.

Agreed. I like concise, so stuff like `format`, `filename`, `class`, `compression`.

**Comment by scharch** [comment]
> The primary purpose should be that the manifest can be provided to an analysis tool/workflow and it can perform some analysis on that dataset without you having to do anything extra.

That sounds suspiciously like a `DataProcessing`...

To me, the primary purpose of `Manifest` has always been about ["_this_ Rearrangements file(s) goes with _that_ Cell file and those Clones"](https://github.com/airr-community/airr-standards/issues/439#issuecomment-657265108) or
> 2\. A way of grouping things on disk so that you know which entities belong together for analysis (`Manifest`)
(https://github.com/airr-community/airr-standards/issues/439#issuecomment-1937820941)

Or the first sentence of #426 is "We need a way to link related files together on the file system."



....ok, ok, I'll leave the horse alone

**Comment by schristley** [comment]
> > The primary purpose should be that the manifest can be provided to an analysis tool/workflow and it can perform some analysis on that dataset without you having to do anything extra.
> 
> That sounds suspiciously like a `DataProcessing`...

I worded it poorly. I don't mean to imply the manifest says "what" analysis to do nor "how" to do that analysis. Just if you are doing an analysis, the tool knowing 

> "_this_ Rearrangements file(s) goes with _that_ Cell file and those Clones"

makes it easier to do the "how" and "what" in an automated fashion.

**Comment by scharch** [comment]
> Now the current design with how @bcorrie describes for an iReceptor download does generate some ambiguity about datasets. He is creating a dataset for each ADC repository, but you as the user might not care or know about those, you might be thinking about all of that data as a single dataset, because it is a single query in your mind.

I agree, but @bcorrie said it had to be this way because `repertoire_id`s are not guaranteed to be unique across repositories. If there's a way to solve that with backend `airr-tools` magic, I'm all in favor!

**Comment by bcorrie** [comment]
> Now the current design with how @bcorrie describes for an iReceptor download does generate some ambiguity about datasets. He is creating a dataset for each ADC repository,

I don't think there is any ambiguity in this at all. There are a bunch of files in an iReceptor Gateway download, and the Manifest that accompanies those files describe how they are related to each other. To me, that is all a manifest should be doing.

The iReceptor Gateway can decide to split the files into a `Dataset` per `Repertoire`, it can combine all data from all repositories into a single `Dataset`, or it can provide a `Dataset` per repository. Each of these options are reasonable, each is possible, and each would have a very different `Manifest`. 

The `Manifest` doesn't imply anything about how those files should be used. All it states is that a group of files are related to each other.

> but you as the user might not care or know about those, you might be thinking about all of that data as a single dataset, because it is a single query in your mind.

Exactly, that is up to the user to decide how to do that given the `Datasets` that they are given.

**Comment by schristley** [comment]
> > Now the current design with how @bcorrie describes for an iReceptor download does generate some ambiguity about datasets. He is creating a dataset for each ADC repository, but you as the user might not care or know about those, you might be thinking about all of that data as a single dataset, because it is a single query in your mind.
> 
> I agree, but @bcorrie said it had to be this way because `repertoire_id`s are not guaranteed to be unique across repositories. If there's a way to solve that with backend `airr-tools` magic, I'm all in favor!

That's not AIRR-compliant. The ADC API, both the paper and documentation, state that `repertoire_id` is globally unique, for this very reason! So that data can be combined. @bcorrie has stated this is true for IPA's newer repositories but some of the old original repositories are not compliant. That will be resolved over time as part of the AKC project, whether that means data is reloaded with better IDs or whether the service is modified to add a simple global CURIE (like `ipa1:`) is added to the ids when data is exported, or something else. It should be considered a temporary issue and not driving requirement for the manifest.

**Comment by bcorrie** [comment]
> I agree, but @bcorrie said it had to be this way because `repertoire_id`s are not guaranteed to be unique across repositories. If there's a way to solve that with backend `airr-tools` magic, I'm all in favor!

Nah, I have given up on that one... We choose to do a `Dataset` per repository because it is the cleanest way for us to present data downloaded from the Gateway.

It is an interesting question though. What uniqueness constraints do we have on data that is described as a `Dataset` in a `Manifest`. If, in the ADC context, if IDs are global (which I have come to terms with for the most part - shudder), it doesn't matter for the ADC. 

I do think that we need to assume that `repertoire_ids` are unique within a `Dataset` - I think we have to be able to assume that a `repertoire_id` from a `repertoire.json` file in the same `Dataset` as a group of `rearrangement.tsv` files in that dataset will uniquely identify the Rearrangements from that `reperotire_id` across all of those TSV files.

**Comment by bcorrie** [comment]
> @bcorrie has stated this is true for IPA's newer repositories but some of the old original repositories are not compliant. That will be resolved over time as part of the AKC project, whether that means data is reloaded with better IDs or whether the service is modified to add a simple global CURIE (like `ipa1:`) is added to the ids when data is exported, or something else. It should be considered a temporary issue and not driving requirement for the manifest.

Yep, agree... @bcorrie has seen the light 8-)

**Comment by schristley** [comment]
> > Now the current design with how @bcorrie describes for an iReceptor download does generate some ambiguity about datasets. He is creating a dataset for each ADC repository,
> 
> I don't think there is any ambiguity in this at all.

Poor wording or incomplete description on my part. I don't mean ambiguity about the Manifest or how it is created. I mean ambiguity in how a tool might interpret it for a particular use.

> The `Manifest` doesn't imply anything about how those files should be used. All it states is that a group of files are related to each other.

Exactly, and thus the ambiguity about how to use. I'm not imply that Manifest should say anything about that, on the contrary. Because the manifest can be created in different ways and still be valid as a manifest, tools cannot automatically make assumptions about it.

**Comment by schristley** [comment]
> I do think that we need to assume that `repertoire_ids` are unique within a `Dataset` - I think we have to be able to assume that a `repertoire_id` from a `repertoire.json` file in the same `Dataset` as a group of `rearrangement.tsv` files in that dataset will uniquely identify the Rearrangements from that `reperotire_id` across all of those TSV files.

Yes, and also that the IDs are internally consistent and resolvable. That is the `Rearrangement` in a dataset should not be using a `repertoire_id` from another dataset, and so on for all of the other AIRR Data Model identifiers. Obviously, this doesn't apply to external ID like Pubmed and so on.

**Comment by scharch** [comment]
> It should be considered a temporary issue and not driving requirement for the manifest.

In that case, is there a need for a `DataSets` array? It seems much cleaner if `Manifest` and "group of related files" are 1:1.

In other words, instead of 
```
Manifest: {
    "DataSets": [
        { "name": "VDJServer",
          "repository_url": "https://vdjserver.org/airr/v1/",
          "accession": "abc123",
          "files": {
             "Repertoire": [ { "file_type": "json", "filename": "vdjserver-metadata.json" } ],
             "Rearrangement": [ { "file_type": "tsv", "filename": "repertoire1.tsv" },
                                { "file_type": "tsv", "compress": "gzip", "filename": "repertoire2.tsv.gz" } ],
             "RepertoireGroup": [ { "file_type": "json", "filename": "group-metadata.json" } ]
           }
        },
        {...}
    ]
}
```

Why not just
```
Manifest: {
    "Repertoire": [ { "file_type": "json", "filename": "vdjserver-metadata.json" } ],
    "Rearrangement": [ { "file_type": "tsv", "filename": "repertoire1.tsv" },
                       { "file_type": "tsv", "compress": "gzip", "filename": "repertoire2.tsv.gz" } ],
    "RepertoireGroup": [ { "file_type": "json", "filename": "group-metadata.json" } ]
}
```

**Comment by bcorrie** [comment]
I think it is important that a `Manifest` be able to describe multiple `Datasets`. Without this, even for the simply use case of an iReceptor download, I would need a `Manifest` to describe how my `Datasets` were related (e.g. these are the manfiests in your download). I don't see a huge downside of having the array, and a significant upside.

**Comment by scharch** [comment]
@bcorrie I think I am not understanding what a `Dataset` is, then. Why don't all of the files in a download "go" together?

**Comment by bcorrie** [comment]
A `Dataset` is a group of files that go together. A `Manifest` describes a group of a `Datasets`. A `Manifest` can capture different groupings of files that describe multiple `Datasets` for a variety of purposes.

**Comment by bcorrie** [comment]
Let me put it this way, if I wanted to split the data downloaded into a `Dataset` per `Repertoire` and there were 1000 Repertoires in the download, I would need 1000 manifest files 8-)

**Comment by scharch** [comment]
> A `Dataset` is a group of files that go together. A `Manifest` describes a group of a `Datasets`. A `Manifest` can capture different groupings of files that describe multiple `Datasets` for a variety of purposes.

I still don't get this. Why isn't `Manifest` a "group of files that go together"? I thought that was the whole point (cf my dead-horse-beating above)

**Comment by scharch** [comment]
> Let me put it this way, if I wanted to split the data downloaded into a `Dataset` per `Repertoire` and there were 1000 Repertoires in the download, I would need 1000 manifest files 8-)

No, it would be one `Manifest` with an array of 1000 files in the repertoire field....???

**Comment by schristley** [comment]
> Why not just
> 
> ```
> Manifest: {
>     "Repertoire": [ { "file_type": "json", "filename": "vdjserver-metadata.json" } ],
>     "Rearrangement": [ { "file_type": "tsv", "filename": "repertoire1.tsv" },
>                        { "file_type": "tsv", "compress": "gzip", "filename": "repertoire2.tsv.gz" } ],
>     "RepertoireGroup": [ { "file_type": "json", "filename": "group-metadata.json" } ]
> }
> ```

It's not that this is wrong. I would just argue that allowing multiple `DataSets` provide flexibility for the user to group together files according to how that user would like to group them (assuming the interlinking of the AIRR Data Model identifiers is valid). The above structure forces all of the files to be combined together.

Here's an example. Let's say I'm going to do a big meta-analysis, and I got 10 studies with data I'm going to include. I think it would be nice to organize them so that the files stay grouped with each study, something like this:

```
Manifest: {
    "DataSets": [
        { "name": "study 1", ... },
        { "name": "study 2", ... },
        { "name": "study 3", ... },
        { "name": "study 4", ... },
        { "name": "study 5", ... },
        {...}
    ]
}
```

Instead of being all merged together. If a few months down the road it's decided, actually we are going to exclude one of these studies, then it's easy to chop it out of the manifest.

Another way to think about it is to change the definition slightly. Define a `Dataset` to be a "group of files that go together" where all of the AIRR Data Model identifiers are internally consistent and resolvable inside the `Dataset`. Then define a `Manifest` to be a "group of datasets that go together". That extra level provides useful flexibility.

**Comment by scharch** [comment]
Yeah, @bcorrie explained it to me on the call this week as wanting to know which `Repertoire` metadata goes with which `Rearrangement`s tsv without having to open all the files in the `Manifest` and checking the ids. So that makes sense to me now.

The new question I have (but have not had time to really think about yet) is how this interacts with `RepertoireGroup`, which is potentially a _third_ level in the "things that go together" hierarchy....

**Comment by schristley** [comment]
You might say, well you don't really need that outer `Manifest` object, just do this:

```
    "DataSets": [
        { "name": "study 1", ... },
        { "name": "study 2", ... },
        { "name": "study 3", ... },
        { "name": "study 4", ... },
        { "name": "study 5", ... },
        {...}
    ]
```

but I think the idea is that there might be additional metadata or info at the `Manifest` level

```
Manifest: {
    "description": "my cool meta-analysis",
    "created": "Mar 5, 2024",
    "DataSets": [
        { "name": "study 1", ... },
        { "name": "study 2", ... },
        { "name": "study 3", ... },
        { "name": "study 4", ... },
        { "name": "study 5", ... },
        {...}
    ]
}
```

**Comment by schristley** [comment]
> The new question I have (but have not had time to really think about yet) is how this interacts with `RepertoireGroup`, which is potentially a _third_ level in the "things that go together" hierarchy....

In my meta-analysis example, I could see each study having `RepertoireGroup` in them from the original analysis for that study. One might have control and tumor groups, another might have pre- and post-treatment groups, and so on. I wouldn't expect that somehow all those groups were automatically aligned with each other, you'd have to do something to say how'd you really like the repertoires grouped together. But you have flexibility, you could 1) create new `Datasets` where each aligns with a `RepertoireGroup` that you want, or 2) insert new `RepertoireGroups` into each `Dataset` according to how you want the repertoires organized, or 3) merge it all together into a single `DataSet` with a `RepertoireGroup` for how you want. All three would be valid.

**Comment by bcorrie** [comment]
> The new question I have (but have not had time to really think about yet) is how this interacts with `RepertoireGroup`, which is potentially a _third_ level in the "things that go together" hierarchy....

The way I think about this is that `RepertoireGroup` is orthogonal (or at least independent) of `Manifest`. `Manifest` is describing files that are related to each other on disk. `RepertoireGroup` is describing how `Repertoires` are grouped together for description or analysis.

Take my iReceptor download example, one `Dataset` per repository and one `Manifest` file for the whole download. I could then have an orthogonal `ReperotireGroup` that has two lists of  `Repertoires` across all of the `Datasets` in the `Manifest` that split the data based on the analysis desired.

So if I am doing a comparison across `male` and `female` across all of the data in the download, I have two `RepertoireGroups`,  one that lists all of the `Repertoires` from `male` subjects and one that lists all of the `Repertoires` from female subjects. If I apply this to the `Datasets` described in the `Manifest`, it is trivial for me to pull out the `Rearrangements` from the `male` subjects. I just look for the `repertoire_ids` from the `male` `RepertoireGroup` in each of the `Datasets`.

If I am doing a comparison across early recovery, late recovery, and no recovery COVID subjects I have three `RepertoireGroups`.

So technically you could have one download, and a bunch of `RepertoireGroups` that described slicing the data in the download in an arbitrary number of ways. Slice by gender, slice by study_id, slice by age, slice by disease_diagnosis, etc.

**Comment by bcorrie** [comment]
In my comment here: https://github.com/airr-community/airr-standards/pull/548#issuecomment-1975581849 I have an example of how you can associate a `RepertoireGroup` file with a `Manifest` so you can link the two as per @schristley comment above.

**Comment by scharch** [comment]
> So if I am doing a comparison across `male` and `female` across all of the data in the download, I have two `RepertoireGroups`, one that lists all of the `Repertoires` from `male` subjects and one that lists all of the `Repertoires` from female subjects.

In this example I download the following files:
male_repertoires1.json, male_rearrangements1.tsv, male_repertoires2.json, male_rearrangements2.tsv, female_repertoires1.json, female_rearrangements1.tsv, female_repertoires2.json, female_rearrangements2.tsv, male_repGroup.json, female_repGroup.json, male_clones.json, female_clones.json (those last two are associated with the `RepertoireGroup`s, not the individual `Repertoire`s).

What does the `Manifest` look like?

**Comment by bcorrie** [comment]
A challenge, I like it... 8-)

> male_repGroup.json, female_repGroup.json,

Do these include the repertoires for all data in the `Manifest` or just the clone files?

**Comment by scharch** [comment]
> A challenge, I like it... 8-)
> 
> > male_repGroup.json, female_repGroup.json,
> 
> Do these include the `repertoire_ids` for all rearrangement and clone data in the `Manifest` or just the clone files?

Only the `repertoire_ids` -at least as currently set up, the `Clone`s only refer to them by `repertoire[_group]_id`.

I think male_repGroup.json should look like this:
```
"Info": { ... },
"RepertoireGroup": {
    "repertoire_group_id": "rg12345",
    "repertoire_group_name": "StudyABC_males",
    "repertoire_group_description": "Repertoires from males to be compared to females",
    "repertoires": [ {"repertoire_id": "males1", "filter": "*"}, {"repertoire_id": "males2", "filter": "*"} ]
}
````

And male_clones.json something like:
```
"Info": { ... },
"Clone": { "clone_id": "m001c001", "repertoire_group_id": "rg12345", [stuff deleted] },
"Clone": { "clone_id": "m001c002", "repertoire_group_id": "rg12345", [stuff deleted] },
"Clone": { "clone_id": "m001c003", "repertoire_group_id": "rg12345", [stuff deleted] },
{ more clones... }
```

**Comment by bcorrie** [comment]
> And male_clones.json something like:
> 
> ```
> "Info": { ... },
> "Clone": { "clone_id": "m001c001", "repertoire_group_id": "rg12345", [stuff deleted] },
> "Clone": { "clone_id": "m001c002", "repertoire_group_id": "rg12345", [stuff deleted] },
> "Clone": { "clone_id": "m001c003", "repertoire_group_id": "rg12345", [stuff deleted] },
> { more clones... }
> ```

I think I am confused - the `Clone` object currently does not refer to `RepertoireGroups` and there is no `repertoire_group_id`. Are you suggesting that such a thing would be added? Our current `Clone` definition says:

"A unique inferred clone object that has been constructed within a single data processing for a single repertoire and a subset of its sequences and/or rearrangements."

**Comment by scharch** [comment]
@bcorrie yes. Sorry, I was sure I had raised this before, but apparently it was only in my head. `Clone`s will have to be defined on a `RepertoireGroup` instead of a `Repertoire` because it is quite common to calculate clonality across many samples from an individual (whether or not that is a time series). In the example we are working on here, it would be ok (and simpler!) to use the old (current) way of doing things, but if `RepertoireGroup` is supposed to be the basic unit of analysis then that's how we should be designing complementary bits like `Manifest`.

**Comment by bcorrie** [comment]
I would think something like this, based on your description...

```
Manifest: {
    "DataSets": [
        { "name": "male1",
          "files": {
             "Repertoire": [ { "file_type": "json", "filename": "male_repertoires1.json" } ],
             "Rearrangement": [ { "file_type": "tsv", "filename": "male_rearrangements1.tsv" "],
           }
        },
        { "name": "male2",
          "files": {
             "Repertoire": [ { "file_type": "json", "filename": "male_repertoires2.json" } ],
             "Rearrangement": [ { "file_type": "tsv", "filename": "male_rearrangements2.tsv" "],
           }
        },
        { "name": "female1",
          "files": {
             "Repertoire": [ { "file_type": "json", "filename": "female_repertoires1.json" } ],
             "Rearrangement": [ { "file_type": "tsv", "filename": "female_rearrangements1.tsv" "],
           }
        },
        { "name": "female2",
          "files": {
             "Repertoire": [ { "file_type": "json", "filename": "female_repertoires2.json" } ],
             "Rearrangement": [ { "file_type": "tsv", "filename": "female_rearrangements2.tsv" "],
           }
        },
        { "name": "male-clones",
          "files": {
             "RepertoireGroup": [ { "file_type": "json", "filename": "male_repGroup.json" } ],
             "Clone": [ { "file_type": "json", "filename": "male_clones.json" "],
           }
        },
        { "name": "female-clones",
          "files": {
             "RepertoireGroup": [ { "file_type": "json", "filename": "female_repGroup.json" } ],
             "Clone": [ { "file_type": "json", "filename": "female_clones.json" "],
           }
        },
        {...}
    ]
}
```

**Comment by scharch** [comment]
> I would think something like this, based on your description...

Ok, yes, at first glance this makes sense to me.
Do we need an "easy" way to distinguish which `Dataset`s are `Repertoire`s and which are `RepertoireGroup`s? Or is looking at the keys within sufficient?

**Comment by schristley** [comment]
You could also do something a little more compact:

```
Manifest: {
    "DataSets": [
        { "name": "male vs female study",
          "files": {
             "Repertoire": [ { "file_type": "json", "filename": "male_repertoires1.json" },
                             { "file_type": "json", "filename": "male_repertoires2.json" },
                             { "file_type": "json", "filename": "female_repertoires1.json" },
                             { "file_type": "json", "filename": "female_repertoires2.json" } ],
             "Rearrangement": [ { "file_type": "tsv", "filename": "male_rearrangements1.tsv" },
                                { "file_type": "tsv", "filename": "male_rearrangements2.tsv" },
                                { "file_type": "tsv", "filename": "female_rearrangements1.tsv" },
                                { "file_type": "tsv", "filename": "female_rearrangements2.tsv" } ],
             "RepertoireGroup": [ { "file_type": "json", "filename": "male_repGroup.json" },
                                  { "file_type": "json", "filename": "female_repGroup.json" } ],
             "Clone": [ { "file_type": "json", "filename": "male_clones.json" },
                        { "file_type": "json", "filename": "female_clones.json" } ],
           }
        },
        {...}
    ]
}
```

And that's given all the separate JSON files that you mention. I generally keep that metadata combined together so it's more like this:

```
Manifest: {
    "DataSets": [
        { "name": "male vs female study",
          "files": {
             "Repertoire": [ { "file_type": "json", "filename": "repertoires.json" } ],
             "Rearrangement": [ { "file_type": "tsv", "filename": "male_rearrangements1.tsv" },
                                { "file_type": "tsv", "filename": "male_rearrangements2.tsv" },
                                { "file_type": "tsv", "filename": "female_rearrangements1.tsv" },
                                { "file_type": "tsv", "filename": "female_rearrangements2.tsv" } ],
             "RepertoireGroup": [ { "file_type": "json", "filename": "repGroup.json" } ],
             "Clone": [ { "file_type": "json", "filename": "clones.json" } ]
           }
        },
        {...}
    ]
}
```

**Comment by scharch** [comment]
> You could also do something a little more compact:

I think both of your examples should be equally valid, but the more verbose way allows direct linkage of the the metadata in repertoire1.json to the the rearrangements in rearrangements1.tsv.

**Comment by bcorrie** [comment]
I agree, one of the key things is that the `Manifest` support all of the above. They are different ways of storing the same data on disk using different file structures. None are "wrong"...

**Comment by bcorrie** [comment]
If I was describing a male vs female study I would lean towards:
```
Manifest: {
    "DataSets": [
        { "name": "male vs female study",
          "files": {
             "Repertoire": [ { "file_type": "json", "filename": "repertoires.json" } ],
             "Rearrangement": [ { "file_type": "tsv", "filename": "rearrangements.tsv" } ],
             "Clone": [ { "file_type": "json", "filename": "clones.json" } ]
             "RepertoireGroup": [ { "file_type": "json", "filename": "maleRepGroup.json" },{ "file_type": "json", "filename": "femaleRepGroup.json" } ],
           }
        },
        {...}
    ]
}
```

**Comment by bcorrie** [comment]
If I wanted to keep the separate data sets:
```
Manifest: {
    "DataSets": [
        { "name": "male vs female study - dataset 1",
          "files": {
             "Repertoire": [ { "file_type": "json", "filename": "male_repertoires1.json" } ],
             "Rearrangement": [ { "file_type": "tsv", "filename": "male_rearrangements1.tsv" } ],
             "Clone": [ { "file_type": "json", "filename": "male_clones1.json" } ]
             "RepertoireGroup": [ { "file_type": "json", "filename": "maleRepGroup.json" },{ "file_type": "json", "filename": "femaleRepGroup.json" } ],
           }
        },
        { "name": "male vs female study - dataset 2",
          "files": {
             "Repertoire": [ { "file_type": "json", "filename": "male_repertoires2.json" } ],
             "Rearrangement": [ { "file_type": "tsv", "filename": "male_rearrangements2.tsv" } ],
             "Clone": [ { "file_type": "json", "filename": "male_clones2.json" } ]
             "RepertoireGroup": [ { "file_type": "json", "filename": "maleRepGroup.json" },{ "file_type": "json", "filename": "femaleRepGroup.json" } ],
           }
        },
        {...}
    ]
}
```

**Comment by bcorrie** [comment]
I made a first attempt to create a spec (airr-schema.yaml only) that reflects the structure that we have been talking in the last dozen or so comments. 

The main issue I have is I don't know how to specify the set of keys that are allowed for each type of data in the AIRR schema. We have a list of them (Repertoire, Rearrangement, ...). We want to specify the type of object for each, so we need something in the spec for that. We also want to allow for other, custom objects so that more than just AIRR data can be part of a Manifest.

So I have done this by creating a `FileObject` and then repeating the definition for each file type. I then replicated the definition for each type of object:

https://github.com/airr-community/airr-standards/blob/cf4eac8be7792d986edcd0185cb4a40a018627ee/specs/airr-schema.yaml#L370

I am pretty sure this is not how we want to do it, but it was the best I could come up with for now.

**Comment by bcorrie** [comment]
BTW, I changed "DataSets" to "data_sets" since most of our keys are lower case.

**Comment by schristley** [comment]
> I made a first attempt to create a spec (airr-schema.yaml only) that reflects the structure that we have been talking in the last dozen or so comments.

This looks pretty good to me and how I'd expect it to be. If you look at `DataFile`, you should see a similarity, both are type-keyed, though in the case of `DataFile` the contents are the actual data while with `DataSet` it's a file object.

> The main issue I have is I don't know how to specify the set of keys that are allowed for each type of data in the AIRR schema.

I don't understand the issue. The spec itself is essentially defining the types of data.

> We also want to allow for other, custom objects so that more than just AIRR data can be part of a Manifest (I added an ADCQuery data type as an example).

This is allowed by the flexibility of the AIRR validation. Extra keys might give a warning about them not being in the schema, but they are essentially ignored. For example this manifest is valid:

```
Manifest: {
    "DataSets": [
        { "name": "male vs female study",
          "files": {
             "Repertoire": [ { "file_type": "json", "filename": "repertoires.json" } ],
             "Rearrangement": [ { "file_type": "tsv", "filename": "rearrangements.tsv" } ],
             "Clone": [ { "file_type": "json", "filename": "clones.json" } ]
             "RepertoireGroup": [ { "file_type": "json", "filename": "maleRepGroup.json" },{ "file_type": "json", "filename": "femaleRepGroup.json" } ],
             "Scott's custom type": "blah blah"
           }
        },
        {...}
    ]
}
```

**Comment by bcorrie** [comment]
> > The main issue I have is I don't know how to specify the set of keys that are allowed for each type of data in the AIRR schema.
> 
> I don't understand the issue. The spec itself is essentially defining the types of data.

I am OK to leave it as is. It just seemed cumbersome to me to have that long shopping list of types of AIRR data objects as the keys.

**Comment by scharch** [comment]
Looks good to me. Should `GermlineSet` also be in the list...?

**Comment by bcorrie** [comment]
> Looks good to me. Should `GermlineSet` also be in the list...?

I think the key remaining issue might be which objects are missing. There is also no Node, Tree, or Alignment objects... Mostly because I wasn't sure of their status (are they going to move from experimental to production or disappear? 8-)

I agree GermlineSet should be there for sure...

**Comment by scharch** [comment]
Right - fwiw my intent is for Tree and Node to come out, though obviously that's not finalized...

**Comment by javh** [comment]
> If you look at `DataFile`, you should see a similarity, both are type-keyed, though in the case of `DataFile` the contents are the actual data while with `DataSet` it's a file object.

The object names are backwards now. :)

> Right - fwiw my intent is for Tree and Node to come out, though obviously that's not finalized...

Even if they stay in, it might make sense for them to live in the same file as `Clone`, so we could dodge the need for separate keys by setting that convention if they remain.

On another topic, I pushed a few suggestions to FileObject:
* Added `checksum` and `version` fields.
* Renamed `file_type` to `format`.

Also, `Manifest` feels rather empty to me now. And I'm wondering about the typical use case.  What about having what is currently `DataSet` be the manifest and having a `ManifestGroup` object for the collections of data sets? Same number of objects, but seems more amenable to the structure where you might have one dataset per folder with a manifest in that folder.

**Comment by bcorrie** [comment]
> Also, `Manifest` feels rather empty to me now. And I'm wondering about the typical use case. What about having what is currently `DataSet` be the manifest and having a `ManifestGroup` object for the collections of data sets? Same number of objects, but seems more amenable to the structure where you might have one dataset per folder with a manifest in that folder.

If I understand correctly, this is just renaming the YAML objects, correct? 

To me, a Manifest lists a group of DataSets where the DataSet describes a set of related files. That is how we have it now.

You are suggesting a ManifestGroup lists a group of Manifests where a Manifest describes a set of related files.

To me the current naming logic makes the most sense but it is just object naming so I could be convinced otherwise...

**Comment by bcorrie** [comment]
> On another topic, I pushed a few suggestions to FileObject:
> 
> * Added `checksum` and `version` fields.
> * Renamed `file_type` to `format`.

Works for me...

**Comment by javh** [comment]
> If I understand correctly, this is just renaming the YAML objects, correct?

It would also mean changing the recommended use (assuming I understand the current intended use) such that the primary usage will be a standalone `Manifest` (`DataSet`) that is not nested in a `ManifestGroup` (analogous to `Repertoire` and `RepertoireGroup`).

How it's setup now implies to me that you will always have groups (arrays) of DataSets even if the group size is 1.

**Comment by bcorrie** [comment]
Personally, I would prefer to not have two different entities that I have to deal with. In your scenario users producing data and code consuming data would have to deal with both `ManifestGroup` and `Manifest`. In the code they would need to be handled differently. For users producing data, the user would be confused as to what they should produce.

It would be equally correct and valid to represent a single "data set" as a `ManifestGroup` with one `Manifest`  or just as a `Manifest`. That is confusing to me...

In the current case there is only on thing that one deals with when talking about data on disk, it is a `Manifest`. If you are describing files with a single `DataSet` the `Manifest` has a data set list of size 1. No confusion.

**Comment by bcorrie** [comment]
This is exactly what our Analysis Pipeline on the iReceptor Gateway does (based on our current implementation of `Manifest`. It takes a Manifest that is a download from a repository (see below) and processes it as follows:

```
{
    "Info": {
        "title": "AIRR Manifest",
        "version": "3.0",
        "description": "List of files for each repository",
        "contact": {
            "name": "iReceptor Gateway",
            "url": "https://gateway.ireceptor.org",
            "email": "support@ireceptor.org"
        }
    },
    "DataSets": [
        {
            "repository": "COVID 19-1",
            "repository_url": "https://covid19-1.ireceptor.org/airr/v1/",
            "repertoire_file": "airr-covid-19-metadata.json",
            "rearrangement_file": "airr-covid-19-rearrangement.tsv",
            "cell_file": "airr-covid-19-cell.json",
            "expression_file": "airr-covid-19-gex.json"
        }
    ]
}
```
It basically searches `airr-covid-19-metadata.json` for all repertoires in the `DataSet` and splits the data into a directory per repertoire. Ideally I would be able to use `RepetoireGroup` for this rather than search a file for all of the `Repertoires`.

In each repertoire directory, I have a `DataSet` that represents a single `Repertoire` with a Manifest that looks like this:

```
{"Info":{},"DataSets":[{
"cell_file":["64d2a2028ab142c7479d0ee1-cell.json"],
"expression_file":["64d2a2028ab142c7479d0ee1-gex.h5ad"],
"rearrangement_file":["64d2a2028ab142c7479d0ee1-rearrangement.tsv"]
}]}
```
Note I am no longer using the AIRR GEX/JSON files, I am using a custom file type that is an H5AD file that is generated when I split out the GEX/JSON by repertoire.

All of the processing is guided by the Manifest file, and in each case there is one `DataSet` per `Manifest`. This is trivial to handle and makes data processing quite simple.

scharch:
> Personally, I would prefer to not have two different entities that I have to deal with. In your scenario users producing data and code consuming data would have to deal with both `ManifestGroup` and `Manifest`. In the code they would need to be handled differently. For users producing data, the user would be confused as to what they should produce.

I ...agree with @bcorrie?? =P


> you will always have groups (arrays) of DataSets even if the group size is 1.

This is already the case for other parts of the schema, like `sample` and `data_processing` within `Repertoire` or `diagnosis` within `Subject`. I assume that the most common use for all of those is an array of length 1...

javh:
Good points. Thanks, @bcorrie and @scharch. I'm sold on leaving `Manifest` and `DataSet` as is.

schristley:
To add another point, we likely want to be able to validate that a `Manifest` is correct, so I think that involves adding an additional parameter to `airr-tools` to load the manifest and run it through `validate_object`

schristley:
should there be a `load_manifest` function, or set of functions, in `interface.py`?

schristley:
> > If you look at `DataFile`, you should see a similarity, both are type-keyed, though in the case of `DataFile` the contents are the actual data while with `DataSet` it's a file object.
> 
> The object names are backwards now. :)

`FileSets` ? bleh ;-)

javh:
From the call:
* Add example to `FileObject.compression` (zip, gz, bz2).
* Add `GermlineSet` and `GenotypeSet` to `DataSet`.
* Then merge.
* Handle validation routines for `Manfiest` in separate PR.

PR #550, opened 2021-09-16
ADC API V2.1 extensions

schristley:
- [ ] add `distinct` operator for /repertoire and maybe others
- [ ] add `isa` operator for /repertoire and maybe others

### DIFF
--- a/specs/adc-api-v2-openapi3.yaml
+++ b/specs/adc-api-v2-openapi3.yaml
@@ -1,0 +1,599 @@
+openapi: '3.0.0'
+
+info:
+  title: AIRR Data Commons API
+  version: 2.0.0
+  contact:
+    name: AIRR Community
+    url: 'http://www.airr-community.org/'
+    email: join@airr-community.org
+  description: >
+    Major Version 2 of the Adaptive Immune Receptor Repertoire (AIRR)
+    data repository web service application programming interface (API).
+  license:
+    name: Creative Commons Attribution 4.0 International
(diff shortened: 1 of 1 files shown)


bcorrie:
@schristley can you elaborate a bit on what you see these operators doing.

I believe I understand `isa` but I also think it only has limited applicability for AIRR `ontology` fields. Would that be correct?

Can you elaborate on `distinct`?

schristley:
> @schristley can you elaborate a bit on what you see these operators doing.
> 
> I believe I understand `isa` but I also think it only has limited applicability for AIRR `ontology` fields. Would that be correct?

Correct, only applies to fields that are ontology-backed otherwise is a query error.

> Can you elaborate on `distinct`?

Thinking like the distinct operator in SQL or Mongo, which collapses the values to uniques and eliminates duplicates. I guess this wouldn't be an operator like `=` which is in the filter but would be more like `facets`. So for example, something like this on the /repertoire endpoint would return all of the unique subjects.

` { "distinct": ["repertoire_id", "subject.subject_id"] }`

or this which would return all of the unique diagnoses

` { "distinct": ["subject.diagnosis.diseases_diagnosis.id"] }`

As with facets, you could still have a query filter to restrict to a subset of data.

Would you dare to do something like this on the /rearrangement endpoint?  ;-D

` { "distinct": ["junction_aa"] }`

bcorrie:
Changed this to v2.1 change to the ADC API

PR #551, opened 2021-09-16
physical quantity format

### LINKED ISSUE #498: Revisit `physical quantity`
In the Schema, the `format` attribute is defined as:
```yaml
Attributes:
        format:
            type: string
            description: Field format. If null then assume the full range of the field data type
            enum:
                - ontology
                - controlled vocabulary
                - physical quantity
```
But `physical quantity` is only used in the whole schema as a property in `disease_length` (where it should be an ontology, see #500). besides that, there is only another use of the term as part of the ontology label of `template_amount_unit` : 
```yaml
template_amount_unit:
...
                ontology:
                    draft: false
                    top_node:
                        id: UO:0000002
                        label:  physical quantity 
```
Where the actual label of `UO:0000002`  in the UO ontology is `mass unit`.

a) Is `physical quantity` a format itself or should it be considered  as part of the UO `ontology` ?
b) and should the label of the `template_amount_unit`ontology be `mass unit`?

### LINKED ISSUE #500: Time unit for `disease_length`
Should we add a time unit property as we did for `collection_time_point_relative` ( #420)  and `age` (#192)?

```yaml
        disease_length:
            type: number
            description: Time duration between initial diagnosis and current intervention
            title: Length of disease
            example: 23
            x-airr:
                miairr: important
                nullable: true
                adc-query-support: true
                set: 1
                subset: diagnosis and intervention
                name: Length of disease
        disease_length_unit:
            $ref: '#/Ontology'
            description: Time unit of disease length
            title:  Length of disease time unit
            example:
                id: UO:0000035
                label: month
            x-airr:
                miairr: important
                nullable: true
                adc-query-support: true
                set: 1
                subset: subject
                name: Time unit of disease length
                format: ontology
                ontology:
                    draft: false
                    top_node:
                        id: UO:0000003
                        label: time unit
```

### LINKED ISSUE #504: did we make a mistake with our conversion from physical quantity format to multiple fields?
After thinking about @franasa recent issues, I wonder if we made a mistake. That is, before with the `physical quantity` format, the data was in a single field, but now it's spread across multiple fields and we've "lost" (or are lacking) any association between those fields. For example, programs have to "know" that if they are working with `collection_time_point_relative`, that it has to also work with `collection_time_point_relative_unit`, but there is nothing in the schema that indicates those fields are tied together.

This started with `age` when we split it out into 3 fields, and maybe that seemed okay at the time because it was a one-off, but we have a few of these physical quantity fields.

Should we instead have kept the fields "grouped" together into an object, like so:

```
age: { min: 10, max: 10, unit: { id: UO:0000036, label: year } }
collection_time_point_relative: { quantity: 12, unit: { id: UO:0000033, label: day } }
template_amount: { quantity: 12, unit: { id: UO:0000024, label: nanogram } }
```

It still then is "one field" from a MiAIRR perspective, and you can see for example that `template_amount` isn't being documented properly in the [data table in the docs](https://docs.airr-community.org/en/stable/miairr/data_elements.html).

schristley:
fix #504 
fix #500 
fix #498

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -2280,7 +2280,6 @@ Sample:
         - anatomic_site
         - disease_state_sample
         - collection_time_point_relative
-        - collection_time_point_reference
         - biomaterial_provider
     properties:
         sample_id:
@@ -2356,9 +2355,10 @@ Sample:
                 name: Disease state of sample
         collection_time_point_relative:
             $ref: '#/TimePoint'
-            description: Time point at which sample was taken, relative to `Collection time event`
+            description: Time point at which sample was taken, relative to `label` event
             title: Sample collection time
             example:
+                label: Primary vaccination
                 value: 14
                 unit:
                     id: UO:0000033
@@ -2382,15 +2382,12 @@ Sample:
         collection_time_point_reference:
             type: string
             description: Event in the study schedule to which `Sample collection time` relates to
-            title: Collection time event
-            example: Primary vaccination
             x-airr:
-                miairr: important
                 nullable: true
-                adc-query-support: true
-                set: 2
-                subset: sample
-                name: Collection time event
+                deprecated: true
+                deprecated-description: Field has been merged with collection_time_point_relative.
+                deprecated-replaced-by:
+                    - collection_time_point_relative
         collection_location:
             $ref: '#/Ontology'
             description: Location where the sample was taken, preferred granularity is country-level
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -561,7 +561,7 @@ TimeQuantity:
     properties:
         quantity:
             type: number
-            description: time quantity
+            description: Time quantity
             example: 30.0
             x-airr:
                 adc-query-support: true
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -554,6 +554,34 @@ PhysicalQuantity:
                         id: UO:0000002
                         label: physical quantity
 
+# A time quantity
+TimeQuantity:
+    description: A time quantity 
+    type: object
+    properties:
+        quantity:
+            type: number
+            description: time quantity
+            example: 30.0
+            x-airr:
+                adc-query-support: true
+                nullable: false
+        unit:
+            $ref: '#/Ontology'
+            description: Unit of time
+            example:
+                id: UO:0000033
+                label: day
+            x-airr:
+                nullable: false
+                adc-query-support: true
+                format: ontology
+                ontology:
+                    draft: false
+                    top_node:
+                        id: UO:0000003
+                        label: time unit
+
 
 
 # Contributor record to describe invididuals and their contribution to a data set
@@ -2191,7 +2219,7 @@ Diagnosis:
                         id: DOID:4
                         label: disease
         disease_length:
-            $ref: '#/PhysicalQuantity'
+            $ref: '#/TimeQuantity'
             description: Time duration between initial diagnosis and current intervention
             title: Length of disease
             example:
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -454,6 +454,10 @@ InfoObject:
                 url:
                     type: string
 
+#
+# General objects
+#
+
 # A time point
 TimePoint:
     description: Time point at which an observation or other action was performed.
@@ -488,9 +492,6 @@ TimePoint:
                     top_node:
                         id: UO:0000003
                         label: time unit
-#
-# General objects
-#
 
 # A time range or interval
 TimeInterval:
@@ -518,8 +519,8 @@ TimeInterval:
                 id: UO:0000033
                 label: day
             x-airr:
-                adc-query-support: true
                 nullable: false
+                adc-query-support: true
                 format: ontology
                 ontology:
                     draft: false
@@ -547,6 +548,7 @@ PhysicalQuantity:
                 label: nanogram
             x-airr:
                 nullable: false
+                adc-query-support: true
                 format: ontology
                 ontology:
                     draft: false
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -320,6 +320,8 @@ Attributes:
                 - ontology
                 - controlled_vocabulary
                 - physical_quantity
+                - time_point
+                - time_interval
                 - CURIE
         ontology:
             type: object
@@ -487,11 +489,73 @@ TimePoint:
                     top_node:
                         id: UO:0000003
                         label: time unit
-
 #
 # General objects
-# TODO: link to global schema with JSON-LD?
 #
+
+# A time range or interval
+TimeInterval:
+    description: Time range or interval for a measurement, observation or action.
+    type: object
+    properties:
+        min:
+            type: number
+            description: Lower/minimum value of the time interval
+            example: 5.0
+            x-airr:
+                adc-query-support: true
+                nullable: true
+        max:
+            type: number
+            description: Upper/maximum value of the time interval
+            example: 10.0
+            x-airr:
+                adc-query-support: true
+                nullable: true
+        unit:
+            $ref: '#/Ontology'
+            description: Unit of the time interval
+            example:
+                id: UO:0000033
+                label: day
+            x-airr:
+                adc-query-support: true
+                nullable: true
+                format: ontology
+                ontology:
+                    draft: false
+                    top_node:
+                        id: UO:0000003
+                        label: time unit
+
+# A physical quantity
+PhysicalQuantity:
+    description: A physical quantity from a measurement or observation.
+    type: object
+    properties:
+        quantity:
+            type: number
+            description: Physical quantity
+            example: -5.0
+            x-airr:
+                adc-query-support: true
+                nullable: true
+        unit:
+            $ref: '#/Ontology'
+            description: Unit of physical quantity
+            example:
+                id: UO:0000024
+                label: nanogram
+            x-airr:
+                nullable: true
+                format: ontology
+                ontology:
+                    draft: false
+                    top_node:
+                        id: UO:0000002
+                        label: physical quantity
+
+
 
 # Contributor record to describe invididuals and their contribution to a data set
 #
@@ -1822,9 +1886,7 @@ Subject:
         - synthetic
         - species
         - sex
-        - age_min
-        - age_max
-        - age_unit
+        - age
         - age_event
         - ancestry_population
         - ethnicity
@@ -1907,52 +1969,30 @@ Subject:
                 subset: subject
                 name: Sex
                 format: controlled_vocabulary
-        age_min:
-            type: number
-            description: Specific age or lower boundary of age range.
-            title: Age minimum
-            example: 60
-            x-airr:
-                miairr: important
-                nullable: true
-                adc-query-support: true
-                set: 1
-                subset: subject
-                name: Age minimum
-        age_max:
-            type: number
+        age:
+            $ref: '#/TimeInterval'
             description: >
-                Upper boundary of age range or equal to age_min for specific age.
-                This field should only be null if age_min is null.
-            title: Age maximum
-            example: 80
+                Age of subject expressed as a time interval. If singular time point then
+                min == max in the time interval.
+            examples:
+                - min: 50
+                  max: 50
+                  unit:
+                      id: UO:0000036
+                      label: year
+                - min: 5
+                  max: 10
+                  unit:
+                      id: UO:0000034
+                      label: week
             x-airr:
-                miairr: important
                 nullable: true
+                miairr: important
                 adc-query-support: true
                 set: 1
                 subset: subject
-                name: Age maximum
-        age_unit:
-            $ref: '#/Ontology'
-            description: Unit of age range
-            title: Age unit
-            example:
-                id: UO:0000036
-                label: year
-            x-airr:
-                miairr: important
-                nullable: true
-                adc-query-support: true
-                set: 1
-                subset: subject
-                name: Age unit
-                format: ontology
-                ontology:
-                    draft: false
-                    top_node:
-                        id: UO:0000003
-                        label: time unit
+                name: Age
+                format: time_interval
         age_event:
             type: string
             description: >
@@ -1968,15 +2008,33 @@ Subject:
                 set: 1
                 subset: subject
                 name: Age event
-        age:
-            type: string
+
+        
+        
+        age_min:
+            type: number
             x-airr:
+                nullable: true
                 deprecated: true
-                deprecated-description: Split into two fields to specify as an age range.
+                deprecated-description: Combined into single age field.
                 deprecated-replaced-by:
-                    - age_min
-                    - age_max
-                    - age_unit
+                    - age
+        age_max:
+            type: number
+            x-airr:
+                nullable: true
+                deprecated: true
+                deprecated-description: Combined into single age field.
+                deprecated-replaced-by:
+                    - age
+        age_unit:
+            $ref: '#/Ontology'
+            x-airr:
+                nullable: true
+                deprecated: true
+                deprecated-description: Combined into single age field.
+                deprecated-replaced-by:
+                    - age
         ancestry_population:
             $ref: '#/Ontology'
             description: Broad geographic origin of ancestry (continent)
@@ -2134,10 +2192,14 @@ Diagnosis:
                         id: DOID:4
                         label: disease
         disease_length:
-            type: string
+            $ref: '#/PhysicalQuantity'
             description: Time duration between initial diagnosis and current intervention
             title: Length of disease
-            example: 23 months
+            example:
+                quantity: 23
+                unit:
+                    id: UO:0000035
+                    label: month
             x-airr:
                 miairr: important
                 nullable: true
@@ -2218,7 +2280,6 @@ Sample:
         - anatomic_site
         - disease_state_sample
         - collection_time_point_relative
-        - collection_time_point_relative_unit
         - collection_time_point_reference
         - biomaterial_provider
     properties:
@@ -2294,10 +2355,14 @@ Sample:
                 subset: sample
                 name: Disease state of sample
         collection_time_point_relative:
-            type: number
+            $ref: '#/TimePoint'
             description: Time point at which sample was taken, relative to `Collection time event`
             title: Sample collection time
-            example: 14
+            example:
+                value: 14
+                unit:
+                    id: UO:0000033
+                    label: day
             x-airr:
                 miairr: important
                 nullable: true
@@ -2305,26 +2370,15 @@ Sample:
                 set: 2
                 subset: sample
                 name: Sample collection time
+                format: time_point
         collection_time_point_relative_unit:
             $ref: '#/Ontology'
-            description: Unit of Sample collection time
-            title: Sample collection time unit
-            example:
-                id: UO:0000033
-                label: day
             x-airr:
-                miairr: important
                 nullable: true
-                adc-query-support: true
-                set: 2
-                subset: sample
-                name: Sample collection time unit
-                format: ontology
-                ontology:
-                    draft: false
-                    top_node:
-                        id: UO:0000003
-                        label: time unit
+                deprecated: true
+                deprecated-description: Field has been merged with collection_time_point_relative.
+                deprecated-replaced-by:
+                    - collection_time_point_relative
         collection_time_point_reference:
             type: string
             description: Event in the study schedule to which `Sample collection time` relates to
@@ -2606,7 +2660,6 @@ NucleicAcidProcessing:
         - template_class
         - template_quality
         - template_amount
-        - template_amount_unit
         - library_generation_method
         - library_generation_protocol
         - library_generation_kit_version
@@ -2643,10 +2696,14 @@ NucleicAcidProcessing:
                 subset: process (nucleic acid)
                 name: Target substrate quality
         template_amount:
-            type: number
+            $ref: '#/PhysicalQuantity'
             description: Amount of template that went into the process
             title: Template amount
-            example: 1000
+            example:
+                quantity: 1000
+                unit:
+                    id: UO:0000024
+                    label: nanogram
             x-airr:
                 miairr: important
                 nullable: true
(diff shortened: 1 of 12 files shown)


schristley:
@bussec My initial attempt just in the openapi3 schema file

schristley:
@javh `age`, that which was deprecated becomes reborn. I hope we don't need to specify a deprecation of deprecation?!

bussec:
@schristley Looks good to me, as long as the point/range distinction does not create a major hazzle for search routines.

bussec (approved):

bcorrie:
My only concern would be the same as @bussec. See #504

schristley:
The other thing I don't quite like is having the `unit` twice for a range, what if people start doing silly stuff like make one unit `year` and another `month`. Having to do conversions on queries sounds complicated... I'd much prefer a `TimeRange` object, similar to `TimePoint`, with a single unit.

schristley:
@bcorrie time interval it is, how does this look? I went with `TimeInterval` instead of `TimeRange` as I think it's more descriptive.

schristley:
as a side note, `TimeInterval` is relative, i.e. not fixed to a specific date. If we find we need to express such intervals in the future then we can create a `DateInterval` where the `min` and `max` are `datetime`

scharch:
Leaving the referenced issues open under the assumption that this will eventually be merged....

javh:
From the call:
* Verify that we still want to do this. @schristley, @bcorrie as this seems ADC-ish.
* If so, resolve conflicts and merge.

schristley:
@javh I certainly hope this can be done for AIRR 2.0. The current design of having separate, unconnected fields (except through a semantic interpretation of the names), is very TSV-ish. I think we should be more JSON-ish by compositing fields together into an object that semantically belong together.

schristley:
The check error will be fixed with #739

specs/airr-schema-openapi3.yaml
```diff
@@ -2308,37 +2366,30 @@ Sample:
                 subset: sample
                 name: Disease state of sample
         collection_time_point_relative:
-            type: number
+            $ref: '#/PhysicalQuantity'
```
bcorrie, line 2369: Shouldn't `collection_time_point' be a `TimePoint`, not a `PhysicalQuantity`

bcorrie: See #689

bcorrie:
I also think:
- `reactivity_value` and `reactivity_unit` in `ReceptorReactivity` should be a `PhysicalQuantity`, no?

bcorrie:
I think this would also resolve #689

bcorrie:
@schristley if no objections I will make the `Sample.collection_time_point` and `ReceptorReactivity.reactivity` changes.

schristley:
> @schristley if no objections I will make the `Sample.collection_time_point` and `ReceptorReactivity.reactivity` changes.

@bcorrie sure, sounds good

schristley:
Hey @bcorrie , I can see you are trying to clean up this PR, let me know if I can help in any way?

bcorrie:
Yes, I was working through it. Thanks for the test data fixes, I gave up last night and was going to dive back in this morning.  You know the test suites better than I do clearly. 

Looks like the checks have passed.

bcorrie:
Only outstanding issue on this is for CellReactivity.reactivity_value should be a PhysicalQuantity which is being covered in https://github.com/airr-community/airr-standards/pull/705#issuecomment-1930607761

So I think this can be merged. @javh and @bussec were both reviewers for this previously so I have left you on if you want to have a look.

bcorrie:
In #646 we discuss whether `collection_time_point_relative` should be a `TimeInterval`.

It also raises the question should we have a `QuantityInterval` or `QuantityRange` object. We can define it but not use it??? Or should all of our `PhyscialQuantity` objects be `QuantityRange` objects?

Not quite done with this yet. 8-)

bcorrie:
In #689 we talk about using the `label` field of `TimePoint` in `collection_time_point_relative` so that we no longer need `collection_time_point_reference`. Is this desirable?

schristley:
> In #646 we discuss whether `collection_time_point_relative` should be a `TimeInterval`.
> 
> It also raises the question should we have a `QuantityInterval` or `QuantityRange` object. We can define it but not use it??? Or should all of our `PhyscialQuantity` objects be `QuantityRange` objects?
> 
> Not quite done with this yet. 8-)

Upscale quantity fields to allow either a range or point, like with `age`...

I'm hesitant to do that until we actually have a use case that requires it. For some quantities, the idea of measuring a range might not be semantic viable. And it has the implication that downstream analysis will always have to handle the fields as ranges in order to do calculations properly, which introduces complexity.

bcorrie:
@bussec @javh I think this is OK to merge. It would be nice to have this merged before any Event stuff gets added, as Event extends/relies on some of these changes. Needs your review (or remove your self from the review list if you would prefer not to 8-)

lang/R/inst/extdata/airr-schema.yaml
```diff
@@ -478,12 +480,74 @@ TimePoint:
...
+                nullable: true
+        unit:
+            $ref: '#/Ontology'
+            description: Unit of the time interval
+            example:
+                id: UO:0000033
+                label: day
+            x-airr:
+                adc-query-support: true
+                nullable: true
```
bussec, line 514: Why is `unit` `nullable: true`? I understand that we need `min` and `max` nullable, but this should always contain a unit of time.

lang/R/inst/extdata/airr-schema.yaml
```diff
@@ -478,12 +480,74 @@ TimePoint:
...
+        unit:
+            $ref: '#/Ontology'
+            description: Unit of physical quantity
+            example:
+                id: UO:0000024
+                label: nanogram
+            x-airr:
+                nullable: true
```
bussec, line 541: Same thing as above, why is this property `nullable: true`?

lang/R/inst/extdata/airr-schema.yaml
```diff
@@ -2046,10 +2104,14 @@ Diagnosis:
                         id: DOID:4
                         label: disease
         disease_length:
-            type: string
+            $ref: '#/PhysicalQuantity'
```
bussec, line 2107: I understand that duration of time is also a physical quantity, but would it not be worthwhile to make this explicit and use `TimeInterval` instead (even if then `min==max`)?

lang/R/tests/data-tests/good_combined_airr.json
```diff
@@ -31,11 +31,13 @@
...
-                "age_min": 27,
-                "age_max": 27,
-                "age_unit": {
-                    "id": "UO_0000036",
-                    "label": "year"
+                "age": {
+                    "min": 27,
+                    "max": 27,
+                    "unit": {
+                        "id": "UO_0000036",
```
bussec, line 38: `id` should be a CURIE

lang/R/tests/data-tests/good_combined_airr.json
```diff
@@ -190,11 +184,13 @@
...
```
bussec, line 191: `id` should be a CURIE

lang/R/tests/data-tests/good_combined_airr.json
```diff
@@ -349,11 +337,13 @@
...
```
bussec, line 344: `id` should be a CURIE

lang/R/tests/data-tests/good_combined_airr.yaml
```diff
@@ -39,11 +39,12 @@ Repertoire:
...
-      age_min: 27
-      age_max: 27
-      age_unit:
-        id: UO_0000036
-        label: year
+      age:
+        min: 27
+        max: 27
+        unit:
+          id: UO_0000036
```
bussec, line 46: `id` should be a CURIE

lang/R/tests/data-tests/good_combined_airr.yaml
```diff
@@ -182,11 +177,12 @@ Repertoire:
...
```
bussec, line 184: `id` should be a CURIE

lang/R/tests/data-tests/good_combined_airr.yaml
```diff
@@ -325,11 +315,12 @@ Repertoire:
...
```
bussec, line 322: `id` should be a CURIE

lang/R/tests/data-tests/good_repertoire.yaml
```diff
@@ -29,11 +29,12 @@ Repertoire:
...
-      age_min: 27
-      age_max: 27
-      age_unit:
-        id: UO_0000036
-        label: year
+      age:
+        min: 27
+        max: 27
+        unit:
+          id: UO_0000036
```
bussec, line 36: `id` should be a CURIE

lang/R/tests/data-tests/good_repertoire.yaml
```diff
@@ -160,11 +155,12 @@ Repertoire:
...
```
bussec, line 162: `id` should be a CURIE

lang/R/tests/data-tests/good_repertoire.yaml
```diff
@@ -291,11 +281,12 @@ Repertoire:
...
```
bussec, line 288: `id` should be a CURIE

lang/python/airr/specs/airr-schema.yaml
```diff
@@ -478,12 +480,74 @@ TimePoint:
...
+                nullable: true
+        unit:
+            $ref: '#/Ontology'
+            description: Unit of the time interval
+            example:
+                id: UO:0000033
+                label: day
+            x-airr:
+                adc-query-support: true
+                nullable: true
```
bussec, line 514: see above -> `nullable` status

lang/python/airr/specs/airr-schema.yaml
```diff
@@ -478,12 +480,74 @@ TimePoint:
...
+        unit:
+            $ref: '#/Ontology'
+            description: Unit of physical quantity
+            example:
+                id: UO:0000024
+                label: nanogram
+            x-airr:
+                nullable: true
```
bussec, line 541: see above -> `nullable` status

lang/python/airr/specs/airr-schema.yaml
```diff
@@ -2046,10 +2104,14 @@ Diagnosis:
                         id: DOID:4
                         label: disease
         disease_length:
-            type: string
+            $ref: '#/PhysicalQuantity'
```
bussec, line 2107: See above -> `TimeInterval`

lang/python/tests/data/good_combined_airr.json
```diff
@@ -31,12 +31,14 @@
...
-                "age_max": 27,
-                "age_unit": {
-                    "id": "UO_0000036",
-                    "label": "year"
-                },
+                "age": {
+                    "min": 27,
+                    "max": 27,
+                    "unit": {
+                       "id": "UO_0000036",
```
bussec, line 38: `id` should be a CURIE

lang/python/tests/data/good_combined_airr.json
```diff
@@ -248,11 +241,13 @@
...
-                "age_min": 27,
-                "age_max": 27,
-                "age_unit": {
-                    "id": "UO_0000036",
-                    "label": "year"
+                "age": {
+                    "min": 27,
+                    "max": 27,
+                    "unit": {
+                       "id": "UO_0000036",
```
bussec, line 248: `id` should be a CURIE

lang/python/tests/data/good_combined_airr.json
```diff
@@ -407,11 +393,13 @@
...
```
bussec, line 400: `id` should be a CURIE

lang/python/tests/data/good_combined_airr.yaml
```diff
@@ -39,11 +39,12 @@ Repertoire:
...
-      age_min: 27
-      age_max: 27
-      age_unit:
-        id: UO_0000036
-        label: year
+      age:
+        min: 27
+        max: 27
+        unit:
+          id: UO_0000036
```
bussec, line 46: `id` should be a CURIE

lang/python/tests/data/good_combined_airr.yaml
```diff
@@ -216,11 +210,12 @@ Repertoire:
...
```
bussec, line 217: `id` should be a CURIE

lang/python/tests/data/good_combined_airr.yaml
```diff
@@ -359,11 +347,12 @@ Repertoire:
...
```
bussec, line 354: `id` should be a CURIE

lang/python/tests/data/good_repertoire.yaml
```diff
@@ -29,11 +29,12 @@ Repertoire:
...
-      age_min: 27
-      age_max: 27
-      age_unit:
-        id: UO_0000036
-        label: year
+      age:
+        min: 27
+        max: 27
+        unit:
+          id: UO_0000036
```
bussec, line 36: `id` should be a CURIE

lang/python/tests/data/good_repertoire.yaml
```diff
@@ -162,11 +156,12 @@ Repertoire:
...
```
bussec, line 163: `id` should be a CURIE

lang/python/tests/data/good_repertoire.yaml
```diff
@@ -295,11 +283,12 @@ Repertoire:
...
```
bussec, line 290: `id` should be a CURIE

specs/airr-schema-openapi3.yaml
```diff
@@ -473,9 +479,67 @@ TimePoint:
...
+        max:
+            type: number
+            nullable: true
+            description: Upper/maximum value of the time interval
+            example: 10.0
+            x-airr:
+                adc-query-support: true
+        unit:
+            $ref: '#/Ontology'
+            nullable: true
```
bussec, line 503: see above -> `nullable` status

specs/airr-schema-openapi3.yaml
```diff
@@ -473,9 +479,67 @@ TimePoint:
...
+        quantity:
+            type: number
+            nullable: true
+            description: Physical quantity
+            example: -5.0
+            x-airr:
+                adc-query-support: true
+        unit:
+            $ref: '#/Ontology'
+            nullable: true
```
bussec, line 531: see above -> `nullable` status

specs/airr-schema-openapi3.yaml
```diff
@@ -2148,11 +2202,15 @@ Diagnosis:
                         id: DOID:4
                         label: disease
         disease_length:
-            type: string
+            $ref: '#/PhysicalQuantity'
```
bussec, line 2205: See above -> `TimeInterval`

specs/airr-schema.yaml
```diff
@@ -478,12 +480,74 @@ TimePoint:
...
+                nullable: true
+        unit:
+            $ref: '#/Ontology'
+            description: Unit of the time interval
+            example:
+                id: UO:0000033
+                label: day
+            x-airr:
+                adc-query-support: true
+                nullable: true
```
bussec, line 514: see above -> `nullable` status

specs/airr-schema.yaml
```diff
@@ -478,12 +480,74 @@ TimePoint:
...
+        unit:
+            $ref: '#/Ontology'
+            description: Unit of physical quantity
+            example:
+                id: UO:0000024
+                label: nanogram
+            x-airr:
+                nullable: true
```
bussec, line 541: see above -> `nullable` status

specs/airr-schema.yaml
```diff
@@ -2046,10 +2104,14 @@ Diagnosis:
                         id: DOID:4
                         label: disease
         disease_length:
-            type: string
+            $ref: '#/PhysicalQuantity'
```
bussec, line 2107: See above -> `TimeInterval`

bussec (approved):
Looks good in general, some minor fixes.

lang/R/inst/extdata/airr-schema.yaml
bcorrie: They have always been that way since they were added (e.g. collection_time_point_relative_unit was always nullable... 8-)

I suppose in a TimePoint or TimeInterval it makes sense to change this behaviour. Having the unit nullable when it was a field by itself makes some sense, much less sense in the new constructs I agree.

schristley: good catch, yes the whole object can be null, but if you specify the object then should always have the values and the units. I pushed a commit to change that.

lang/R/tests/data-tests/good_combined_airr.json
bussec: Ok, I just fixed this throughout all example files (Python and R) as part of #741 .

lang/R/tests/data-tests/good_combined_airr.json
bussec: Done as part of #741

lang/R/tests/data-tests/good_combined_airr.json
bussec: Done as part of #741

lang/R/tests/data-tests/good_combined_airr.yaml
bussec: Done as part of #741

lang/R/tests/data-tests/good_combined_airr.yaml
bussec: Done as part of #741

lang/R/tests/data-tests/good_combined_airr.yaml
bussec: Done as part of #741

lang/R/tests/data-tests/good_repertoire.yaml
bussec: Done as part of #741

lang/R/tests/data-tests/good_repertoire.yaml
bussec: Done as part of #741

lang/R/tests/data-tests/good_repertoire.yaml
bussec: Done as part of #741

lang/python/tests/data/good_combined_airr.json
bussec: Done as part of #741

lang/python/tests/data/good_combined_airr.json
bussec: Done as part of #741

lang/python/tests/data/good_combined_airr.yaml
bussec: Done as part of #741

lang/python/tests/data/good_combined_airr.yaml
bussec: Done as part of #741

lang/python/tests/data/good_combined_airr.yaml
bussec: Done as part of #741

lang/python/tests/data/good_repertoire.yaml
bussec: Done as part of #741

lang/python/tests/data/good_repertoire.yaml
bussec: Done as part of #741

lang/python/tests/data/good_repertoire.yaml
bussec: Done as part of #741

bcorrie:
@schristley and I both did this in parallel - I think I merged them OK. I added a couple of adc_query_support attributes that were missing.

bcorrie:
@bussec remaining issue from above is `disease_length`, I believe it should be a `PhysicalQuantity` as I don't think a `TimeInterval` makes sense since it is a length not a time span.

schristley:
> @bussec remaining issue from above is `disease_length`, I believe it should be a `PhysicalQuantity` as I don't think a `TimeInterval` makes sense since it is a length not a time span.

I agree it should be `PhysicalQuantity` because it is a length or duration. If you go to a `TimeInterval` then what value would you put for `min`, why would it be anything other than zero? If min is zero, it's the same result as `PhysicalQuantity`, while if it's non-zero, then what is it relative to?

With patient data, we don't want to use dates, so what we are seeing with the AKC work (and you see this when getting clinical data), is that there is a defined `T=0` event, and then everything after that is a relative time point. But we don't have that in AIRR yet.

In the future redesign there will be time points for specific events, so the description, "Time duration between initial diagnosis and current intervention" would suggest an event at "initial diagnosis" and an event at "current intervention" and this length could be computed from those.

schristley:
> > @bussec remaining issue from above is `disease_length`, I believe it should be a `PhysicalQuantity` as I don't think a `TimeInterval` makes sense since it is a length not a time span.
> 
> @bcorrie I agree that `TimeInterval` (or `TimePoint`) sound odd, but they would enforce the correct units for a measurement of time, which `PhysicalQuantity` does not. It is probably more a question how we call the `Time*` objects, but we can worry about that another day as the current solution does not create any limits here. So no objections from my side.

I see. Yes, you are right, I missed that.

So we need to fix this because `PhysicalQuantity` has the top ontology node as `mass unit` and thus technically we are not allowed to specify time units with it.

bcorrie:
Do we want to add a `TimeQuantity` object - to enforce units. Or just have a general `Quantity` object with a higher root node? Otr just change the current root node and leave it as a `PhysicalQuantity`.

I would be tempted to change the name to `Quantity` and change the root node to `UO:0000000, unit`

@bussec @schristley thoughts?

schristley:
I think we are pretty much required to have a `TimeQuantity` object to enforce time units. If we change the root to `UO:0000000, unit` then that will allow all types of incorrect units to be used.

An interesting aside, [LinkML](https://linkml.io/linkml/schemas/slots.html#slot-usage) has `slot_usage` which allows for this. That is, we could define a general `Quantity` but then indicate when used as `disease_length` that its `range` is a time unit, while other usages of the slot might have `range` as mass units, and so on. I would guess how this gets translated into JSON schema ends up being not much different from what we are doing manually.

bussec:
+1 for introducing a `TimeQuantity` object.

bcorrie:
I will work on this...

lgcowell:
[0]     Lindsay Cowell reacted to your message:
________________________________

specs/airr-schema.yaml
bcorrie: Created TimeQuantity and made the change

specs/airr-schema-openapi3.yaml
bcorrie: Created TimeQuantity and made the change

lang/R/inst/extdata/airr-schema.yaml
bcorrie: Created TimeQuantity and made the change

bcorrie:
This has been done - ready for merge, unless we can find anything else...

PR #568, opened 2021-12-05
Germline labels

### LINKED ISSUE #571: Define J gene donor splice site
`AlleleDescription.j_donor_splice` aims to define the donor splice site 3' of a J gene. However, as the actual splicing occurs between two nucleotides, it needs to be defined whether the referenced position is the last exonic or first intronix nucleotide.

williamdlees:
closes #571 

Add a description of the file structure to the Germline Sets section, and fix minor typos. Add GermlineSet and GenotypeSet to the list of high-level schema objects.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -407,7 +407,8 @@ fields = ['Name', 'Type', 'Attributes', 'Definition']
 tables = ['Repertoire', 'Study', 'Subject', 'Diagnosis', 'Sample', 'CellProcessing', 'NucleicAcidProcessing',
           'PCRTarget', 'SequencingRun', 'RawSequenceData', 'DataProcessing',
           'Rearrangement', 'Alignment', 'Clone', 'Tree', 'Node', 'Cell',
-          'RearrangedSequence', 'GermlineSequence', 'GeneDelineationV', 'GeneDescription', 'GermlineSet']
+          'RearrangedSequence', 'UnrearrangedSequence', 'SequenceDelineationV', 'AlleleDescription', 'GermlineSet',
+          'GenotypeSet','Genotype']
 for spec in tables:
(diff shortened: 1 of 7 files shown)


schristley:
The file structure will need to match the [DataFile](https://github.com/airr-community/airr-standards/blob/cf2ceaa71cae180a69867d0b541148146102f703/specs/airr-schema-openapi3.yaml#L156) which also has a direct correspondence to how data is returned from the ADC API.

williamdlees:
Thanks Scott. I have added a reference to the AIRR DataFile,

bussec (commented):
While the Germline objects are part of the AIRR Schema, I don't think that we should include them into MiAIRR as this would be an incompatible change that we would have to push until the next major release. Also, they are not currently not minimal information that would be created in the majority of the experiments.

Therefore I would suggest to revert the changes in lines 36 and 51 or the `overview.rst`.

williamdlees:
> 

Done, but can we please discuss their inclusion in the next major release.

bussec (approved):

bussec:
@williamdlees There is no problem in including the Germline object in the next release of the Schema. But if we consider this to be **required** metadata for all studies, this would be an incompatible change of the MiAIRR standard an thus would have to wait until v2.0. Does this clarify my previous point?

schristley:
MiAIRR does require a germline set reference, `germline_database` in set 5 (DataProcessing). It might be reasonable to now require that field to contain the globally unique germline reference ID, that is, `germline_set_ref` instead of being free-form text?

bussec:
@schristley I agree, we should move in this direction. But IMO that's another pull request :smiley:

schristley:
added some more detail about file structure, I think this might be ready to merge.

schristley (approved):

PR #575, opened 2022-01-13
Remove MiXCR from Rearrangement support list

### LINKED ISSUE #565: Can I ask and request why does MiXCR still does not support AIRR standards?
I can see in the documentation that MiXCR was listed to support the AIRR rearrangement schema since version 2.2.1 for 'output'. While other tools like ChangeO have long since complied with AIRR standards, MiXCR seems to push back this comply to their next version ever since. As of today, version 3.0.13 does not comply and there is still no ETA for this feature in their future versions. I recommend removing MiXCR from the list until they prove to support AIRR output formats, while encouraging them to integrate this feature in the next version 3.0.14 as promised.

bussec:
Fixes #565


javh (approved):
Thanks for taking care of this.

Issue #471, opened 2020-10-13
abundance value and other counts for a clone?


schristley:
Some experimental protocols utilize controls that account for PCR amplification bias and other biases, thus allowing abundance between clones to be relatively compared within a repertoire sample. Should we have a field to hold this abundance value?

Furthermore, should we have a count field for the multiplicity (`duplicate_count`) of rearrangements? We have `sequence_count` but is that meant to be the number of rearrangement records, or the sum of `duplicate_count`? If an abundance value is not available for the clone, this count offer some notion of abundance even though it's potentially biased.

bcorrie:
MiXCR has a cloneFraction field that seems to map to abundance... Currently this field is not mapped to an AIRR field when loading a clone into the iReceptor Turnkey. This seems reasonable to me...

scharch:
> We have `sequence_count` but is that meant to be the number of rearrangement records, or the sum of `duplicate_count`?

Number of records. I am fine with adding an `abundance` field, but not keen on what you call multiplicity. If the analyst believes multiplicity is accurate enough, s/he can store it in `abundance`. If not, then it's not probably not meaningful to keep.

scharch:
> MiXCR has a cloneFraction field that seems to map to abundance

Relatedly, I think we should define the field as a fraction --so maybe it should be `frequency` instead of `abundance`??

schristley:
> I am fine with adding an `abundance` field, but not keen on what you call multiplicity. If the analyst believes multiplicity is accurate enough, s/he can store it in `abundance`. If not, then it's not probably not meaningful to keep.

That makes sense. I was thinking a downstream user would want to decide for themselves, and one field over the other would indicate confidence without digging through the experimental protocols, but I see that enforcing that would be problematic.

> so maybe it should be frequency instead of abundance??

sure.

schristley:
> Relatedly, I think we should define the field as a fraction --so maybe it should be `frequency` instead of `abundance`??

Actually, after giving this a second thought, having `frequency` can be an issue if any tools want to "renormalize" for some reason. With just a frequency, you need a total to change that frequency back into an absolute count.

Issue #317, opened 2020-01-16
Extend Clone to single-cell context


scharch:
Starting to think about this in the context of generating a lot of 10x VDJ data... it seems we will want to (eventually) have a way for `Clone`s to contain `cell`s (see https://github.com/airr-community/airr-standards/issues/273#issuecomment-568649516), instead of (or maybe in addition to) `Rearrangement`s.

Just a marker for now, need to think more about what kind of representation would make sense...

schristley:
Should the Clone definition also contain both chains? Right now it seems to support only one.

scharch:
@schristley, I think it will have to support `germline_alignment` (and all the related fields) as an array, of some sort, yes.

schristley:
In a separate call, @bussec and I discussed how to do this flexibly. It would be nice not to be limited to strictly two chains. It also is hard to come up with a terminology that covers both T and B cells. There was also the desire to be able to annotate non-productive chains. Using a dictionary or array object should allow multiple entries. Using a controlled vocabulary, we could use T and B cell specific terms to annotate/tag the chains. At the same time, we should make it easy to access the primary annotations directly.

javh:
> It also is hard to come up with a terminology that covers both T and B cells.

This is a rather vexing problem. We've been using "heavy" for IGH, TRB and TRD and "light" for IGK/L, TRA and TRG, which is wrong. Maybe `long_chain` and `short_chain`?

scharch:
> Can we call them heavy_chain, light_chain, alpha_chain, beta_chain, etc., with a controlled vocabulary specific to cell and chain type?

@schristley I was just coming here to suggest essentially the same thing.

It'll still get complicated, though: if each chain is a dict with keys something like `{id, type, is_productive}`, then a `Cell` would be an array of those and the "members" of `Clone` ends up being an array of arrays of dicts. Does that seem workable?

> At the same time, we should make it easy to access the primary annotations directly.

Each `Cell` in the `Clone` has a `cell_id` and a list of `sequence_id`s that link back to the rearrangements TSV - do you think that is sufficient?

javh:
> Can we call them heavy_chain, light_chain, alpha_chain, beta_chain, etc., with a controlled vocabulary specific to cell and chain type?

This is hard to use (have to check every object for field presence), set required fields for (none or all are required?), and convert to a TSV (lots of missing data). But, it would be more explicit and support dual BCR+TCR expressing cells if you believe in such things:

[...]

PR #586, opened 2022-02-12
Clone updates

### LINKED ISSUE #161: more documentation for duplicate_count
From a comment by Mathieu of vidjil group:

In particular, we used the `duplicate_count` key to describe the number of reads gathered in each clone, as you suggested us in your previous email. To enable a better interoperabilityn maybe this suggestion could be clearly stated in AIRR documentation of this field? For example, taking inspiration on what was already on the `sequence_id` field,  "duplicate_count (...) This may also be a total number of gathered sequences in cases where query sequences have been combined in some fashion by the tool."

### LINKED ISSUE #543: Clone spec mapping for common tools
Hello!

The iReceptor team is hoping to load some clones data, so we've been reviewing the clones [spec](https://github.com/airr-community/airr-standards/blob/3d78de34c887921836502f2256c0d57f9fb37876/specs/airr-schema.yaml#L2957).

I've created a mapping file with my best guess at mapping clones output from MiXCR, Change-o, 10X, and ImmuneDB. 

[clone.mapping_with_counts_examples-2021-10-29.xlsx](https://github.com/airr-community/airr-standards/files/7443603/clone.mapping_with_counts_examples-2021-10-29.xlsx)

Hoping for a review from the community to see if we are in agreement. Some specific questions that came out of this exercise:

1) It seems that most tools do not include v/d/j alignment start/end positions in the output. Am I missing something here - maybe a data processing step or a command-line argument?

2) For single-cell data, it became clear we'll need to resolve this issue: [extend clones to the single-cell context](https://github.com/airr-community/airr-standards/issues/317). Specifically for 10X data, clone_id comprises multiple loci, so we should decide if/how to split this out. We suggest modifying clone_id as follows (example is for BCR data): 

change clone_id = “CLONEA” to three clones with clone_id = “CLONEA_IGH”, clone_id = “CLONEA_IGL”, and clone_id = “CLONEA_IGK”.

3) In the same vein, we'll need to consider how to count clones derived from single-cell data. I came up with 3 options; option 1 (aggregating clone counts across loci) isn't very useful; option 2 (locus-specific clone count) makes more sense and can be implemented fairly easily; option 3 (cell-specific clone count) is most informative but might be trickier to implement right now, as it links to the cell spec. Any thoughts on how best to proceed?

Thanks!

bcorrie:
Closes #543 
Closes #161

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -3557,15 +3557,14 @@ Rearrangement:
         consensus_count:
             type: integer
             description: >
-                Number of reads contributing to the (UMI) consensus for this sequence.
+                Number of reads contributing to the UMI consensus or contig assembly for this sequence.
                 For example, the sum of the number of reads for all UMIs that contribute to
                 the query sequence.
         duplicate_count:
             type: integer
             description: >
                 Copy number or number of duplicate observations for the query sequence.
-                For example, the number of UMIs sharing an identical sequence or the number
-                of identical observations of this sequence absent UMIs.
+                For example, the number of identical reads observed for this sequence.
             title: Read count
             example: 123
             x-airr:
@@ -3574,6 +3573,12 @@ Rearrangement:
                 set: 6
                 subset: data (processed sequence)
                 name: Read count
+        umi_count:
+            type: integer
+            description: >
+                Number of distinct UMIs represented by this sequence.
+                For example, the total number of UMIs that contribute to
+                the contig assembly for the query sequence.
         cell_id:
             type: string
             description: >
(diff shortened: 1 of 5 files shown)


javh:
Thanks, @bcorrie. I pushed some updates for us to hammer on. I also added `umi_count` to Rearrangement and changes some of the wording in there.

One of the things we need to figure out is whether `clone_count` should be the count of:
1. Total sequences.
2. Unique sequence.
3. Either of the above, depending upon tool and analysis context.

I'm certain we talked about this before - I'll try to find where later.

schristley:
I put the description for `clone_count` back to its original, which is to imply that the clonal assignment tool, or companion tool, decides what is the clonal abundance (count).

javh:
@schristley, I don't think original description is correct. Two examples where this is plainly wrong are:

1. Classical bulk BCR sequencing clone sizes, which are often calculated as the number of unique variants.
2. Adaptive's data which does not provide non-normalized counts - only counts normalized to their standard curves.

schristley:
> @schristley, I don't think the original description for `clone_count` is correct (which is why I changed it). Two examples where this is plainly wrong are:

@javh Okay, but your description makes it sound like it's just a count of rearrangement records. The original intent #471 of `clone_abundance` which got renamed to `clone_count` was to record what the analysis tool (and/or experimental protocol) inferred to be the clone size. "Non-normalized absolute" is meant to indicate that a frequency shouldn't be provided, as other analysis might want to perform their own normalization. Is this better?

Non-normalized absolute count of the _inferred_ number of members (immune cells) in this clone.

javh:
"just a count of rearrangement records" will often be correct. Whether it's total sequences or unique sequences will depend upon whether duplicates were removed before calculating `clone_count`.

javh:
@schristley, I like it. That seems to cover everything better than the original wording and my wording. We can probably drop the last sentence, as we don't need to provide suggestions for how to perform relative abundance calculations that aren't covered by a standard field.

schristley:
We already starting making changes for #161 so we should make the other changes as well.

scharch:
NOTE: This PR does **not** address #317

bcorrie:
@javh there is a the Travis CI check that is not passing. Expected but waiting - is that expected?

bcorrie:
Finally - synced all the specs so they pass consistency check 8-)

Hope I didn't screw up any definitions in the progress - @javh you are on for review, can you double check...

javh:
Looks good to me. Thanks, @bcorrie. I'll merge this.

That weird Travis "expected but waiting" thing should be gone in every PR now.

Also, we need to get in touch with 10x about the `umi_count` field being in v1.4.

PR #588, opened 2022-02-17
Added content type to curl post queries

### LINKED ISSUE #576: ADC curl POST commands in docs need a header
Some repositories (at least VDJServer) are strict and require the header to be of the correct type. The documentation doesn't say this and the curl commands don't work against VDJServer. This doesn't work:

curl --data @/data/src/AIRR/adc-api-tests/repertoire/pass-query1-2_repertoire.json https://vdjserver.org/airr/v1/repertoire

While this does:

curl -H "Content-Type: application/json" --data @/data/src/AIRR/adc-api-tests/repertoire/pass-query1-2_repertoire.json https://vdjserver.org/airr/v1/repertoire

bcorrie:
-H "Content-Type: application/json"

Closes #576

### DIFF
--- a/docs/api/adc_api.rst
+++ b/docs/api/adc_api.rst
@@ -145,7 +145,7 @@ API is available for VDJServer's data repository.
 
 .. code-block:: bash
 
-  curl https://vdjserver.org/airr/v1
+  curl -H "Content-Type: application/json" https://vdjserver.org/airr/v1
 
 The response should indicate success.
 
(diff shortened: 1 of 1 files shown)


bcorrie:
@schristley I think this is ready to merge/close with the exception of the VDJServer examples having hardcoded IDs

schristley:
@bcorrie I believe these changes are included in #601

javh:
Can we void this one then?

bcorrie:
It looks like it - the header changes are in #601 as well. I am closing...

Issue #590, opened 2022-02-21
Field consistency check for RearrangedSequence


javh:
The RearrangedSequence Germline schema includes some fields with very similar, or identical, meaning to fields in Rearrangement and/or Alignment. particularly the following:

* `seq_start`: `Alignment.sequence_start`, `Rearrangement.v_sequence_start`?
* `seq_end`: `Alignment.sequence_end`, `Rearrangement.v_sequence_end`?
* `notes`: `description`, `name`? What's the use case?

Can we clarify the intended meaning and rename accordingly?

williamdlees:
RearrangedSequence is a record which points to the existence of evidence in a repository. For example this could be PCR amplified cDNA in Genbank supporting the existence of a D allele, in which case seq_start and seq_end would delimit the sequence of the D allele utself, within the seuqnece depositied at Genbank. Although there is similarity, I don't think it is such that we should seek to use the same name in both cases.

Notes is a free-text field in which curators can add notes. In OGRDB these sometimes run to a page or two.

javh:
That sounds identical to `Alignment:sequence_start` and `Alignment:sequence_end` to me, or at the very least so similar that `seq_` and `sequence_` are unclear distinctions. I suppose it's similar to `germline_start` / `germline_end` as well.

Regarding `notes`, is this similar enough to `Study:study_description` and `Repertoire:repertoire_description` that it should be named `_description`?

williamdlees:
sorry, I was put off by the _v.

I will change notes to _description, with a description indicating that these are for curational notes.

javh:
> I will change notes to _description, with a description indicating that these are for curational notes.

Okay. Though, if you think curation notes are different from a description, then we should use a different name. Just trying to understand what is and is not equivalent and harmonize whenever appropriate.

williamdlees:
I don't think there's a sufficiently strong distinction to merit a different name.

Issue #592, opened 2022-02-27
Update documentation to state that the -call fields in a Rearrangement should cite the label field in an AlleleDescription


williamdlees:

scharch:
@javh seems trivial, should go in 1.5

scharch:
@williamdlees can you make the desired update

javh:
From the call:
Update AlleleDescription.label.description to be consistent with `_call` fields

Issue #591, opened 2022-02-27
Tests for GermlineSet should check that there is only one AlleleDescription for each allele in the list


williamdlees:

scharch:
@williamdlees closing as done

scharch:
nope sorry misread that...

schristley:
This validation test is beyond standard schema validation supported by the reference libraries. This validation will be defined by AKC and will be integrated back once an overall validation architecture is in place.

schristley:
Didn't get the validation framework in place for the beta release, so moving forward

Issue #562, opened 2021-10-01
what's the scope of the identifiers in the germline schema objects?


schristley:
This is important in relation to the ADC, both for the API which will allow queries, but also for repositories that store multiple studies that use multiple germline sets.

- [x] `germline_set_id`. I believe this should be a globally unique identifier. It's important that this identifier being resolvable, i.e. there is a clear procedure to retrieve the germline set data from its id. Services that manage and generate identifiers should probably use the decentralized identifier standard. Individuals that wish to publish their own germline sets should likely use such a service so they can be provided with a resolvable, persistent identifier.
- [x] `allele_description_id`. Is this only for internally linking `GermlineSet` and `AlleleDescription`? If so, are ADC repositories allowed to assign whatever value they wish to insure uniqueness? Also this implies that `GermlineSet` and `AlleleDescription` data must always be kept "together" otherwise the linkage can be lost.
- [x] `sequence_id`. Is this only for internally linking in `AlleleDescription`?
- [x] `sequence_delineation_id`. Same as `sequence_id`?
- [x] `receptor_genotype_set_id` and `receptor_genotype_id`?
- [x] `mhc_genotype_set_id` and `mhc_genotype_id`?

Additional questions regarding the ADC API:

- [x] What query end points should we have? A `germline` end point seems obvious. An `allele` end point?
- [x] If `germline_set_id` is a globally unique identifier then the API should be straight forward.
- [x] For an `allele` end point, what is the identifier that will return the unique record?

javh:
`sequence_id` is used as the primary key across objects for any observed sequence. Some of these `_id` fields seem redundant with `sequence_id` functionally. Eg, `RearrangedSequence. repository_id` and `UnrearrangedSequence. assembly_id` may be more appropriate as `sequence_id`.

williamdlees:
`RearrangedSequence.repository_id` would be, for example, the accession id in Genbank. We don't have control over its value
The Unrearranged sequence fields such as assembly_id match this - in that they reflect the common way the sequence is referred to in a database such as ensenbl

keep separate `id `and `ref `fields for `germline `and `allele_description `- `keep ids in as internal linker, `ref `is external curie type. Add a ref to `AlleleDescription`

williamdlees:
As agreed in the standards call on the 21st:
- the ids are unique identifiers
- refs are CURIE-style references 
- both GermlineSet and AlleleDescrioption should have refs, so that they can be queried via an API
[...]

PR #593, opened 2022-02-27
Germline doc

williamdlees:
This closes multiple points raised in [#559]( https://github.com/airr-community/airr-standards/issues/559) and [#562](https://github.com/airr-community/airr-standards/issues/562) - see comment in those threads, and the change notes below. In addition I have raise two new minor issues [#591]( https://github.com/airr-community/airr-standards/issues/591) and [#592]( https://github.com/airr-community/airr-standards/issues/592) so that #559 and #562 can be closed once this pull request is accepted and merged.


williamdlees:
Also now addresses #556 and #572

williamdlees:
And is also intended to close #590

williamdlees:
germline_description_description doesn't trip off the tongue. Changed the _descriptions to curational_notes, hope that's ok

javh:
I think if the only place we ever plan to use `_notes` is in the germline objects, then I think that's fine.  If we think we might add similar fields outside these objects, maybe we can do a quick informal poll?

Some suffix suggestions:
* `_doc`
* `_notes`
* `_comment`
* `_description`
* None (just `curation`)

Prefix ideas:
* `curational_`
* `curation_`
* None (just `notes`)

I might be a bit overly cautious about future-proofing here, but... thoughts?

scharch:
I kinda like just `comments`
A prefix/suffix seems like overkill unless we think we will also have some other kind of `_notes` (or some other `curation_`, I suppose)

bussec:
@williamdlees I updated the sections of the Schema that were changed by #524, #564 and #603, but excluded the (non-MHC) germline objects. The merge to master should now work again, the only problem is that the R test is failing again (@javh, any ideas)?

javh:
I just checked the R package locally and I'm not seeing the same `Duplicate map key: 'InformationProvider'` error from `yaml::yaml.load_file`.  I'll have to fight with it.

javh:
The R check failure was due to merging against master generation a duplicate copy of the `InformationProvider` object in the specs, which is why the error wasn't present locally. I'm not sure how that happened, but I blame the force-push creating an improper merge conflict resolve.

Seems fixed now.

bussec:
@javh Thanks for all the troubleshooting! I really hate to say it but you fixed the consistency check, but now the merge conflicts are back :cry: I did so testing locally and the order of the commits seems to matter, so let me check whether squashing solves the issue...

javh:
Haha. Okay, well.. Let's hope. We may have to clean it up post-merge or something.  I think we just have the "notes" fields to deal with for this PR, yes?

bussec:
Finally! It did not require squashing, a rebase to the current `master` was sufficient. So know we got consistency checks, commit history and the ability to actually merge this :tada: Everyone who is on a local repo, please remember to update this branch before making any new commits.

bussec (approved):

bussec:
@javh 

> I think if the only place we ever plan to use _notes is in the germline objects, then I think that's fine. [...]  Maybe we can do a quick informal poll?

In general I hope that we will not need additional comments fields like this, as they might be abused as dumps for information that should go into other existing fields. However, I recognize that when representing/exchanging GLDB data there is a valid use case for comments on the curation process. So it is ok to make an exception here, but IMO we do not need to agree on a common suffix.

bussec:
@javh Is there anything more we need to discuss regarding the `curational_notes`?

javh:
Not really, I think you're right that it's unlikely to have a broader use case, so we just need to pick something.  Any objections to just `curation`?

williamdlees:
No objection from me.

williamdlees:
This would be ok with me. The description already says ‘curational notes’ so it doesn’t really matter if it’s in the name. In my use case at least.

PR #603, opened 2022-04-22
Added changes to MHC genotype objects per standards meeting

### LINKED ISSUE #573: Grouping of MHC genotype information
Is it meaningful to have a grouping for `MHCGenotype` based on MHC class I/class II?  

In contrast to IG and TR, both classes of MHC genes and both chain types of Class II are located on the same locus (except in *Osteichthyes*). Thus, if `MHCGenotype.genotype_class` actually annotates the locus, it would always be `MHC` and therefore could be dropped. This would also make `MHCGenotype` redundant, as there would only be a single instance of this object per subject, so all properties could as well be moved into the `MHCGenotypeSet`.

Alternatively, we could decide to group the genes into two objects, based on their classification as either Class I or II, which is not always obvious from the gene name (e.g., in mouse).

bcorrie:
Closes #573

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -28,161 +28,161 @@ Ontology:
 
 # Map to expand CURIE prefixes to full IRIs
 CURIEMap:
-  CL:
-    type: ontology
-    default:
-      map: OBO
-      provider: OLS
-    map:
-      OBO:
-        iri_prefix: "http://purl.obolibrary.org/obo/CL_"
-  DOID:
-    type: ontology
-    default:
-      map: OBO
-      provider: OLS
-    map:
-      OBO:
-        iri_prefix: "http://purl.obolibrary.org/obo/DOID_"
-  MRO:
-    type: ontology
-    default:
-      map: OBO
-      provider: OLS
-    map:
-      OBO:
-        iri_prefix: "http://purl.obolibrary.org/obo/MRO_"
-  NCBITAXON:
-    type: taxonomy
-    default:
-      map: OBO
-      provider: OLS
-    map:
-      OBO:
-        iri_prefix: "http://purl.obolibrary.org/obo/NCBITaxon_"
-      BioPortal:
-        iri_prefix: "http://purl.bioontology.org/ontology/NCBITAXON/"
-  NCIT:
-    type: ontology
-    default:
-      map: OBO
-      provider: OLS
-    map:
-      OBO:
-        iri_prefix: "http://purl.obolibrary.org/obo/NCIT_"
-  UBERON:
-    type: ontology
-    default:
-      map: OBO
-      provider: OLS
-    map:
-      OBO:
-        iri_prefix: "http://purl.obolibrary.org/obo/UBERON_"
-  UO:
-    type: ontology
-    default:
-      map: OBO
-      provider: OLS
-    map:
-      OBO:
-        iri_prefix: "http://purl.obolibrary.org/obo/UO_"
-  ORCID:
-    type: catalog
-    default:
-      map: ORCID
-      provider: ORCID
-    map:
-      ORCID:
-        iri_prefix: "https://orcid.org/"
-  DOI:
-    type: identifier
-    default:
-      map: DOI
-    map:
-      DOI:
-        iri_prefix: "https://doi.org/"
-  ROR:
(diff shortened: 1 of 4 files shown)


javh (approved):
Seems fine to me, except for the schema consistency check failure.

bcorrie:
Anyone else (@bussec) have any thoughts on this? If not I will fix the consistency checks and then merge...  I left the consistency checks unfixed in case we made changes - didn't want to have to do them N times 8-)

bussec:
@bcorrie I find `MH1` and `MH2` *highly* uncommon. In contrast to `IG` and `TR` they are not even used in the gene symbols. I also check again with  [MRO](https://www.ebi.ac.uk/ols/ontologies/mro) (which is used by IEDB and which we are trying to implement for the `Receptor` object) and while they unfortunately do not provide any top-level definition, they use the string  "MHC class (I|II)` consistently. Therefore, I would opt for going with this, can make the necessary changes if you want.

bussec:
One more thing: We IMO never had an explicit rule about this, but I just realized that most of our `enum`s  are low-caps snake_case -- the main exception being the IG/TR loci fields and and two values for `sex`. If we want to stay consistent with this, then `mhc_class_i` and `mhc_class_ii` would be the values of choice.

bcorrie:
@bussec no objections what so ever. You can make the changes if you like, I will get to them probably later this week if  you don't...

Brian

bussec (approved):

javh:
Spaces in terms (`MHC class I`) and names people aren't likely to use in tool output/publication (`mhc_class_i`) are both worrisome to me.

bussec:
@javh Agreed. For my perspective there are the following options (note that we should have support of invariant/non-classical MHC as well):

1. Common textbook terms with spaces, also used in MRO: `MHC class I`, `MHC class II`, `non-classical MHC`
2. Terms defined by IMGT (U Montpellier) that are not used by anyone else: `MH1`, `MH2`, `RPI-MH1Like`
3. Terms defined by IMGT/HLA (EBI): *There are none!* (they only describe gene locus, not a gene group)
4. The current "nerd-case" terms: `mhc_class_i`, `mhc_class_ii`, `non_classical_mhc`
5. A compromise between 1 and 4, e.g., `MHC-I`, `MHC-II`, `MHC-invariant`

javh:
> A compromise between 1 and 4, e.g., MHC-I, MHC-II, MHC-invariant

I've seen `MHC1` and `MHC-1` in a few publications / databases as well. 

The invariant chain abbreviation seems to be `Ii`. That's awful.

bussec:
`Ii` refers specifically to [*CD74*](https://en.wikipedia.org/wiki/CD74), which is involved in Class II presentation. The "invariant" MHC that I was referring to are things like CD1d or HLA-F, which are structurally closer to MHC-I. There are multiple terms for these gene/molecules: invariant, non-polymorphic, non-classical. In my suggestion I went for "invariant" but I see that there is potential for confusion with `Ii`, so maybe `MHC-nonclassical` would be a better choice.

javh:
@bussec Ah, gotcha. I did not know that. Well that makes things harder.

bussec:
And just for avoidance of doubt: Yes, Roman numerals are the sole culprit of the fall of the Roman Empire. But for MHC classes this has been used by immunologist for decades, so I would stick to it if possible.

bussec:
@bcorrie @javh If we agree on the terms `mhc_class` this is good to go from my side. Additional changes since yesterday:
* Add MRO for mark-up of parental genes (so that people do not have to parse the name, which is a pain for non-human species)
* Removed the strict controlled vocabulary from `mhc_genotyping_method`, will add this in #593 
* `mhc_alleles` and `mhc_class` are now not nullable, as an `MHCGenotype` object is useless without them (and in this case should be omitted as a whole)
* changes some field names to be more consistent and more distinct from the IG/TR germline gene stuff

javh:
@bussec, Good with me. I like the compromise names. They are clean while being familiar enough that they won't cause confusion.

Speaking of, if we're making stuff up, we could use `MHC-N` or `MHC-NC` instead of `MHC-nonclassical`. Though, I don't have a preference, because `-nonclassical` has the benefit of being immediately understood.

bussec:
@bcorrie Will try to merge this now... the errors are due to the changes in the #574 PR, not the #524 for which I already did the rebase.

Issue #605, opened 2022-04-27
Update ADC API documentation for ADC API v1.2 (AIRR Spec v1.4)


bcorrie:
We need to update the ADC API documentation to reflect what is in the ADC v1.2/AIRR v1.4 release.

PR #608, opened 2022-05-04
Update ADC docs for v1.2 (AIRR v1.4)

### LINKED ISSUE #597: Add documentation for ADC API requirements
Reflect the following as discussed at CRWG and Standards:

- TSV output SHOULD be supported for all data that is commonly represented in tabular structures
- All API endpoints SHOULD return JSON encodings
- For some API endpoints it is possible to request TSV files
- For any endpoints supporting TSV encodings these will be clearly documented in the API Spec and/or Documentation
- Endpoints that do not support TSV can reject TSV requests
- If API endpoint returns a field then the content of that field in JSON and TSV MUST be equivalent
- Bottom line: API Endpoints that are JSON or JSON+TSV should be defined in the spec/docs - the repositories need to follow this spec.

bcorrie:
Closes #597 #605

### DIFF
--- a/docs/api/adc_api.rst
+++ b/docs/api/adc_api.rst
@@ -628,9 +628,9 @@ be returned.
 
 .. code-block:: bash
 
-  curl https://covid19-1.ireceptor.org/airr/v1/rearrangement/{clone_id}
+  curl https://covid19-1.ireceptor.org/airr/v1/clone/{clone_id}
 
-Where clone_id is the ID of a clone object in the repository. The response will provide the ``Rearrangement`` data in JSON format.
+Where clone_id is the ID of a clone object in the repository. The response will provide the ``Clone`` data in JSON format.
 
 .. code-block:: json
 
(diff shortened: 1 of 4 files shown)


bcorrie:
@schristley do you want to have a look at this? Basic sections for new API entry points now exist.

schristley:
Enabled branch on [RTD](https://docs.airr-community.org/en/adc-docs/api/adc_api.html).

The `CellExpression` schema has title of `Cell` instead. Doesn't look like `airr/v1/expression/{expression_id}` is in the API spec? Receptor schema link is not active. `CellExpression` schema seems to be missing from the AIRR Data Model with the other schemas. Otherwise looks good.

bcorrie:
> * The `CellExpression` schema has title of `Cell` instead.

@schristley I can't find where this occurs???

bcorrie:
> * Doesn't look like `airr/v1/expression/{expression_id}` is in the API spec?

Nice catch, fixed.

schristley:
> > * The `CellExpression` schema has title of `Cell` instead.
> 
> @schristley I can't find where this occurs???

From the endpoints table, there is a link to the CellExpression schema

bcorrie:
> * `CellExpression` schema seems to be missing from the AIRR Data Model with the other schemas.

Sorry - not clear what is missing here. Is this in the docs or in the actual schema?

schristley:
In the docs, https://docs.airr-community.org/en/adc-docs/datarep/overview.html

javh:
> So are the schemas that we are adding for v1.4 really still "Experimental" as per the above Docs pages? How do they become not "Experimental" if not through an official release?

Yeah, they are still experimental. It's fuzzy, but they exit "Experimental" when they become stable. Ie, we stop messing with them except for minor updates, we start caring about breaking backwards compatibility, they are well tested, etc.

They enter "Experimental" when we put them in a release and make them available to the public to experiment with. "Experimental" status is a disclaimer that the schema isn't stable.

bcorrie:
> > So are the schemas that we are adding for v1.4 really still "Experimental" as per the above Docs pages? How do they become not "Experimental" if not through an official release?
> 
> Yeah, they are still experimental. It's fuzzy, but they exit "Experimental" when they become stable. Ie, we stop messing with them except for minor updates, we start caring about breaking backwards compatibility, they are well tested, etc.
> 
> They enter "Experimental" when we put them in a release and make them available to the public to experiment with. "Experimental" status is a disclaimer that the schema isn't stable.

OK, gotcha... So a goal would probably be to have all of our current "Experimental" schemas/apis as non-Experimental for the 2.0 release.

javh:
Ideally, yeah, they are experimental for a release then they become production ready in the next release. But, that doesn't seem to happen often in practice. `Alignment` is still experimental and probably always will be, because there doesn't seem to be much of a use case so there's no point in finalizing it (and no real way to test it if we did).

bussec:
@bcorrie LGTM, do you want to merge this before or after I merge #404?

bcorrie:
Sounds like less work for me if I merge first 8-)

bcorrie:
I will pull the trigger now...

PR #610, opened 2022-05-06
Expression

bcorrie:
We overlooked providing an object id for the CellExpression object, realized this when I added the API docs...


bcorrie:
I think this is minor, @schristley @bussec @javh want to review?

bussec (approved):
Looks good, thanks!

bcorrie:
Above fixes due to #608

PR #611, opened 2022-05-09
add germline_set_ref to DataProcessing

### LINKED ISSUE #595: what is the expected workflow/use case for subject-specific germline sets?
If we get to point of using subject-specific germline sets, this has the potential to create many germline sets, potentially one for each subject repertoire in the ADC. Is this the desired design?

By subject-specific germline set I mean a workflow like this:

- A genotype tool is used, e.g. vdjbase, which determines the subject genotype. Alternatively maybe loci sequencing is done to determine the subject genotype.
- Given a standard germline set, remove all of the alleles except the ones in the subject genotype. Use that custom germline set for sequence annotation and downstream analysis.

When that repertoire is loaded into the ADC, so too does the custom germline set.

I think this is the most obvious design but there are alternatives. This might also inform the decision about normal/denormal for `AlleleDescription` in `GermlineSet`. Presumably in the above, there would be many germline sets but the allele descriptions would be the same record across many of them.

schristley:
closes #595

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -2785,6 +2785,13 @@ DataProcessing:
                 set: 5
                 subset: data (processed sequence)
                 name: V(D)J germline reference database
+        germline_set_ref:
+            type: string
+            description: Unique identifier of the germline set and version, in standardized form (Repo:Label:Version)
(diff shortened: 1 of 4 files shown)


williamdlees (approved):
Thanks Scott!

javh (approved):

Issue #278, opened 2019-11-13
locus_species type definition


javh:
There's a new (ish?) field in the Rearrangement spec called `locus_species` that didn't have a `type` property, which caused the docs build to fail on RTD.

I added `type:string` in the spec on master, just to get the docs to build (see 0c02017).

Is this correct?

javh:
From @schristley:

> @javh actually no, the field is an ontology object, $ref: '#/Ontology', so it is an object. Because $ref is specified, there should not be a type. So the doc script needs to be updated to handle that. I guess this is the first ontology field that's been added to Rearrangement.

javh:
Okay, @schristley. I'll fix it. What should the type value be in the TSV then?

javh:
Oh... And did we test that not having a `type` property for schema field doesn't break the R and python reference APIs?

schristley:
The travis checks passes so I assumed everything was okay, but yeah there might be an issue somewhere. I already did a bunch of changes for the repertoire metadata, but there could be a rearrangement specific pathway...

javh:
Just skimming over the code, I kind of expect the python API to work: `schema.Schema.type` returns None if `type` is not found) okay. But, the R API looks like it'll choke: `load_schema` expects "description" and "type" to exist.

schristley:
It's a pull request #260, and it looks like we can revert it. Maybe we should do that until we work out handling ontology in rearrangements in more detail?

javh:
Can we just put in a dummy type for now instead? Will that break anything? Maybe "object"?

Looks like we'll still have to tweak the R package to deal with unrecognized types, but it should probably do that anyway.

schristley:
> Can we just put in a dummy type for now instead? Will that break anything?

Maybe, I haven't tested in the ADC API yet. It uses swagger middleware and will usually complain if something doesn't match the OpenAPI spec.

bcorrie:
> Can we just put in a dummy type for now instead? Will that break anything? Maybe "object"?

If we use a dummy type for an ontology term, in particular in a TSV file, I would recommend using string. In particular because you can have a JSON object as a string in a TSV (e.g. {"locus_species" : {"id":10,"value":"Homo Sapiens"}} - with appropriately escaped quotes of course).

schristley:
@javh @bussec Sorry I missed this in my review, I was assuming these fields were in repertoire metadata...

[...]

PR #614, opened 2022-05-23
Add species information to `locus` fields

### LINKED ISSUE #137: How to indicate xenografted or transgenic animals

bussec:
Fixes #137 

Notes that the `cell_species` part was already fixed in #260. IIRC the `locus_species` part was not merged back then as it created problems with the representation of ontology information in the TSV.

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1210,7 +1210,7 @@ GenotypeSet:
             items:
                 $ref: '#/Genotype'
 
-# This enumerates the alleles and gene deletions inferred in a single subject. Included alleles may either be listed by reference to a GermlineSet, or
+# This enumerates the alleles and gene deletions inferred in a single subject. Included alleles may either be listed by reference to a GermlineSet, or 
 # listed as 'undocumented', in which case the inferred sequence is provided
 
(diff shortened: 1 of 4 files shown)


scharch:
@bussec @javh any reason this was never merged?

javh:
I don't recall why this was schedule for v2.0. Maybe because #362 was?

bussec:
@scharch See the discussion in #278 , but IMO this should be solved now. I will update and rebase this soon.

PR #618, opened 2022-06-28
Support gzipped rearrangement files

grst:
csv files are commonly available in compressed form. This modification allows to read gzip-compressed CSV files directly with `read_rearrangement`


javh:
Thanks, @grst. We'll get this into the next release (soon).

javh (approved):

PR #621, opened 2022-07-07
Minor fixes to ADC API schema files

bussec:
Both ADC API schema files currently define for the `expression_id_response` object (response of the `/expression/{expression_id}` endpoint) a `Cell` property that IMO should be called `CellProperty`. This patch fixes this issue.


bcorrie:
You are correct, those responses should be CellExpression - thanks...

bcorrie (approved):
Thanks - good catch.

PR #628, opened 2022-08-15
Additions and clarifications to ADC API documention

### LINKED ISSUE #623: Behavior of API queries against properties in lists of records
The schema contains several cases in which top-level objects contain an array of objects, e.g., `Subject` can contain several `Diagnosis` records. These lower-level objects then contain several properties. Assuming the following case (ignoring ontologies for the `disease_*` properties for simplicity):

```
* Subject
  * diagnosis
    (Diagnosis record 1)
      * disease_diagnosis: "rheumatoid arthritis"
      * disease_length: "20 years"
    (Diagnosis record 2)
      * disease_diagnosis: "pancreatic ductal adenocarcinoma"
      * disease_length: "6 months"
```
      
Now imagine a user who is looking for long-term survivors of pancreatic cancer (disease duration > 3 years) issues a query to `/repertoire` using the following payload:

```
{
  "filters":{
    "op":"and",
    "content": [
      {
        "op":"=",
        "content": {
          "field":"subject.diagnosis.disease_diagnosis",
          "value":"pancreatic ductal adenocarcinoma"
        }
	  },{
        "op":">",
        "content": {
          "field":"subject.diagnosis.disease_length",
          "value":"3 years"
        }
	  }
	]
  }
}
```

Would the example Subject record above be returned for this query or not?

bussec:


schristley (commented):
The first sentence seems out of place. It's mentioning response data while this section is about queries, and I don't know what "flattened by the query handler" means. I think you are just saying we use JSON as a data format and it supports nested objects and arrays...?

There is an existing section title called _Queries against Arrays_ and this text seems to address a identical? similar? issue. Should the sections be combined? The language seems to suggest to me that it is addressing a larger topic but it really only seems to mention arrays.

I'm also unclear about the document structure. It's titled like an independent section? But I think it should at least be under "Request Parameters" because it's about querying?

Can we stick a table of contents at the beginning of this doc? It's become very long with many sections, it's hard to see the structure and find stuff...

bussec:
From the call: Tabled, @bussec still has to review.

bcorrie:
I have added a bunch of fixes/updates so that the API docs are consistent across all API endpoints. This included some consistency fixes to the adc_query_support x-airr attribute and the creation of related tables in the adc_api_limits document.

bcorrie:
The adc_query_support changes in the spec for Receptor and ReceptorReactivity were done on #705 because that is where the current final version of those object's fields are defined.

bcorrie:
Fixes #661

bcorrie:
I believe #661 is fixed

bcorrie:
I believe #623 is fixed

bcorrie:
I believe fixes related to #705 are fixed

bcorrie (approved):
I think this looks good to me...

bcorrie:
@bussec I think this looks good. Any chance you could weave your github mastery to resolve the conflicts above? The conflicts are because we split adc_api.rst into multiple files. 

If I don't get a thumbs up from you in a few days I will attempt it - no worries - I just think you will probably be able to do it more quickly 8-)

bcorrie:
We will still need to do `Receptor` and `CellReactivity` changes a bit, after #705 is finished, but I don't think we need to wait this for those changes to be finalized.

bcorrie:
@bussec I am working on resolving conflicts...

bcorrie:
@bussec on second hand maybe you could... I don't like the squash/merge model I don't want to mess things up... The conflicts look minor but the changed files that it is showing when I try to merge are substantial so not what I would expect.

bcorrie:
OK, I am tackling the merge again @bussec

bcorrie:
OK, I think this is done, I will merge early next week unless issues are raised...

docs/api/adc_api_requests.rst
```diff
@@ -0,0 +1,625 @@
...
+
+If the end result that is required it to find all disease diagnoses of "pancreatic ductal adenocarcinoma"
+that have a disease length of over 10 years, searching for ``disease_diagnosis = pancreatic ductal adenocarcinom`` and ``disease_length > 10``
+will result in the above Subject being returned, even though the subject has not had pancreatic ductal adenocarcinom for more than 10 years.
+This is because there is a diagnosis of pancreatic ductal adenocarcinom and a disease length
+of more than 10 years but from a different diagnoses. This is a correct response to the query, but does not return the desired outcome.
+
+In order to acheive the desired outcome, it is necessary to search for one of the conditions (e.g. ``disease_diagnosis = pancreatic ductal adenocarcinom``),
+compile a list of ``repertoire_ids`` that meet that condition, and then search for the second condition (e.g. ``disease_length > 10``)
+across those ``repertoire_ids``.
```
bussec, line 348: @bcorrie I don't have strong opinions on the "global" vs. "local" handling of such requests, but I would like to note that the text here reads different from what we [discussed in the ticket](https://github.com/airr-community/airr-standards/issues/623#issuecomment-1214496963). I you agree, we should note this in the issue, too.

bussec:
Besides the comment above looks good to me, changes regarding the decision to remove `ReceptorReactivity` will be handled in #705 . I can take a stab at the merge conflicts.

bcorrie:
@bussec how many PhD's does it take to resolve a merge conflict... 8-)

I figured it out 8-)

docs/api/adc_api_requests.rst
bcorrie: @bussec the above documentation reflects this comment: https://github.com/airr-community/airr-standards/issues/623#issuecomment-1208073359

Or at least it was meant to. Not sure I am clear what you think is missing...

bussec: @bcorrie Yes, it reflects your comment, which however pre-dates our discussions on this topics in [call 2022-08](https://github.com/airr-community/airr-standards/wiki/Minutes_Standards_2022-08), which is what I was referring to. My recollection of the discussion was that "local" behavior is what most users would expect, and that therefore it is the way how the API should behave. In addition, the potential problems resulting from such an implementation with searches requiring a "global" behavior, should be documented and potential work-arounds be provided. Again, we can reverse this decision, but then this needs to be done explicitly, we should not just drop it.

In addition, from a logical side, I have problems understanding why the procedure described in lines 346-348 should yield the desired outcome -- i.e., simulating "local" behavior, which would **not** return the example record -- as the first query would clearly return the repertoire containing this record, but so would the second query, even when filtering on those repertoires which were already returned by the first query.

bussec: Fine with me, let's get this thing merged ;-)

PR #629, opened 2022-08-16
DataFile and RepertoireGroup modifications

javh:
Per #581, `DataFile` changes:
* Remove `AlleleDescription`, `DataProcessing`, and `Tree`
* Add `GenotypeSet`

Per the August 8, 2022 call:
* Rename `RepertoireGroup` to `RepertoireSet` for consistency with other schema.

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -345,44 +345,16 @@ DataFile:
                 nullable: false
         Repertoire:
             type: array
-            description: List of Repertoire metadata
+            description: List of repertoires
             items:
                 $ref: '#/Repertoire'
             x-airr:
                 nullable: false
-        RepertoireGroup:
+        RepertoireSet:
             type: array
-            description: List of Repertoire groups
+            description: List of repertoire sets
             items:
-                $ref: '#/RepertoireGroup'
-            x-airr:
-                nullable: false
-        AlleleDescription:
-            type: array
-            description: List of allele descriptions
-            items:
-              $ref: '#/AlleleDescription'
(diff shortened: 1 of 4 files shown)


schristley (approved):

schristley:
hopefully you can merge this and #581 without too much hassle

javh:
Yes, I hope so. Once this is in, then I can finish the R package and #581 should be good to go.  Everything else is done with it. (As near as I can remember anyway - I'm sure I forgot something and we'll need to patch release.)

bussec (approved):

scharch:
I don't think I understand why `Clone` is a top-level object but `Tree` isn't. Neither is an object which is "within" another object (at least if I am reading the schema correctly).

javh:
Good point. It doesn't look like it's setup that way in the schema. Should it be?

Though, my motivation for cutting is more so I wouldn't have to figure out how to parse `Tree.properties.nodes.additionalProperties`.

javh:
Also, @schristley and I were on the fence about cutting `Clone`, so I went with "no".

scharch:
> Should it be?

I.... need to think more. Let's leave it for now and revisit for 2.0?

scharch (approved):

javh:
> I.... need to think more. Let's leave it for now and revisit for 2.0?

Seems wise.

schristley:
> > Should it be?
> 
> I.... need to think more. Let's leave it for now and revisit for 2.0?

It's that way now because there was the use case that multiple trees could be defined for a clone, so instead of embedding a single tree into the clone object, `Tree` allows multiple to be defined.

PR #633, opened 2022-08-18
v1.4 doc updates

schristley:


javh:
I just pushed final (from my perspective) edits to the R and python library docs. Build is here:
https://docs.airr-community.org/en/doc-1.4-fixes/

I made one actual change to the interface functions in the libraries, which was to rename the `adf` argument to `model` under the reasoning that the name and purpose of `DataFile` may change in v2.0. I don't love the name `model`, but it will let us convert it from a boolean to a string so we can do stuff like `model="DataFile"`, `model="Manifest"`, etc for different use cases without breaking backwards compatibility for the default behavior (which is currently what's in `DataFile`). I don't know what we'll do with all that for v2.0 - just trying to future proof against argument name changes.

bcorrie (approved):

bussec (approved):

lang/python/README.rst
```diff
@@ -140,40 +142,34 @@ keeping all of the existing fields from the original file. The
 Validating AIRR data files
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The ``airr`` package can validate repertoire, rearrangement, germline set
-and genotype data files
+The ``airr`` package can validate AIRR Data Model and Rearrangement TSV files
```
scharch, line 145: This is super nit-picky, but this sounds like "Data Model TSV files and Rearrangement TSV files"...

scharch (approved):

williamdlees (approved):

lang/python/README.rst
javh: Good catch. Fixed in f2cb874.

javh (approved):

PR #643, opened 2022-09-12
V1.4.1 adc api patch

bcorrie:
Should be all the fixes required


bcorrie:
@javh I think this is good...

bcorrie:
We can delete #612 once merged...

PR #651, opened 2022-10-26
Example implementation of MHCGenotypeSet

jday1:
See #652 for information about this PR.


bcorrie:
@jday1 it looks like this problem was solved with #695 - is  that correct. Can we close this?

PR #668, opened 2023-01-23
Fix representation of information on ontology nodes in Docs tables

### LINKED ISSUE #662: Ontology fields use "value" rather than "label"
See here: https://docs.airr-community.org/en/stable/miairr/data_elements.html
for at least `collection_time_point_relative_unit` but I think this is all fields.

This:
"Ontology: { top_node: { id: NCIT:C63536, value: Study}}"
should read
"Ontology: { top_node: { id: NCIT:C63536, label: Study}}"

bussec:
Fixes #662

### DIFF
--- a/.github/workflows/py-unittest.yaml
+++ b/.github/workflows/py-unittest.yaml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


javh:
Bump actions python version to v3.8. v3.6 is EOL.

Issue #671, opened 2023-02-01
Consistent list of possible Ig/TCR loci throughout the schema


bussec:
Field that encode Ig/TCR loci are implemented as controlled vocabularies (i.e., `enum`) throughout the schema. However, the list of available values differs between `Rearrangement`.`locus`, `PCRTarget`.`pcr_target_locus`, `AlleleDescription`.`locus` and `GermlineSet`.`locus`, as the former two contain `IGI` as additional option (which is an additional Ig light chain locus in bony fish).
Whether by addition or removal, the enum lists should be consistent throughout the schema.

javh:
Addressed in #674.

Issue #663, opened 2022-12-13
genotype locus field has nullable:false


schristley:
This causes a validation error when a new blank template repertoire is created. Either this field needs to be `nullable:true` or given a default value.

```
Repertoire at array position 1 with validation error: Non-nullable field "subject.genotype.receptor_genotype_set.genotype_class_list.locus" is null or missing
```

schristley:
Actually there a bunch of fields in `Genotype` and `MHCGenotype` with this issue.

javh:
Same issue as #665.  Fields should be set `nullable:true` unless they are identifiers or mandatory references, though null values may be not recommended. Not logical default for locus. @williamdlees , please confirm.

javh:
Also, check all germline nullability fields.

williamdlees:
Perhaps I have misunderstood, but I have a feeling that we are putting the cart before the horse here. The way it's defined, a genotype *has* to have one of the seven values listed for locus to mean anything. In a data file, finding a genotype with a null locus, or a locus with some value other than those listed, should trigger an error. To me, the validation being mentioned here is being done at the wrong point: it should be done when the record is complete, not when it is in the process of construction. Alternatively, the mandated fields could be passed as mandatory arguments to a constructor, or initialised in a dict when the object is first created.  The implementor will know what the locus should be set to when the object is first created. Advising users to create blank records and then fill in the blanks is an open invitation to leave out necessary information, and weakening validation makes this worse.

If limitations in the way things have been implemented mean that this is unachievable, then the field can of course be changed so that it is not mandatory but I feel this shouldn't go without challenge, both because it weakens the validation quite considerably and secondly because it sends a message to implementors that the field is optional, when it really isn't.

schristley:
> In a data file, finding a genotype with a null locus, or a locus with some value other than those listed, should trigger an error. To me, the validation being mentioned here is being done at the wrong point: it should be done when the record is complete, not when it is in the process of construction.

This is valid point and one that we discussed when first introducing `Repertoire` and its identifiers. The same logic applies, if you have a data file with repertoires that don't have a `repertoire_id`, is it valid or invalid? Complicating the scenario is that there are numerous situations where the id's may be legitimately blank because they just haven't been assigned yet in some process. Now you need to make the decision, can data files sometimes be invalid and sometimes be valid? We decided that it's simpler for the user if they can always assume that data files should be valid. This then implies that blank template objects also need to be valid.

[...]

Issue #659, opened 2022-11-09
collection_time_point_relative has string example instead of number


schristley:
```
        collection_time_point_relative:
            type: number
            nullable: true
            description: Time point at which sample was taken, relative to `Collection time event`
            title: Sample collection time
            example: "14"
            x-airr:
                miairr: important
                adc-query-support: true
                set: 2
                subset: sample
                name: Sample collection time
```

should have `example: 14`

Issue #656, opened 2022-11-01
AIRR Data Model identifiers in the new draft objects not marked as identifiers


schristley:
In many of the new draft objects, the AIRR Data Model identifier fields are not tagged with the `identifier: true` attribute in `x-airr`. This attribute is needed so that valid data is returned from the ADC API.

javh:
We should make a pass over the `_id` fields to make sure all that should be `identifier: true` are set as such.  IIRC, `required: true` & `nullable: false` implies `identifier: true` for `_id`... Is there any reason not to be explicit and add the tag?

bcorrie:
@schristley can you elaborate on which fields you think should have this characteristic.

I see that `cell_id` in the `Cell` object does not have `identifier:true` which seems wrong to me. But I wouldn't say that that is the case for all `_id` fields. Do you have a list that are causing you issues? How is your middleware using these fields?

schristley:
I think the fields to mark as identifier will be obvious as you scan through. It's those fields required to represent the relationships between AIRR objects. I tend to ask myself how would the code look if I needed to access some random field, say `mhc_genotyping_method `, what identifiers would I use?

PR #674, opened 2023-02-19
Various schema updates

### LINKED ISSUE #677: De-nest Receptor.reactivity_measurements
`Receptor.reactivity_measurements` is a nested object. Move to a standalone object, such as `ReceptorReactivity` and embed via `$ref` instead.

### LINKED ISSUE #679: Copyright on docs.airr-community.org is not current
It has been pointed out to me that our copyright ends at 2021.

### LINKED ISSUE #686: `discriminator` keyword is being used improperly in the schema
I discovered this when using a JSON validator. We have a `discriminator` keyword on all of our objects, however its usage is not correct. It's meant to be used with `oneOf` where the `propertyName` in the `discriminator` indicates what property can be used to distinguish the appropriate object in the `oneOf` union.

Because we aren't using `oneOf` anywhere, we should essentially delete all of the discriminators.

javh:
Fixes #677
Fixes #679 
Fixes #686 

1. Set enum fields in Germline schemas to `nullable: true`. (#663)
2. Set `_id` fields in Germline schemas to `identifier: true`. (#656)
3. Added "IGI" to all locus fields. (#671)
4. Added quotes around values for various string enums.
5. Minor formatting issues. (#659)


javh:
From the call:
+ Add `identifier: true` to `Acknowledgement.acknowledgement_id`, `Cell.cell_id`, `CellExpression.expression_id`, `Receptor. receptor_id`, `Clone.clone_id`, `Tree.tree_id`, `RepertoireGroup.repertoire_group_id`
+ `CellExpression.reperotire_id`, is misspelled in `required`.

javh:
From the call:
* Need to address `identifier` issues from prior call notes.
* `nullable` has been addressed in other PRs (by @williamdlees), so we need to defer to those changes instead of those in this PR.
* Check for `snake_case` in all enums. Enforce `snake_case` instead of requiring quotes around enums, except in cases of currently in-use fields, which will require a version bump to change the enum values. Require quotes for in-use fields with spaces in enum values.
* Resolve merge conflicts, propagate schema to `/lang`, and merge.

javh:
Addressed issues above except for the enum `snake_case` topic, which is moved to #673.

PR #680, opened 2023-03-20
Update nullable fields, add additional fields to AlleleDescription

williamdlees:
In germline objects, remove all nullable: false entries and add x-airr: qualifiers instead.

- Move aligned sequence to SequenceDelineationV and change co-ordinates in SequenceDelineationV to be against the unaligned sequence rather than the aligned sequence. This makes AlleleDescription delineation-agnostic and removes reliance on an alignment for the co-ordinates
- Add fields to AlleleDescription for Allele Similarity Cluster designation


williamdlees:
fixed UndocumentedAllele name

williamdlees:
Withdrawn while I work on updating the test data

bussec (commented):
@williamdlees please find my comments below:

* The following properties should IMO be `essential`:
    - `RearrangedSequence.derivation`
    - `RearrangedSequence.observation_type`
    - `AlleleDescription.species`
    - `GermlineSet.species`
    - `MHCGenotype.mhc_class`
* `AlleleDescription.coding_sequence.description`: Is it intended to refer to IMGT Ontology here?
* `AlleleDescription.allele_similarity_cluster*` needs more description and an example. The description should clarify whether "similarity" is different from "identical" and if yes to which degree.
* Could `Genotype.receptor_genotype_set_id` be renamed to `genotype_set_id` to avoid confusion with the `receptor_*` properties of the `Receptor` object.
* MHCGenotype.mhc_alleles is an array-of-objects (and not an array-of-refs)


* As a note: According to [wiki](https://en.wiktionary.org/wiki/derivation#Noun) the term "derivation" does not refer to the source of the process (or its at least ambigous).

javh:
From the call:
* `species`, `mhc_class`, and `observation_type` as essential makes sense.
* Check with IARC on `RearrangedSequence.derivation`.
* `coding_sequence.description` can we be more explicit here? Can we use the following as a guide? https://docs.airr-community.org/en/stable/miairr/specification_miairr_ncbi.html.
* Can't change `Genotype.receptor_genotype_set_id` in a minor release. Maybe v1.5.
* Replacement of object (`MHCGenotype.mhc_alleles`) with reference should be fine for v1.4.2.

williamdlees:
- `species`, `mhc_class`, and `observation_type` marked as essential.
- No strong views from IARC regarding RearrangedSequence.derivation (the name waa reviewed and agreed by IARC some years ago)> Possible alternatives should they be required could be `derived_from`, `sample_source`, `nucleic_acid_class.`.
- `coding_sequence.description`changed to `nucleotide sequence of the core coding region, i.e. the coding region of a D-, J- or C- gene, or the coding region of a V-gene excluding the leader`. I think this is pretty specific now.
- object (MHCGenotype.mhc_alleles) replaced with reference to a new top-level object MHCAllele.

bcorrie:
Re-structuring of MHCGenotype and Genotype objects to include object references rather than embedded objects looks like it is correct to me.

bcorrie (approved):
MHCGenotype and Genotype object restructuring (no more embedded objects) looks like it is correct to me.

javh:
From the call:
* Add additional description and clarification for `AlleleDescription.allele_similarity_cluster*` in the Germline documentation (see above comment from @bussec).

PR #685, opened 2023-04-14
Fix Issue #684

jday1:
See https://github.com/airr-community/airr-standards/issues/684 for context.

Introduced GenotypeSets to allow testing of mhc_genotype_set and receptor_genotype_set.

Other minor bug fixes.

PR #695, opened 2023-06-12
Copy of PR #685

javh:
Copy of #685 so I can fight with merge conflicts.


javh:
Merge conflicts are resolved, but there are python unit test failures.

PR #706, opened 2023-07-24
Remove functional NULL values from subject.sex

### LINKED ISSUE #670: Remove functional NULL values from `subject`.`sex`
`subject`.`sex` is the only property that still defines the terms `not collected` and `not applicable` in its enum list. This seems to be a leftover from the days when this field was still considered to be what is now "essential". As it is now only considered to be "important" and is `nullable: true`, the use for the two enum terms above is rather limited. Should we remove them?

bussec:
Fixes #670

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1766,8 +1766,6 @@ Subject:
                 - pooled
                 - hermaphrodite
                 - intersex
-                - "not collected"
-                - "not applicable"
             description: Biological sex of subject
             title: Sex
             example: female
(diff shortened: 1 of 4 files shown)


schristley:
@bussec `physical_linkage` also has a NULL-like value (none), change it to null?

bussec:
@schristley No, as `physical_linkage` is `x-airr.miairr` = `essential` and thus not nullable. Also, `physical_linkage` being `none` has a different meaning ("it is NOT linked") than `null` ("no information").

javh (approved):
`null` is added in #708, so this looks good to merge as is.

PR #708, opened 2023-07-25
add null enum for nullable fields

### LINKED ISSUE #697: nullable flag and enum has validation issues
While working on airr-js, I was getting validation errors for fields with `nullable: true` and `enum` when the field value was `null`. I'm using the ajv package for validation, instead of custom written like our python library. Apparently, there has been various discussions about this as OpenAPI schema and JSON schema aren't 100% equivalent. From what I can gather, the [conclusion](https://github.com/OAI/OpenAPI-Specification/issues/1900#issuecomment-578409376) was that `null` needs to be added to the `enum` list. When I do that, as shown below, then validation passes.

```
        sex:
            type: string
            enum:
                - male
                - female
                - pooled
                - hermaphrodite
                - intersex
                - "not collected"
                - "not applicable"
                - null
            nullable: true
            description: Biological sex of subject
            title: Sex
            example: female
            x-airr:
                miairr: important
                adc-query-support: true
                set: 1
                subset: subject
                name: Sex
                format: controlled vocabulary
```

This really only applies to the OpenAPI V3 spec. This is a simple fix but it might effect UIs that display those enum lists.

schristley:
fixes #697 

Also fixes a few fields in V2 spec missing `nullable` flag

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1044,7 +1044,6 @@ AlleleDescription:
                 enum:
                     - likely_truncated
                     - likely_full_length
-                    - null
             description: Controlled-vocabulary tags applied to this description
             x-airr:
                 nullable: true
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -535,19 +535,23 @@ RearrangedSequence:
             enum:
                 - DNA
                 - RNA
+                - null
             description: The class of nucleic acid that was used as primary starting material
             x-airr:
                 miairr: important
+                nullable: true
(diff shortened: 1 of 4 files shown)


bussec:
@schristley Properties that are `x-airr.miairr` = `essential` are by definition not nullable.

schristley:
@bussec I was wondering about that, why fields were marked essential but had `nullable: true`. Are the `miairr` tags on the germline fields correct and I can fix their nullable state, or are they still being debated?

schristley:
@bussec I'll fix the schema in a separate PR so it can merged for 1.5 and add some additional consistency checks as it seems a few things are being missed, and the two schema versions are getting increasingly out of sync and/or inconsistent.

schristley:
removed null for essential fields

PR #719, opened 2023-08-08
Property type redo

bcorrie:
Redo of property_type fixes, with just property_type field but adding detection keywords to suggested strings.

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -4400,9 +4400,11 @@ CellExpression:
         property_type:
             type: string
             description: >
-                Keyword describing the property type and detection method used to measure the property value. 
-                The following keywords are recommended, but custom property types are also valid: "rna_umi",
-                "surface_protein_barcode", and "antigen_barcode".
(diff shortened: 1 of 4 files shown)


bcorrie:
Changes as I think we discussed, @bussec I added three possible values, the three former types with a detection method of count, which is I think applicable for our use case. Can you add some other possibilities (and correct mine if in error).

javh (approved):
I pushed some suggested terms: "rna_umi", "surface_protein_barcode", and "antigen_barcode".

I also set `property_type` as `required` and `nullable: false`.

Yay? Nay? I'm good to merge if those changes are okay.

scharch:
@javh it's a nit, but all 3 are UMIs. Also, I'm not crazy about calling them "barcodes" instead of just `surface_protein` or something. But I don't care enough to make an issue of it if others feel differently.

bussec:
I now added "fluorescence_intensity" as the measured parameter and added `_by_` as a human-readable separator between the property type and the detection method.

bussec (approved):

javh:
@scharch I agree with you. But, I we decided to make the field a combination of readout and method.

@bussec I'm good with those changes.

Issue #723, opened 2023-08-22
Validate AIRR objects against AIRR validation 'levels'  


jday1:
MiAIRR has 3 levels of compliance.

```
- essential
- important
- defined
```

And I would argue, implied is a 4th level which is `none`. I.e. the AIRR object is not MiAIRR Compliant.

I want to introduce logic to check an AIRR object against each of the 4 levels of compliance which does not currently exist.

I.e. the end goal is to have a `validate_object` function which looks like

```
def validate_object(self, obj, missing=True, nonairr=True, context=None, miairr_compliance_level="essential"):
```

This would allow a user to validate the compliance level of their AIRR object.

The first step in this for Synteny is to allow the case where MiAIRR compliance is not met. This is done is PR https://github.com/airr-community/airr-standards/pull/724

PR #724, opened 2023-08-22
Issue 723: Add MiAIRR compliance flag

jday1:
This PR implements the first step for https://github.com/airr-community/airr-standards/issues/723.

A further benefit of this PR is that it fully aligns SyntenyBio's fork with airr-community which will allow us to the airr package directly in the future. This has been ongoing work at Synteny for several months and with this merge and release it will allow us to more actively contribute to AIRR.


javh:
Thanks!  Given that we are in feature freeze to get the v1.5 release out, this might not make it into v1.5.  I'll take a closer look to be sure.

If not, there will be a lot of merges after the release and we can work this into that.

javh:
From the call:
* No objection to the additional validation level.
* Check if this is the best implementation approach. Eg, instead change the behavior of `validate` to multiple levels (`none`, `miairr`, `strict`, whatever).

lang/python/airr/schema.py
```diff
@@ -42,7 +44,7 @@ class Schema:
     _to_bool_map.update({x: False for x in false_values + [0, False]})
     _from_bool_map = {k: 'T' if v else 'F' for k, v in _to_bool_map.items()}
 
-    def __init__(self, definition):
+    def __init__(self, definition: Union[str, dict]):
```
scharch, line 47: What is this change doing?

lang/python/airr/schema.py
```diff
@@ -335,15 +338,19 @@ def validate_object(self, obj, missing=True, nonairr=True, context=None):
         # first warn about non-AIRR fields
         if nonairr:
             for f in obj:
-                if context is None: full_field = f
-                else: full_field = context + '.' + f
+                if context is None:
+                    full_field = f
+                else:
+                    full_field = context + '.' + f
```
scharch, line 344: @javh @bussec what are the guidelines (if any) for style changes like this?

jday1:
I realised this was a duplicate of https://github.com/airr-community/airr-standards/pull/750. I've closed this PR in favour of the other as we have made more progress on that.

PR #725, opened 2023-08-22
Move Docs terminology into tabular format, update content

bussec:
The AIRR Docs contain several, but not all, terms of the AIRR Terminology. To create a single source of truth, the current terminology is added in tabular (CSV) format. The previous table in the RST is replaced by a reference to the CSV.


scharch (approved):

bcorrie:
I think this is an AIRR 2.0 issue, not ADC 2.0???

bussec:
@bcorrie Not sure whether it is actually an AIRR 2.0 issues, but I agree that it is not one for ADC 2.0 ;-)

PR #731, opened 2023-11-03
Allow missing values in logical fields when validating tabular data

kelliemac:
This is regarding the R package.

Currently, when I use airr::read_rearrangement() I will get warnings like "Warning: vj_in_frame is not logical for row(s): ..." when the only values in that column are TRUE, FALSE, and NA. This stems from the fact that checks on the logical fields currently only allow TRUE and FALSE, not NA. I think that NA should be a valid value in a logical field.


javh:
Thanks for submitting this PR @kelliemac. I'm sorry this one slipped through the cracks. This looks good to me.

kelliemac:
Thank you @javh ! I'm happy that this package exists and glad to be able to contribute in a small way.

PR #746, opened 2024-02-06
Change pub_ids to be array of CURIEs

### LINKED ISSUE #740: Make Study.pub_ids an array of strings
Currently `Study`.`pub_ids` is a single string and it is unclear how multiple references would be separated (probably by comma, but is this documented). It seems appropriate to change this into an array of strings.

bcorrie:
closes #740

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1145,9 +1145,13 @@ GermlineSet:
             x-airr:
                 miairr: important
         pub_ids:
-            type: string
+            type: array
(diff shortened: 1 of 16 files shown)


bcorrie:
Fixes have been made, @bussec do you want to review? Closes #740

bussec:
@bcorrie Looks good to me in general, two points regarding consistency:
1. Could we used bracket notation [ foo, bar ] for the YAMLs, instead of the single lines?
2. Could we use the same two identifiers throughout the schema?

bussec (approved):

Issue #667, opened 2023-01-23
Genotype and MHCGenotype are too deeply nested


javh:
Both Genotype and MHCGenotype are nested more deeply than typical for the schema.  `MHCGenotype.mhc_alleles` and and `Genotype.*_alleles` could be pulled out into top level objects with refs in Genotype and MHCGenotype.

williamdlees:
Making receptor_genotype_set and mhc_genotype_set independent properties of Subject makes sense to me as these are independent objects.

I am less sure of the logic behind splitting documented_alleles, undocumented_alleles and deleted_genes into high-level objects., These are inherent related components of the genotype. Is there a need to reduce the depth of nesting to a specific level? If so could you articulate the reason please, so that we can consider options.

javh:
The reason is mostly stylistic. In the rest of the schema, objects live in other objects via `$ref:`s not by nesting the definition. They would still be inherent related components if their definition was at the top level. But, it should be easier to validate if we use the same style throughout.

Aside from that, we might be able to apply some abstraction to `documented_alleles`, `undocumented_alleles`, and `deleted_genes`. Not sure on that. We'd have to discuss.

javh:
Thanks, @williamdlees. I think for this particular issue, we can just focus on the style fix and maybe the allele abstraction question.  The validation should follow from that. If the style is right, the validation should be fine, but, if not, then we'll need to fix the validation.

schristley:
> * `undocumented_alleles.allele_name` should be renamed to `label`, but we'll need to check if this breaks backwards compatibility in the ADC and therefore needs to wait until v2.0 (likely yes).

@javh Can you deprecate and add `label` in with v1.4.2? That way I can start using `label` now.

bcorrie:
Deprecation and changing a label breaks backwards compatibility so it shouldn't be part of v1.4.2, no? Any field name/type changes shouldn't be a part of a minor release...

schristley:
> Deprecation and changing a label breaks backwards compatibility so it shouldn't be part of v1.4.2, no? Any field name/type changes shouldn't be a part of a minor release...

Right, yeah minor release should just be software and minor fixes. Could be a v1.5.x then. I wasn't sure if saying v2.0 because the change required a new version of MiAIRR, or there's just an expectation to get v2.0 released before a v1.5...

bcorrie:
> * De-nesting Germline schema: De-nesting shouldn't break anything, so we can include this is v1.4.2.

[...]

PR #749, opened 2024-02-12
Addition of Diagnosis TimePoint

### LINKED ISSUE #735: Why doesn't Diagnosis have a time_point associated with it???
A subject can have multiple diagnoses, and in fact diagnoses are not really diagnoses at all, since they can actually be "interventions" - for example an immunization.

Shouldn't a diagnosis have a time_point_relative and a time_point_relative_reference?

bcorrie:
closes #735


bcorrie:
This is a minimal change to make use of the Event object in the current structure that we have. No real semantic or structural changes, just syntax changes so that we are using the `Event` object everywhere an "event" occurred in our current specification. 

With this specification, we still have an array of `Diagnosis` where each `Diagnosis` has an `Event` that captures the `Event`-like features of a diagnosis. We do not have a general array of `Events` in `Subject`.

To do this correctly, with an array of `Events`, I think is beyond the scope of v2.0. In talking with @schristley for the AKC, it makes more sense for this to wait until v2.1.

So for v2.0 I would suggest we either make some simple changes to introduce the `Event` concept in v2.0 similar to what we have now, or if we want to consider the redesign of this we should move this issue out of the v2.0 milestone and into v2.1

I am fine with either. The nice thing about doing this now is that we add Event information like TimePoint to Diagnosis, which is currently missing. But there is a good argument for doing this once and doing it correctly - which I don't think fits in the scope of v2.0.

bussec:
I agree that the "maximalist" version of this is a substantial change and that it will require time and effort to get it done correctly. However, if we take the meaning of a major version number seriously, then IMO it should go into v2.0 exactly for that reason and not into a later point release. If we keep moving more and more stuff to v2.1, we might be able to release v2.0 in time, but will not meet the goal of having only non-major changes afterwards. In the end we need to agree on what we want/need to complete before we are "done" with the schema and what our timeline is. I have added this topic to the agenda of our next call.

bcorrie:
Yep, but at the same time, if we keep adding major new things we will never release 2.0. 

I think the more we see things being pushed to 2.1 the less I am convinced that "having only non-major changes afterwards" might not be practical - unless you are willing to accept that the current WG will be around for another year - and the v2.0 release is now targeted at the 2025 meeting 8-)

bcorrie:
So we are back to the question what is in v2.0 and what is for later discussion.

Currently we have a minimalist change (https://github.com/airr-community/airr-standards/pull/749/files):

- We have added an Event object
- We added a `diagnosis_event` field of type `Event` that helps capture when a diagnosis happened. This did not exist before, there was no timepoint associated with a `Diagnosis` (See #735 )
- Changed `collection_time_point_*` to be a `collection_event`

These are relatively minor changes that clean up several issues in the original spec.

There are no major changes here IMO, and I would lean towards doing this for v2.0 (with minor extra tweaks) and leave the "re-engineering" of the concept of events in a study to a later discussion/release.

schristley:
I consider this a fairly major semantic change. Adding a time point seems reasonable, but inserting the Event object in the middle will create a lot of churn when we change things later.

> leave the "re-engineering" of the concept of events in a study to a later discussion/release.

IMO it doesn't. If forces the Event object to be used now in all the code. Every data entry screen has to be written to capture an "Event" instead of just recording a time point. Every python script and analysis tool has to write code that traverses through that "Event" in order to access a time point. Every query has to be written to traverse through that "Event" object to operate on the time point. And if all of that gets done, then we will be stuck with this structure because changing it is [too much work](https://github.com/airr-community/airr-standards/issues/667#issuecomment-1936440433). I can already see you writing the post later about how it breaks backward compatibility ;-)

I'm not opposed to adding an Event object. I would just prefer it be added in a way that doesn't require mandatory usage now with all the associated data and code changes. Even more so if this is just considered preliminary and we expect the design to change later.

Another idea which will create less churn is to put the `Event` object inside `TimePoint`, that is flip the structure, so the new fields in Diagnosis are `TimePoint`. Here we are just reversing the relationship, instead of saying an Event occurs at TimePoint, we are saying TimePoint is associated with Event. This would allow usage of the time point without requiring an event object to always be in the middle.

bcorrie:
I am also OK to move this entire change out of v2.0 - I see your point if it is just going to change in the next version why do work now that will be thrown away in the near future.

So I would be OK with moving this out of the v2.0 milestone completely.

bcorrie:
Remember this all started because `Diagnosis` doesn't have a `TimePoint`. It hasn't had a time point for 6 years (since MiAIRR was introduced) and it hasn't appeared to be a problem to anyone up until now except to us, so I am OK for it NOT to have a time point for another year 8-)

bcorrie:
> Another idea which will create less churn is to put the `Event` object inside `TimePoint`, that is flip the structure, so the new fields in Diagnosis are `TimePoint`. Here we are just reversing the relationship, instead of saying an Event occurs at TimePoint, we are saying TimePoint is associated with Event. This would allow usage of the time point without requiring an event object to always be in the middle.

The easiest thing to do is just add a `TimePoint` to the `Diagnosis` object. This doesn't even break backwards compatibility, it just adds a field. We can revert the changes to  `collection_time_point_*` and drop `Event` for now.

This actually fixes #735 without changing anything else.

bcorrie:
@bussec any more thoughts. I am not going to do any more edits on this until we decide what we actually want for v2.0... 8-)

bussec:
@bcorrie I am ok with only doing small changes until mid-2024 (whatever that version will be called). Right now I have no opinion whether a `TimePoint` should have `Events` or an `Event` should have a `TimePoint`, but I agree with @schristley that if we only want to do small changes now, the former one is clearly (plus addition of a `TimePoint` to `Diagnosis`) is probably the least invasive thing to do.

Having said this, I still think that the whole "description-of-what-happend-when" package (including issues like #486 ) needs to be addressed before we go on hiatus, as it substantially limits the usefulness of the AIRR Schema if we don't.

bcorrie:
@bussec @schristley I have reverted the Event changes and added a TimePoint to Diagnosis. This resolves #735 and we have agreed to do the big "event" lift after 2.0. So this can be merged I think.

bussec:
Resolved merge conflicts. @schristley are we good to go?

schristley:
@bcorrie @bussec Minor point, but the test code creates a template repertoire object and validates it, so it will already test `null` for that field. My suggestion is that you put the example values into the "good" test data files. Otherwise it looks good to go.

schristley (approved):

bcorrie:
Added tests for proper validation for R and python as per @schristley request above. Merging...

PR #752, opened 2024-02-16
v2.0 doc changes

### LINKED ISSUE #617: Expected behavior of `facets` on array of string
What is the expected behavior when an ADC API query requests aggregation (via the `facets` request parameter) on a field that holds an array of strings, e.g., `study`.`keywords_study`. For example if such a field holds the array `['A', 'B', 'D']`  should the aggregation 

1. Increase the count for the each of the strings independently `{'A':1, 'B':1, 'D':1}`, or
2. Count the joint occurrence of the strings `{'A,B,D':1}`?

Note that the example provided in the docs does not match this case 1:1 as `pcr_target` is an array of objects, and `pcr_target_locus` contains only a single string.

### LINKED ISSUE #751: Review docs for queries against array objects
This needs to be done for the v2.0 release.

Fixing the query functionality will be a v2.1 release task as per #623

bcorrie:
Fixes #617
Fixes #751

### DIFF
--- a/docs/api/adc_api_requests.rst
+++ b/docs/api/adc_api_requests.rst
@@ -582,6 +582,9 @@ subjects each with two IGH repertoires.
     ]
   }
 
+Note: ADC API facet requests differ from those in the GDC API on which the ADC API is based. In the ADC
+API it is allowed to request a facet count on a field that is being filtered, whereas in the GDC API
+filters on the facet'ed field are ignored (see `Genomic Data Commons (GDC) API Facets`_ restriction #2).
 
 Queries on Nested Information
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


bussec:
@bcorrie If we want to collect all doc changes for v2.0 in this PR, maybe we could move the changes to the ADC API spec files to another PR? This way we could merge them potentially sooner (as they seem uncontroversial to me) and avoid collisions when we try to merge this PR down the road...

bcorrie:
@bussec I originally created this to contain all docs, but realized there were enough ADC API doc changes. So this is all ADC API doc changes, I have changed the title to that effect.

bcorrie:
I plan on finishing the two remaining ADC API doc changes maybe today and then merging this.

bussec:
@bcorrie Ok, thanks for clarifying. PR looks good to me!

PR #758, opened 2024-02-21
Initial moves to OpenAPI v3

schristley:
Start building the base so we only need to edit the OpenAPI v3 spec, and the v2 spec can be generated from v3. Also clean up the test suites so there is one common set of test data files used by all languages.


schristley:
```
$ make

Helper commands for AIRR Standards repository

make gen-v2       -- Generate OpenAPI V2 spec from the V3 spec
make docs         -- Build documentation
make lang-copy    -- Copy spec files to language directories
make data-copy    -- Copy test data files to language directories
make checks       -- Run consistency checks on spec files
make tests        -- Run all language test suites
make python-tests -- Run Python test suite
make r-tests      -- Run R test suite
make js-tests     -- Run Javascript test suite
```

bussec:
Looks good! And we should implement the `make data-copy` function rather sooner than later. #741 will put the the JSON/YAML data files in sync again, but the TSVs still need fixing.

bussec (approved):

bcorrie:
I am not up to speed on the automation, but I am all in favor of keeping the specs in sync but not much I can contribute as reviewer. I will leave it in your hands to approve this, am removing myself as reviewer. 8-)

schristley:
Do we 1) remove the test data file from the lang directories or 2) leave them there? Probably 2) so that the lang directories are "complete" without additional steps. Similar to schema files, so should add consistency checks to verify the test data files match.

schristley:
the problem we will have with generating the v2 spec from v3 is that the output will not be pretty, will lose comments, etc., unless we write a sophisticated translation program, which probably isn't worth the effort.

schristley:
I created conflicts when merging in master incorrectly, so starting over with #767

Issue #545, opened 2021-09-03
need to write doc for async API


schristley:

bcorrie:
@schristley assume you want to resolve this for v2.0?

schristley:
also add doc about `extensions`

PR #759, opened 2024-02-22
extensions property

### LINKED ISSUE #555: how to know if a repository supports an ADC extension?
I was thinking we could add an `extensions` array into the info object, which might be an easy way to programmatically determine if a repository supports an ADC extension. While we only have one extension now with the async API, CRWG has been considering adding the stats API, and we might have more in the future. It could be a simple keyword array like this:

```
extensions:
  - asynchronous
  - statistics
```

schristley:
closes #555

### DIFF
--- a/docs/api/adc_api_overview.rst
+++ b/docs/api/adc_api_overview.rst
@@ -60,6 +60,8 @@ to be followed.
    (timeout) should be used if the API does not complete an operation because of an
    internal time limit, and HTTP 413 (Content too large) should be returned when either
    max_size or max_query_size are exceeded.
+*  Extensions beyond the standard API, e.g., support for the Async API, should be specified
+   with the `extensions` property in the `/info` endpoint.
 
 Repository operation principles
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


schristley:
@bcorrie good to merge?

bcorrie:
@schristley the addition looks good, but it doesn't look like there is any documentation on the asynchronous endpoint. Am I missing them somewhere? It looks like the async spec is kind of an orphan (from a docs perspective at least).

If there aren't any docs there probably should be (an adc_api_extensions.rst), and we probably need a paragraph in the overview page (rather than the bullet point) that describes what extensions are (maybe a "Extensions" section above the "Authentication" section on the overview page) and point the Extensions docs???

schristley:
> @schristley the addition looks good, but it doesn't look like there is any documentation on the asynchronous endpoint. Am I missing them somewhere? It looks like the async spec is kind of an orphan (from a docs perspective at least).

Do you mean this #545 ? Yes, it still needs to be written.

bcorrie:
Ah yes - if that is in the works then I think good to merge this.

bcorrie (approved):

PR #770, opened 2024-03-01
Change germline_set_ref to be a CURIE

### LINKED ISSUE #553: Use CURIEs to link Germline to Repertoire/Rearrangement
Should resolve Task 3 in #157 

- link repertoire/rearrangement to germline. The link is currently at the rearrangement level but might want to reconsider this, maybe at the repertoire level with software processing and annotation.

See https://github.com/airr-community/airr-standards/issues/157#issuecomment-921872075 for discussion.

bcorrie:
Closes #553


bcorrie:
@williamdlees I made an attempt to make germline_set_ref a CURIE. It passes checks, have a look at what I did and see if it seems OK.

bcorrie:
The main issue I see is that the python test data has IDs from IMGT for germline_set_ref of the form:

`"germline_set_ref": "IMGT:Homo sapiens:2022.1.31"`

Would this be referring to one of these: https://www.imgt.org/download/V-QUEST/archives/

bcorrie:
Also, the CURIE PREFIX I used is OGRDB_GERMLINESET. CURIEs are supposed to make things short, and this prefix is long but if you are going to have other things that you want to look up (Alleles?) using a CURIE, we need something descriptive and different.

```
                "allele_description_id": "OGRDB:A00301",
                "allele_description_ref": "OGRDB:Mouse_IGH:IGHV-2DBF",
```

williamdlees:
Thanks Brian. This looks good to me and I think it is the way to go. If other providers of germline sets (I can only think of Mixcr and IMGT for the time being) wish to offer sets in AIRR Standard format, we can encourage them to provide a standardised URL and create a CURIE. But until they do, it isn't really an issue. And we do have other fields for name, date and version that can fill in if necessary.

schristley:
Hi @williamdlees , my plan for our AKC 1-on-1 next week was to review this with you in the context of the OGRDB API because right now what's in the pull request does not resolve to a URL that returns the germline set. We will need to change the PR and/or the API so that it does. It won't be hard to do.

schristley:
> Also, the CURIE PREFIX I used is OGRDB_GERMLINESET. CURIEs are supposed to make things short, and this prefix is long but if you are going to have other things that you want to look up (Alleles?) using a CURIE, we need something descriptive and different.

@bcorrie My suggestion is to not make separate CURIEs for each data type. If you remember how James described it, there is a global part and a local part. `OGRDB:` is a sufficient prefix for everything the global OGRDB service provides. While the local part, @williamdlees will have control over. That design allows OGRDB to provide additional services down the road without requiring new CURIE prefixes, instead the local part can be enhanced.

williamdlees:
I’ve been thinking about Scott’s helpful comments.

 

Currently the field is described as ‘Unique identifier of the germline set and version, in standardized form (Repo:Label:Version)’. I think the overall motivation of this change is to provide a URI that will download the set, rather than this standardized form. That seems a reasonable thing to do, and doesn’t impact any user code out there, because the standardized form isn’t really usable by code today. An example of a URL from OGRDB, which works today,  would be https://ogrdb.airr-community.org/api/germline/set/Human/IGH_VDJ/8/airr_ex.  As I understand a CURIE, it would provide a shorthand that avoids the need to write out the URL in full in the rest of the schema definition. The thing that worries me is that it bakes OGRDB into the definition.

 

We don’t intend OGRDB to be the only source of germline sets. It would be great if others, for example IMGT and MiLabs, started to support the MiAIRR standard for germline sets. However, if they do this, would we need to add CURIES to the schema definition for their systems? To do so would be much more effort on our part than we save by adding a URL shortcut. And I fear it might also discourage them, and send a general message that, for the MiAIRR standard, OGRDB is the single repository for reference sets.

 

Is there a way to use a CURIE optionally in a URI? If so, this might be a way forward, although I don’t see much benefit in the CURIE here myself. The alternative would be to change the field description to ‘URL of the germline set in the repository from which it can be downloaded’ and use the URL above as an example.

 

All the best

 

William

schristley:
> I think the overall motivation of this change is to provide a URI that will download the set, rather than this standardized form.

That's correct, though there is a bit more to this. We want germline sets to conform to the FAIR principles, and more importantly for the OGRDB service (and the data it provides like germline sets) to conform to the FAIR principles.

> As I understand a CURIE, it would provide a shorthand that avoids the need to write out the URL in full in the rest of the schema definition.

Exactly, it is primarily shorthand for the full URL, but it is important not to think of it purely as "URL to download the germline set". It is more than that. It is a permanent ID so that if you look at two studies, you can simply compare the IDs to know that they are using the exact same germline set, or not. It is also "F"indable and "R"eusable because I can download the exact same germline set based upon its ID. Some people tend to interpret the "R" as reproducibility.

> The thing that worries me is that it bakes OGRDB into the definition. We don’t intend OGRDB to be the only source of germline sets. It would be great if others, for example IMGT and MiLabs, started to support the MiAIRR standard for germline sets. However, if they do this, would we need to add CURIES to the schema definition for their systems?

Yes, if they want to conform to the FAIR principles, which they should. It is easy for us to add CURIEs for their systems. The "I"nteroperability of FAIR is where the standard schemas and formats become important. But interoperability doesn't imply that everybody has to agree and conform to the same schema, there are alternative ways to be interoperable.

> To do so would be much more effort on our part than we save by adding a URL shortcut. And I fear it might also discourage them, and send a general message that, for the MiAIRR standard, OGRDB is the single repository for reference sets. Is there a way to use a CURIE optionally in a URI? If so, this might be a way forward, although I don’t see much benefit in the CURIE here myself. The alternative would be to change the field description to ‘URL of the germline set in the repository from which it can be downloaded’ and use the URL above as an example.

I'm not sure why you think this would be more effort on our part. It's really effort on their part. The push toward FAIR-ness is pretty much unstoppable at this point, and if they want to be used and relevant then they need to conform to the FAIR principles. Note that this is not a blanket statement, thou must use the AIRR standard. And while "standard" can be used as a bludgeon to keep people in line, it can also just mean that a group agrees to the same set of general principles. While the field does not (technically) require that germline sets be in the AIRR standard format, because this is the AIRR Community and these are our AIRR Standards, there is that implication.

williamdlees:
ThatOn 3 Mar 2024, at 21:10, Scott Christley ***@***.***> wrote:﻿

I think the overall motivation of this change is to provide a URI that will download the set, rather than this standardized form.

That's correct, though there is a bit more to this. We want germline sets to conform to the FAIR principles, and more importantly for the OGRDB service (and the data it provides like germline sets) to conform to the FAIR principles.

As I understand a CURIE, it would provide a shorthand that avoids the need to write out the URL in full in the rest of the schema definition.

Exactly, it is primarily shorthand for the full URL, but it is important not to think of it purely as "URL to download the germline set". It is more than that. It is a permanent ID so that if you look at two studies, you can simply compare the IDs to know that they are using the exact same germline set, or not. It is also "F"indable and "R"eusable because I can download the exact same germline set based upon its ID. Some people tend to interpret the "R" as reproducibility.

The thing that worries me is that it bakes OGRDB into the definition. We don’t intend OGRDB to be the only source of germline sets. It would be great if others, for example IMGT and MiLabs, started to support the MiAIRR standard for germline sets. However, if they do this, would we need to add CURIES to the schema definition for their systems?

Yes, if they want to conform to the FAIR principles, which they should. It is easy for us to add CURIEs for their systems. The "I"nteroperability of FAIR is where the standard schemas and formats become important. But interoperability doesn't imply that everybody has to agree and conform to the same schema, there are alternative ways to be interoperable.

To do so would be much more effort on our part than we save by adding a URL shortcut. And I fear it might also discourage them, and send a general message that, for the MiAIRR standard, OGRDB is the single repository for reference sets. Is there a way to use a CURIE optionally in a URI? If so, this might be a way forward, although I don’t see much benefit in the CURIE here myself. The alternative would be to change the field description to ‘URL of the germline set in the repository from which it can be downloaded’ and use the URL above as an example.

I'm not sure why you think this would be more effort on our part. It's really effort on their part. The push toward FAIR-ness is pretty much unstoppable at this point, and if they want to be used and relevant then they need to conform to the FAIR principles. Note that this is not a blanket statement, thou must use the AIRR standard. And while "standard" can be used as a bludgeon to keep people in line, it can also just mean that a group agrees to the same set of general principles. While the field does not (technically) require that germline sets be in the AIRR standard format, because this is the AIRR Community and these are our AIRR Standards, there is that implication.

williamdlees:
Scott I take your point about the URI being a permanent identifier, but that does not to my mind have anything to do with CURIES. Do we have to use a CURIE to comply with FAIR? If not I would much rather not in this instance on the  grounds that 1 we save no appreciable effort by using a CURIE here and 2 it does take effort to check out the schema, make a change to it, coordinate with other changes and all the other stuff we end up doing when making a change to the standard.Thanks for your helpWilliamOn 3 Mar 2024, at 21:10, Scott Christley ***@***.***> wrote:﻿

I think the overall motivation of this change is to provide a URI that will download the set, rather than this standardized form.

That's correct, though there is a bit more to this. We want germline sets to conform to the FAIR principles, and more importantly for the OGRDB service (and the data it provides like germline sets) to conform to the FAIR principles.

As I understand a CURIE, it would provide a shorthand that avoids the need to write out the URL in full in the rest of the schema definition.

Exactly, it is primarily shorthand for the full URL, but it is important not to think of it purely as "URL to download the germline set". It is more than that. It is a permanent ID so that if you look at two studies, you can simply compare the IDs to know that they are using the exact same germline set, or not. It is also "F"indable and "R"eusable because I can download the exact same germline set based upon its ID. Some people tend to interpret the "R" as reproducibility.

The thing that worries me is that it bakes OGRDB into the definition. We don’t intend OGRDB to be the only source of germline sets. It would be great if others, for example IMGT and MiLabs, started to support the MiAIRR standard for germline sets. However, if they do this, would we need to add CURIES to the schema definition for their systems?

Yes, if they want to conform to the FAIR principles, which they should. It is easy for us to add CURIEs for their systems. The "I"nteroperability of FAIR is where the standard schemas and formats become important. But interoperability doesn't imply that everybody has to agree and conform to the same schema, there are alternative ways to be interoperable.

To do so would be much more effort on our part than we save by adding a URL shortcut. And I fear it might also discourage them, and send a general message that, for the MiAIRR standard, OGRDB is the single repository for reference sets. Is there a way to use a CURIE optionally in a URI? If so, this might be a way forward, although I don’t see much benefit in the CURIE here myself. The alternative would be to change the field description to ‘URL of the germline set in the repository from which it can be downloaded’ and use the URL above as an example.

I'm not sure why you think this would be more effort on our part. It's really effort on their part. The push toward FAIR-ness is pretty much unstoppable at this point, and if they want to be used and relevant then they need to conform to the FAIR principles. Note that this is not a blanket statement, thou must use the AIRR standard. And while "standard" can be used as a bludgeon to keep people in line, it can also just mean that a group agrees to the same set of general principles. While the field does not (technically) require that germline sets be in the AIRR standard format, because this is the AIRR Community and these are our AIRR Standards, there is that implication.

williamdlees:
I don't think so.

I am all for defining a specific syntax for the germline set reference.

From my point of view, there would be no problem making it a CURIE if 
we're ok with the right-hand-side being a URL, which as far as I can see 
is permitted by the CURIE syntax definition. You could say 'what's the 
point'?  but I don't think the specific case for making the reference a 
CURIE has been fully articulated.

I think we've probably gone through the arguments for and against other 
options, so I'll hold off repeating them!

All the best

William


------ Original Message ------
From "Jason Vander Heiden" ***@***.***>
To "airr-community/airr-standards" ***@***.***>
Cc "William Lees" ***@***.***>; "Mention" 
***@***.***>
Date 12/08/2024 19:30:05
Subject Re: [airr-community/airr-standards] Change germline_set_ref to 
be a CURIE (PR #770)

schristley:
Hi @williamdlees , Now with the [OGRDB API](https://ogrdb.airr-community.org/api_v2/swagger/#/) defined and live, I think we can resolve this properly. The path `/germline/set/{germline_set_id}/{release_version}` looks like the most specific way to reference a germline set. That's compatible with using a CURIE, e.g. `OGRDB:germline/set/9606.IGH_VDJ/9` where OGRDB is defined as `https://ogrdb.airr-community.org/api_v2/`. What do you think?

schristley:
Hi @bcorrie , in this current PR, there are two fields `germline_database` and `germline_set_ref` in data processing where `germline_database` has a text description of the germline set, while `germline_set_ref` is a CURIE that is resolvable (currently just OGRDB supports). Would the idea be to keep both of these fields in data processing? If yes, then I think that resolves William's issue above of having a way to describe the source when a CURIE isn't available.

schristley:
@bcorrie Should we only use the CURIE if it is resolvable and provides an AIRR-compatible GermlineSet JSON response? Or should we also use CURIE for IMGT as you [suggest](https://github.com/airr-community/airr-standards/pull/770#issuecomment-1972189461) above to point to a web file? The former gives us the possibility of doing automated stuff (in the AKC for example) because we are guaranteed of the data format, in the latter case we would have to know in which cases we can do that (OGRDB) and which cases we cannot (IMGT, and maybe others)

bussec:
> we should also use "CURIE" for IMGT and other germline databases to have consistent reference IDs for them.

@schristley What does this mean exactly? Should there be

1. only one (`GERMLINEDB`) prefix or
2. multiple prefixes (`ORGDB`, `IMGT`, ...) but with a consistent structure for the local part (e.g., always following a `/{species}/{set}/{release}` schema)  or
3. multiple prefixes and resource-specific local parts, but at least CURIEs (i.e. resolveable URIs) for all germline resources?

schristley:
> > we should also use "CURIE" for IMGT and other germline databases to have consistent reference IDs for them.
> 
> @schristley What does this mean exactly? Should there be
> 

It's a syntactic difference when I say "CURIE". My understanding is that the strict definition of a CURIE is meant to be a simple prefix, that's just concatenated to the beginning, so I was thinking that some usages might not exactly match the definition (if you have to construct the URI using variable replacement). I don't think this is a concern anymore, as you [found](https://github.com/airr-community/airr-standards/issues/553#issuecomment-2863311126) that slashes can be used in the local part of the CURIE, so that means the OGRDB CURIE should be fine. It also looks like the IMGT files can be defined as CURIEs as well.

Based upon that, I would recommend 3) as that provides the greatest flexibility for a resolvable URI.

> 3. multiple prefixes and resource-specific local parts, but at least CURIEs (i.e. resolveable URIs) for all germline resources?

schristley (approved):

schristley:
@bcorrie @williamdlees This should be ready to merge.

PR #748, opened 2024-02-11
RepertoireGroup development for v2.0

bcorrie:
In the descriptions of each field, it uses the wording "repertoire collection" not "repertoire group". Should this object be called `RepertoireCollection`

closes #578

PR #773, opened 2024-03-04
RepertoireGroup updates for v2.0, take two

### LINKED ISSUE #578: RepertoireGroup refinements
A couple of questions that came up in an iReceptor Plus meeting today:

- From an ADC perspective, repertoire_ids are typically unique within a given context (in the ADC they are unique within a repository). It seems useful to add a `repertoire_group_url` to the fields of RepertoireGroup to uniquely identify the source of the repertoire and where it was found. Or do we consider this implicit in the sense that one would have a separate RepertoireGroup for each repository in an analysis. This implicit definition is challenging if you want to have a RepertoireGroup that is grouping repertoires from multilple repositories.
- I am not sure only having `repertoire_id` is sufficient to completely describe the groups of repertoires that we might want to group together. For example, if we wanted a RepertoireGroup that only contained IGH and I have a different SampeProcessing or DataProcessing for these rearrangements, I would need to provide a `sample_processsing_id` or a `data_processing_id` to accurately describe the set of rearrangements that are being considered as part of this RepertoireGroup.
- Do we really want a TimePoint object in the RepertoireGroup? This seems to imply that RepertoireGroups are targeted at data over multiple time points. Although that might be a common use case, it is certainly not the only one that make sense...

scharch:
Replaces #748 
Closes #578 

So far: just some docs updates...


scharch:
I went very simple with the filters and just linked the docs page to what you have written for the API. If it looks ok, I will copy to the other files.

schristley:
@scharch Are the filters only suppose to apply to rearrangements? Or are you expecting that it can be specified using a typed-key like in the [example](https://github.com/airr-community/airr-standards/issues/578#issuecomment-1975265885)?

scharch:
> @scharch Are the filters only suppose to apply to rearrangements? Or are you expecting that it can be specified using a typed-key like in the [example](https://github.com/airr-community/airr-standards/issues/578#issuecomment-1975265885)?

The latter. Does that need to be in the schema explicitly?

schristley:
> > @scharch Are the filters only suppose to apply to rearrangements? Or are you expecting that it can be specified using a typed-key like in the [example](https://github.com/airr-community/airr-standards/issues/578#issuecomment-1975265885)?
> 
> The latter. Does that need to be in the schema explicitly?

It doesn't need to be but I think it would make it more clear, as it's not obvious that should be done based upon the doc blurb. It would also make it more explicit what AIRR objects can be filtered. Thinking about it, using `Repertoire:` seems reasonable at the group level, but then that conflicts with the `repertoires` array...

javh:
From the call:
* This might be stale.
* Need to review.

bcorrie:
Here is an example of what we should be able to do - I will revisit.

https://github.com/airr-community/airr-standards/issues/578#issuecomment-1962158840

schristley:
@scharch Any other changes to `RepertoireGroup`?

PR #776, opened 2024-03-06
Restructuring of Cell

### LINKED ISSUE #477: Representing cell_phenotype using RNA-seq marker gene expression

### LINKED ISSUE #768: Re-visit implementation decisions around `Cell` object
A discussion around the necessity and the architecture of the `Cell` object and its associated object recently [arose in #705](https://github.com/airr-community/airr-standards/pull/705#issuecomment-1964644855). Indeed the current relations and links between `Cell`, `CellExpression`, `CellReactivity` and other objects (incl. `Receptor` and `Rearrangement`) appear overly complicated. Various solutions have been proposed, including:

1. `Cell` should become a container for other `Cell*` objects, which would then not exists as independent top-level objects anymore
2. `Cell` should be completely removed and cells would only be represented in a data set via a `cell_id` in various other objects, but not as an object on its own.

javh:
Initial changes to `Cell`, `CellExpression` and `CellReactivity` per March 4, 2024 call:
* Remove `Rearrangement` and expression links in `Cell`.
* Renamed `CellExpression` to `Expression`.
* Renamed `CellReactivity` to `Reactivity`.

Also:
* Added `cell_subset`, `cell_phenotype`, and `cell_label` for CL term, marker definition, and free text annotation, respectively. 
* Not sure I like these names, but I stuck with what was already in the `CellProcessing` object.
* Not sure we need `cell_phenotype`, but I thought I'd include it for discussion purposes.

I only modified the main v2 spec. I will propagate once we agree on the changes.

Closes #768
Closes #477

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -2516,7 +2516,8 @@ CellProcessing:
                         label: lymphocyte
         cell_phenotype:
             type: string
-            description: List of cellular markers and their expression levels used to isolate the cell population
+            description: >
+                List of cellular markers and their expression levels used to isolate the cell population.
             title: Cell subset phenotype
             example: CD19+ CD38+ CD27+ IgM- IgD-
             x-airr:
@@ -4587,7 +4588,8 @@ Cell:
                         label: lymphocyte
         cell_phenotype:
             type: string
-            description: List of cellular markers and their expression levels used to isolate the cell population.
+            description: >
+                List of cellular markers and their expression levels used to isolate the cell population.
             title: Cell subset phenotype
             example: CD19+ CD38+ CD27+ IgM- IgD-
             x-airr:
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -2526,6 +2526,20 @@ CellProcessing:
                 set: 3
                 subset: process (cell)
                 name: Cell subset phenotype
+        cell_label:
+            type: string
+            description: >
+                Free text cell type annotation. Primarily used for annotating cell types that are not 
+                provided in the Cell Ontology.
+            title: Cell annotation
+            example: age-associated B cell
+            x-airr:
+                miairr: defined
+                nullable: true
+                adc-query-support: true
+                set: 3
+                subset: process (cell)
+                name: Cell annotation
         cell_species:
             $ref: '#/Ontology'
             description: >
@@ -4506,7 +4520,6 @@ Cell:
     type: object
     required:
         - cell_id
-        - rearrangements
         - repertoire_id
         - virtual_pairing
     properties:
@@ -4523,32 +4536,6 @@ Cell:
                 nullable: false
                 adc-query-support: true
                 name: Cell index
-        rearrangements:
-            type: array
-            description: >
-                Array of sequence identifiers defined for the Rearrangement objects associated with this cell
-            title: Cell-associated rearrangements
-            items:
-                type: string
-            example: [id1, id2]
-            x-airr:
-                miairr: defined
-                nullable: true
-                adc-query-support: true
-                name: Cell-associated rearrangements
-        receptors:
-            type: array
-            description: >
-                Array of receptor identifiers defined for the Receptor objects associated with this cell
-            title: Cell-associated receptors
-            items:
-                type: string
-            example: [id1, id2]
-            x-airr:
-                miairr: defined
-                nullable: true
-                adc-query-support: true
-                name: Cell-associated receptors
         repertoire_id:
             type: string
             description: Identifier to the associated repertoire in study metadata.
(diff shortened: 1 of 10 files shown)


scharch:
Looks fine. Will need docs updates, at minimum for the name of `Expression` and `Reactivity`. But I think we should also add explicit guidance that says either "put RNAseq data in `Expression`" or explain how external data (eg GEO) can be linked at the `Sample` level.

bcorrie:
Yes, looks good to me as well... Will hold off on review check until final is ready, but I think safe to sync the specs.

bussec (approved):

bussec:
Looks good! Wondering whether we should also introduce a `cell_label` property into the `CellProcessing` object...

javh:
From the call:
* Add `cell_label` to `CellProcessing`.
* Set `Cell.cell_subset` to `miairr: important`. Verify this is clear in the documentation.
* Merge.

bcorrie:
Hate to say it, but I am trying to use the spec as it is now, and I see an issue...

Many of the Receptors in IEDB do not have the full variable domains for one chain, let along multiple chains. e.g. https://www.iedb.org/receptor/184739 has V, J, CDR3 for alpha and beta chains, but there is no full V domain.

The AIRR Spec has the V Domain for both chains as required and non-nullable. Is that really what we want. We would have no way of capturing IEDB:184739  in the AIRR Receptor schema if that was the case???

bcorrie:
I am OK to make the above a separate issue so we can merge the current PR.

bcorrie:
Asking the question another way. Lets say I analyze a single cell study and I find that some cells match (at some level of matching) a known Receptor in IEDB. Lets say the CDR3 and VJ genes are the same. So I create a `Receptor` object and link that `Cell` to that Receptor by adding the `receptor_id` to the `Cell.receptors` array. 

But I am only allowed to do this if the IEDB entry is a paired chain with full V Domain for both chains. If you look at IEDB and search for paired chain receptors you are limited to 31,346 out of the 189,435 "Receptor like things" in IEDB.

This seems like a significant restriction.

scharch:
> Many of the Receptors in IEDB do not have the full variable domains for one chain, let along multiple chains. e.g. https://www.iedb.org/receptor/184739 has V, J, CDR3 for alpha and beta chains, but there is no full V domain.

It's a TCR, so the full V domain is implied. I think it's ok.

bcorrie:
> > Many of the Receptors in IEDB do not have the full variable domains for one chain, let along multiple chains. e.g. https://www.iedb.org/receptor/184739 has V, J, CDR3 for alpha and beta chains, but there is no full V domain.
> 
> It's a TCR, so the full V domain is implied. I think it's ok.

Problem is that the `receptor_variable_domain_1_aa` field is required (as is the domain 2 field), non-nullable, and says: 
```
        receptor_variable_domain_1_aa:
            type: string
            description: >
                Complete amino acid sequence of the mature variable domain of the Ig heavy, TCR beta or TCR delta chain.
                The mature variable domain is defined as encompassing all AA from and including first AA after the the
                signal peptide to and including the last AA that is completely encoded by the J gene.
            example: >
              QVQLQQPGAELVKPGASVKLSCKASGYTFTSYWMHWVKQRPGRGLEWIGRIDPNSGGTKYNEKFKSKATLTVDKPSSTAYMQLSSLTSEDSAVYYCARYDYYGSSYFDYWGQGTTLTVSS
            x-airr:
                nullable: false
                adc-query-support: true
```
So because we don't have the AA variable domains, you can't create an AIRR standards compliant `Receptor` object for https://www.iedb.org/receptor/184739

It seems to me that our requirements are to stringent.

bussec:
The stringency in the schema is deliberate, as the purpose of `Receptor` was always to be complementary to the potentially incomplete receptor information in IEDB. People are free to download the IEDB database and all `Receptor` objects from the ADC and use whatever algorithm and arbitrary threshold they want to make guesses whether they are sufficiently similar. But I don't think that we need to degrade our schema for that.

bcorrie:
I understand that use case, but if that is indeed the case, then we do not have a mechanism to link an AIRR entity to a known entity in another repository (e.g. IEDB) unless it meets those strict criteria.

There are 150,000+ such entities in IEDB that contain Receptor -> antigen/epitope reactivity information that we are unable to link to given the above constraints. My understanding of the AIRR `Receptor` object was that it would help us make such a link. You aren't telling me I have been wrong all this time are you 8-)

bcorrie:
> People are free to download the IEDB database and all `Receptor` objects from the ADC and use whatever algorithm and arbitrary threshold they want to make guesses whether they are sufficiently similar. But I don't think that we need to degrade our schema for that.

But shouldn't our schema have a mechanism to capture when such an algorithm is used to annotate an AIRR entity (`Cell` or `Rearrangement`) with such specificity/reactivity information.

bcorrie:
Suggest we move this discussion to #781 and merge this PR.

bcorrie (approved):

javh:
There's currently no `Cell` documentation to verify w.r.t. CL importance, so we'll come back to it.

PR #778, opened 2024-03-21
Clone-schema-updates

scharch:


scharch:
From the call:
* push `members` and `inferred_members` down to a single revived but skinny `Node` array
* `Node` should have both `sequence_id` and `cell_id` with only one populated (but maybe add a `level` shortcut too?)
* Change `Cell.inferred` and `Rearrangement.inferred` to an enum to match `Repertoire.type` and add this to `Node` as well
  * Does an inferred `Node` need to be part of a `Repertoire` at all? Maybe this allows us to drop that bit entirely...
  * preference was for 'observed' or 'physical' in the current enum
* Add `repertoire_id` to `Node` to facilitate retrieval of underlying data
* Did not discuss what to do about `data_processing_id`

javh:
Random thoughts:
* For `repertoire_type` what about `observed` instead of `physical`?
* Maybe the `inferred` boolean for Rearrangement could instead be `sequence_type` and use the same enum as `repertoire_type`? Same for `Cell` and a `cell_type` field.
* As an alternative to `level`/`members` in `Clone`, we could have `sequences` / `cells` which are explicitly typed and set the expectation to only populate one (we could still retain `level` in this case).
* Another alternative to `level`/`members` in `Clone` would be to have a different version of `Node` that contains the `sequence_id`/`cell_id` fields (without the junction and whatnot), and possibly the `level`/`sequence_type`/`cell_type`, and possibly the `repertoire_id` so that doesn't need to be fetched out of `RepertoireGroup` somehow. Ie, `members` contains `Node` references.

schristley:
> * [ ]  What do we do about `data_processing_id`? Can `RepertoireGroup`s get their own `DataProcessing` objects?

@scharch It's best to think of `DataProcessing` as an independent, orthogonal object to the other (top-level) AIRR objects. Anything that can be part of `DataProcessing` should have a `data_processing_id` so yes including one in `RepertoireGroup` makes sense.

scharch:
OK, let me know how this looks.
* I changed "physical" to "observed" and made `rearrangement_type` and `cell_type` enums instead of `inferred` booleans
* I re-added `Node` as a way get `repertoire_id` in. It feels a little "overweight" but probably the best solution?

scharch:
I updated the docs, but wasn't sure about the "file format" bit - I mostly copied from `Repertoire` but I think that's probably out of date, because it doesn't mention `DataFile`. Also, do we need any additional documentation around the new `repertoire_type`, `rearrangement_type`, `cell_type` enums, or are the field descriptions sufficient?

Before I sync schemas, a couple of other questions:
1. Do `Clone` and `Node` both need to reference the `repertoire_group_id`? I think maybe we could take it out of `Node`, especially since that also has the specific `repertoire_id` already...
2. `Clone` is currently defined only on a `RepertoireGroup`. But a single `Repertoire` is probably still a pretty normal use case. Do we want to make people define a `RepertoireGroup` in that case? I lean toward yes, because it keeps things clean and consistent, but I get how that could be annoying.
3. Does anyone remember if having `Clone.inferred_ancestor` be just an _id was intentional? Seems like it should be a `Node`.
4. I think we've discussed before (long, long ago) and are happy with allowing only a single tree per `Clone` but figured I should double check.
5. Do we have a way of enforcing (well, checking) that `Node.sequence_id` and `Node.cell_id` are mutually exclusive? Or are we just documenting that and trusting users?
6. Should we allow `Node.node_type` to include 'simulated' on the theory that you can build trees of simulated `Rearrangement`s?

scharch:
From the call:
-Ok to remove `repertoire_group_id` from `Node`
-Add `repertoire_id` to `Clone` and specify alternates in documentation
-yes inferred_ancestor should be a `Node`
-one tree string
-documentation only for mutually exclusive (see also `repertoire_group_id` vs `repertoire_id` in `Clone` above)
-add simulated to enum
-add documentation about expected consistency between `Repertoire.repertoire_type` and `Rearrangement.rearrangement_type`

javh:
`nullable` is used in the "openapi3" specs. `xairr: nullable` is used on the other spec files (OpenAPI v2), because v2 doesn't have a built-in `nullable` attibute (that's the biggest difference between v2 and v3 for us).

scharch:
Stuck again :(
schema consistency test says:
```
Clone object is different between V2 and V3 spec
{'properties': {'clone_id': {delete: ['nullable']}}}
```
but they look the same to me. do the versions have different defaults or something?
https://github.com/airr-community/airr-standards/blob/d9c61689d629368956df2ef77ee1389b3fb33e2f/specs/airr-schema-openapi3.yaml#L4844-L4850
https://github.com/airr-community/airr-standards/blob/d9c61689d629368956df2ef77ee1389b3fb33e2f/specs/airr-schema.yaml#L4569-L4575

javh:
I don't think so... Those look the same. Let me fiddle with it for a min.

javh:
Okay, schema consistent test is fixed. It was a little confusing because the error message wasn't super clear. The problem was with the v2 vs v3 specs in `/lang/python` (not `/specs`), so I tweaked the error message.

Then the same error with `node_id` after getting past `clone_id`, so I fixed that. Pull?

Node.js ones are failing, but I've never looked at those...

scharch:
The last thing from the original checklist is unit tests - just good/bad examples in tests/data?

javh:
> just good/bad examples in tests/data?

We'll also need to add the examples to the lang packages and write the appropriate tests to check the test data, but I think adding good/bad examples to `/tests/data` is enough to close this. Maybe we can open a separate PR for the R/python test writing?

scharch:
No worries @schristley that was the conclusion from the last call, see:
https://github.com/airr-community/airr-standards/blob/5ed2f7fe7e10ed995cba29e321feac6e2a4e8988/specs/airr-schema-openapi3.yaml#L4851-L4866

and 

https://github.com/airr-community/airr-standards/blob/5ed2f7fe7e10ed995cba29e321feac6e2a4e8988/docs/datarep/clone.rst?plain=1#L10-L12

scharch:
> > just good/bad examples in tests/data?
> 
> We'll also need to add the examples to the lang packages and write the appropriate tests to check the test data, but I think adding good/bad examples to `/tests/data` is enough to close this. Maybe we can open a separate PR for the R/python test writing?

OK, hopefully I formatted the yaml correctly...

PR #799, opened 2024-07-26
minimums for physical quantities

schristley:
JSON schema supports defining minimum and maximum values for numbers and integers. As these fields are physical quantities, they should always be positive. Adding these provide for better validation.

I only did for fields in `Repertoire`. We can consider adding for other fields if it makes sense.

This is for V1. Need to think how to do for V2 as we've restructured the fields.

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -1808,6 +1808,7 @@ Subject:
                 format: controlled_vocabulary
         age_min:
             type: number
+            minimum: 0
             description: Specific age or lower boundary of age range.
             title: Age minimum
             example: 60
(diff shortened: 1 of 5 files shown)


bussec (approved):

javh:
@schristley Can you please clarify what you are referring to here:
> This is for V1. Need to think how to do for V2 as we've restructured the fields.

From the call:
* Do we need to add validation of min/max values to python/R reference libraries?

javh (approved):

schristley:
> @schristley Can you please clarify what you are referring to here:
> 
> > This is for V1. Need to think how to do for V2 as we've restructured the fields.

For V2, we have nested objects like PhysicalQuantity that's replacing the V1 fields. We need to make sure that the minimum is applicable to all.

> 
> From the call:
> 
> * Do we need to add validation of min/max values to python/R reference libraries?

Yes we should, though if we move to a standard validation library, at least for python, we get that check automatically.

PR #800, opened 2024-08-12
Fix ability to capture healthy controls

### LINKED ISSUE #164: Question around study_group_description
Again, from Emily, sparked by our discussion at the Vocab/Ontology meeting...

We have been treating it solely as case and control, but never actively defined "case" - generally the studies we work this have a single case that can be described in diagnosis, etc. Essentially, should we change this and define case and simply refer to a control as a control? Or should we alter it so that we define the two more effectively. Eg. study on SLE, case is patients with flares, control is patients without flares. Would we want this defined specifically in study_group_description?

Maybe this is a discussion for the Vocab/Ontology diagnosis group?

### LINKED ISSUE #516: How do I find Healthy Controls using AIRR fields?
Since Healthy isn't a disease, there is no DOID entry for it. So you can't say `diagnosis.disease_diagnosis = Healthy`.

It isn't safe to assume that `diagnosis.disease_diagnosis` == null means that the subject is healthy at this time point, because the  data might simply be missing.

iReceptor currently uses an internal controlled vocabulary for study.study_description to capture which study arm a subject is in. Obviously not ideal as this is not consistent. I think this needs to be sorted out, as this is a pretty important distinction to be able to make and we have had several users ask about how to determine whether a given repertoire from a given individual is a healthy control data set or not... Maybe as part of the OntoVoc sprint?

bcorrie:
Closes #164 
Closes #516

### DIFF
--- a/specs/airr-schema.yaml
+++ b/specs/airr-schema.yaml
@@ -2279,6 +2279,18 @@ Subject:
                 set: 1
                 subset: subject
                 name: Relation type
+        study_group_description:
+            type: string
+            description: Designation of study arm to which the subject is assigned to
+            title: Study group description
+            example: control
+            x-airr:
+                miairr: important
+                nullable: true
+                adc-query-support: true
+                set: 1
+                subset: Subject
+                name: Study group description
         diagnosis:
             type: array
             description: Diagnosis information for subject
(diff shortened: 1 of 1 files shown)


javh:
Maybe we skip this for 2.0 as a solution seems somewhat difficult.  The main challenge is that the Disease Ontology doesn't allow mapping to healthy/normal. Is this an alternative here?

Issue #813, opened 2024-11-12
bugs in airr-schema.yaml / airr-schema-openapi3.yaml


LonnekeScheffer:
Some issues encountered when auto-generating LinkML from airr-schema.yaml / airr-schema-openapi3.yaml
These are all relatively easy/small fixes.

- [ ] Property `unit` occurs as an ontology across different classes. This ontology does not have the same meaning. E.g.: `TimePoint` (unit=UO:0000003), `TimeInterval` (unit=UO:0000033), `PhysicalQuantity` (unit=UO:0000024), `TimeQuantity` (unit=UO:0000033). The field 'unit' should be renamed to something unique (time_unit, time_interval_unit, physical_quantity_unit)
- [ ] Property `release_version` has inconsistent type. In `AlleleDescription`, type=integer; in `GermlineSet`, type=number. Should be integer in both cases. 
- [ ] Ontology `species` is missing source node in some places. This ontology is correctly defined in `Subject`, but in `AlleleDescription` and `GermlineSet`, the expected field `x-airr/ontology/top_node/id` is missing. 
- [ ] Some enums are defined several times, and differ in that they contain 'null' in some places, but not in others: `locus` (contains 'null' in `Rearrangement`, but not in `AlleleDescription`, `GermlineSet`, `Genotype`); `mhc_class` (contains 'null' in `Reactivity`, not in `MHCGenotype`)
- [ ] For ontology `property` in class `Expression` the expected field `x-airr/ontology/top_node/id` is missing. 
- [ ] For ontology `antigen` in class `Reactivity` the expected field `x-airr/ontology/top_node/id` is missing. 
- [ ] Property `reactivity_ref` is defined with two different types. Inside class `Reactivity`, the type is array. Inside class `Rearrangement`, the type is string, but the description explicitly states that this string should be a comma separated list. I would propose changing the type to array. (and besides, `Rearrangement` also contains `reactivity_id` which is a comma separated list as string, the type for that should then probably change as well.)
- [ ] Property `nodes` of class `Tree` should have `$ref: '#/Node'` directly under `nodes`, now it is placed under a field called `additionalProperties` which is not used anywhere else

schristley:
[...]

Issue #733, opened 2023-11-13
Change CellExpression proprty_type to enum?


bcorrie:
In #699 we added the property_type field to CellExpression as a string with documented preferred values.

We noted that we probably should change this to an enum.

bcorrie:
Currently we have this:
```
               Keyword describing the property type and detection method used to measure the property value.
                The following keywords are recommended, but custom property types are also valid:
                "mrna_expression_by_read_count",
                "protein_expression_by_fluorescence_intensity", "antigen_bait_binding_by_fluorescence_intensity",
                "protein_expression_by_dna_barcode_count" and "antigen_bait_binding_by_dna_barcode_count".
```

bcorrie:
Can we come up with a set of enum values for this? Outside of my area of expertise... 8-)

bcorrie:
@bussec @javh any objections to leaving this as a string with "suggested values". If no objections I will close this issue.

bussec:
@schristley I would also prefer to have an ontology, but then we would have to split the field again, as I doubt that there is an ontology that would capture both components of our keywords (i.e., what was measured and how it was measured) and we decided a while ago, that having two components is potentially confusing to curators.

bussec:
For the call: Leave the field as it is, i.e. free text with suggested keywords. This is clearly not perfect, but better than having a solution that is more formalized but less flexible.

PR #814, opened 2024-11-26
AKC updates for v1.6

schristley:
Consistency issues with the schema file as reported by the AKC project.

### DIFF
--- a/lang/R/inst/extdata/airr-schema.yaml
+++ b/lang/R/inst/extdata/airr-schema.yaml
@@ -4439,19 +4439,17 @@ CellExpression:
                 adc-query-support: true
                 name: Property type and detection method
         property:
-            $ref: '#/Ontology'
+            type: string
             title: Property information
             description: >
                 Name of the property observed, typically a gene or antibody identifier (and label) from a 
                 canonical resource such as Ensembl (e.g. ENSG00000275747, IGHV3-79) or 
                 Antibody Registry (ABREG:1236456, Purified anti-mouse/rat/human CD27 antibody).
-            example:
-                id: ENSG:ENSG00000275747
-                label: IGHV3-79
+            example: ENSG:ENSG00000275747
             x-airr:
                 miairr: defined
                 adc-query-support: true
-                format: ontology
+                format: CURIE
                 name: Property information
         property_value:
             type: number
@@ -4606,19 +4604,17 @@ ReceptorReactivity:
             x-airr:
                 nullable: false
         antigen:
-            $ref: '#/Ontology'
+            type: string
             description: >
                 The substance against which the receptor was tested. This can be any substance that
                 stimulates an adaptive immune response in the host, either through antibody production
                 or by T cell activation after presentation via an MHC molecule.
             title: Antigen
-            example:
-                id: UNIPROT:P19597
-                label: Circumsporozoite protein
+            example: UNIPROT:P19597
             x-airr:
                 nullable: false
                 adc-query-support: true
-                format: ontology
+                format: CURIE
         antigen_source_species:
             $ref: '#/Ontology'
             description: The species from which the antigen was isolated
--- a/lang/js/airr-schema-openapi3.yaml
+++ b/lang/js/airr-schema-openapi3.yaml
@@ -4732,20 +4732,18 @@ CellExpression:
                 adc-query-support: true
                 name: Property type and detection method
         property:
-            $ref: '#/Ontology'
+            type: string
             nullable: true
             title: Property information
             description: >
                 Name of the property observed, typically a gene or antibody identifier (and label) from a 
                 canonical resource such as Ensembl (e.g. ENSG00000275747, IGHV3-79) or 
                 Antibody Registry (ABREG:1236456, Purified anti-mouse/rat/human CD27 antibody).
-            example:
-                id: ENSG:ENSG00000275747
-                label: IGHV3-79
+            example: ENSG:ENSG00000275747
             x-airr:
                 miairr: defined
                 adc-query-support: true
-                format: ontology
+                format: CURIE
                 name: Property information
         property_value:
             type: number
@@ -4897,19 +4895,17 @@ ReceptorReactivity:
                 The type of antigen before processing by the immune system.
             example: protein
         antigen:
-            $ref: '#/Ontology'
+            type: string
             nullable: false
             description: >
                 The substance against which the receptor was tested. This can be any substance that
                 stimulates an adaptive immune response in the host, either through antibody production
                 or by T cell activation after presentation via an MHC molecule.
             title: Antigen
-            example:
-                id: UNIPROT:P19597
-                label: Circumsporozoite protein
+            example: UNIPROT:P19597
             x-airr:
                 adc-query-support: true
-                format: ontology
+                format: CURIE
         antigen_source_species:
             $ref: '#/Ontology'
             nullable: true
--- a/lang/python/airr/specs/airr-schema.yaml
+++ b/lang/python/airr/specs/airr-schema.yaml
@@ -4439,19 +4439,17 @@ CellExpression:
                 adc-query-support: true
                 name: Property type and detection method
         property:
-            $ref: '#/Ontology'
+            type: string
             title: Property information
             description: >
                 Name of the property observed, typically a gene or antibody identifier (and label) from a 
                 canonical resource such as Ensembl (e.g. ENSG00000275747, IGHV3-79) or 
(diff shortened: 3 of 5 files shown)


javh:
From the call:
* Do we need do a v1.5 release if we're going to release v2.0?
* Are the changes to TimePoint and RepertoireGroup backwards incompatibility with v1.5? IIRC, the TimePoint changes weren't intended for v1.5 (I could be wrong).

schristley:
> From the call:
> 
> * Do we need do a v1.5 release if we're going to release v2.0?

Yes, the ADC is still using the v1.x release and probably will for a while yet, even when v2.0 is released.

> * Are the changes to TimePoint and RepertoireGroup backwards incompatibility with v1.5? IIRC, the TimePoint changes weren't intended for v1.5 (I could be wrong).

yeah I'm pretty sure some of the changes are backwards incompatible, which is also why need v1.x for awhile as it requires converting data and code for the v2 schema.

javh:
Got it. Doesn't that mean we need to back out those TimePoint and RepertoireGroup changes from this PR, as they aren't compatible with v1.5? (But leave the RepertoireGroup changes in the v2.0 PR.)

bcorrie:
Yes, I was going to mention that. If there are changes that are not backwards compatible, then it is a 1.6 release, not 1.5

bcorrie:
> > From the call:
> > 
> > * Do we need do a v1.5 release if we're going to release v2.0?
> 
> Yes, the ADC is still using the v1.x release and probably will for a while yet, even when v2.0 is released.

@schristley why do you feel these fixes are needed in 1.5. None of them are critical (they aren't bug fixes per se), and some of them are breaking backwards compatibility to change the name of a field (e.g. ` unit` -> `time_point_unit`) or change the type in a way that isn't particularly critical (`number` => `integer`). Both break backwards compatibility, and require repository and/or API implementation changes. This would also make 1.5 "more" incompatible with 1.4, which would make things worse, not better, at least in my opinion. If I am not mistaken most of the ADC repositories are v1.4 not v1.5 - I know all of ours are running v1.4.

The `RepertoireGroupDetail` perhaps provides better structure, but doesn't change anything other than the JSON/YAML. So this isn't really necessary is it as it isn't fixing a bug.

Why can't these wait for 2.0?

The one exception I could see is the Taxonomy additions, as those changes should have been there in the first place.

bcorrie:
> If I am not mistaken most of the ADC repositories are v1.4 not v1.5 - I know all of ours are running v1.4.

Indeed, the v1.2 ADC API spec, which is the latest is coupled to the v1.4.1 specification. We did not release a version of the ADC API spec that supported v1.5. It is confusing I admit, as the API version numbering and AIRR Standard version numbering are too tightly coupled, but we never resolved that. 

So from an AKC perspective, I think we should be looking at either extracting data from repositories using the ADC v1.2/AIRR v1.4 specs or the ADC v2/AIRR v2 specs.

schristley:
> > > From the call:
> > > 
> > > * Do we need do a v1.5 release if we're going to release v2.0?
> > 
> > 
> > Yes, the ADC is still using the v1.x release and probably will for a while yet, even when v2.0 is released.
> 
> @schristley why do you feel these fixes are needed in 1.5.

I'm being loose with versioning, if the schema isn't backward compatible then yes it's 1.6. Call them bugs, call it schema cleaning or harmonization, but these are issues brought up in #813 for AKC integration. I don't think the ADC has to move to v1.6, might just move to v2.0 if that's ready.

bcorrie:
> I'm being loose with versioning, if the schema isn't backward compatible then yes it's 1.6. Call them bugs, call it schema cleaning or harmonization, but these are issues brought up in #813 for AKC integration. I don't think the ADC has to move to v1.6, might just move to v2.0 if that's ready.

"Ready" is an interesting term, but I think the AKC would be best served by having fixes we have identified now integrated into v2.0, and not make any changes to the 1.x versions. There might be some minor fixes we could do, but anything that breaks backwards compatibility I would suggest is not a good way to go.

bcorrie:
So can we close this issue and focus on the v2.0 pull request #815

schristley:
> > I'm being loose with versioning, if the schema isn't backward compatible then yes it's 1.6. Call them bugs, call it schema cleaning or harmonization, but these are issues brought up in #813 for AKC integration. I don't think the ADC has to move to v1.6, might just move to v2.0 if that's ready.
> 
> "Ready" is an interesting term, but I think the AKC would be best served by having fixes we have identified now integrated into v2.0, and not make any changes to the 1.x versions. There might be some minor fixes we could do, but anything that breaks backwards compatibility I would suggest is not a good way to go.

Fixes are going into v2, but the ADC and the AKC are not converting to V2 yet, and I don't think we even have a timeline for that. So for now, we need to work with 1.x

bcorrie:
> So for now, we need to work with 1.x

Yes, but I am suggesting that the AKC work with the existing ADC v1.2/AIRR v1.4 and any changes driven by the AKC would go into v2.0.

Currently VDJServer is running ADC API v1.2, just like the iReceptor repositories. Are you intending to upgrade VDJServer to AIRR 1.5.x or AIRR v1.6?

I am not intending on upgrading any of the iReceptor repositories beyond the current ADC API v1.2/AIRR v1.4 until we upgrade to v2.0. So even if there is a new version of the AIRR Standard, the data that is pulled out of the ADC is unlikely to have these new features that you are suggesting unless the repositories maintainers upgrade. The v2.0 upgrade is going to be hugely time consuming by itself...

schristley:
> Currently VDJServer is running ADC API v1.2, just like the iReceptor repositories. Are you intending to upgrade VDJServer to AIRR 1.5.x or AIRR v1.6?

From a source code perspective, VDJServer is already using AIRR 1.5 release-branch for the airr-js code that I added. Will I convert the data to v1.6 before going to v2.0. Probably not. I don't know, it might depend upon the timing, but I don't think it is necessary. At the same time, I don't think the changes are that major either. I have to review all of the changes completely before I make any decision there.

The AKC is using AIRR 1.x in a couple different ways. One is your conversion script, but that's tied to the version of the data itself and essentially independent from the AIRR standards version. The other is the schema conversion, and that's what would be affected, but right now we don't have anything using that linkml schema

bcorrie:
> The AKC is using AIRR 1.x in a couple different ways. One is your conversion script, but that's tied to the version of the data itself and essentially independent from the AIRR standards version.

The converter works on the data that comes out of the ADC, but it does need to know which version of the standard the data is being produced with. It would need a separate configuration file to do the conversion correctly if the ADC API version was different.

schristley:
> Got it. Doesn't that mean we need to back out those TimePoint and RepertoireGroup changes from this PR, as they aren't compatible with v1.5? (But leave the RepertoireGroup changes in the v2.0 PR.)

@javh The RepertoireGroup changes are purely structural, to follow our practice with array of objects, similar to what we did with the GermlineSet restructure a while back, and it should be backward compatible. It's the TimePoint changes that break, but there we are saved by the fact that no ADC data uses TimePoint so it's a moot point. v1.6 release will be the first step for better AKC integration. AKC cannot use v2.0 immediately because ADC data isn't available in v2.0 format, but we have the schema [translated](https://github.com/airr-knowledge/ak-schema/blob/main/src/ak_schema/schema/ak_airr_v2.0.yaml) so we will be ready.

javh:
@schristley, I see. We were hoping to entirely avoid any more releases before v2.0 and just have people switch to v2.0 en masse. But, if this is a huge problem, maybe we can revisit that decision?  If we do a v1.6, can we skip the software releases and make it another "schema only" release? Can the ADC switch to v2.0?

scharch:
@schristley Can we move the `RepertoireGroupDetail` fix to #733? It seems to fit better there, but don't know if doing that would mess with your desired AKC timeline.

...Also, can we come up with a better name? Especially if we move the filters from #733 into this new object, it feels more like a `RepertoireProcessing` (??) or something like that... Thoughts?

schristley:
@scharch I think you mean #773 ? Yes I can move `RepertoireGroupDetail` there. I wasn't sure of good name to use. How about `RepertoireList` or `RepertoireFilter`?

scharch:
@schristley whoops yeah, sorry
I think `RepertoireFilter` is better than "List" because the whole point is to reference a ...list of these from `RepertoireGroup`

schristley:
@scharch Yes, this PR will be for 1.6 while #773 is for 2.0 release. The 1.x stuff is maintained on a separate release branch.

scharch (approved):

PR #815, opened 2024-11-26
AKC updates for v2.0

### LINKED ISSUE #813: bugs in airr-schema.yaml / airr-schema-openapi3.yaml

schristley:
Consistency issues with the schema file as reported by the AKC project.

### DIFF
--- a/.github/workflows/py-unittest.yaml
+++ b/.github/workflows/py-unittest.yaml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.8' ]
+        python-version: [ '3.9' ]
     steps:
       - uses: actions/checkout@v2
 
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,9 @@ help:
 	@echo "make data-copy       -- Copy test data files to language directories"
 	@echo ""
 	@echo "make docker-latest   -- Build docker image with latest tag"
+	@echo "  Example: docker run -v \$$PWD:/work -it airrc/airr-standards bash"
 	@echo ""
+	@echo "  (environment setup required and/or run in docker)"
 	@echo "make checks          -- Run consistency checks on spec files"
 	@echo "make tests           -- Run all language test suites"
 	@echo "make python-tests    -- Run Python test suite"
@@ -58,4 +60,5 @@ r-tests:
 
 js-tests:
 	@echo "Running Javascript test suite"
-	cd lang/js; npm test
+	cd lang/js; rm -rf node_modules; rm -rf coverage
+	cd lang/js; npm install; npm run eslint; npm run test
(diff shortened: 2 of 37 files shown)


javh:
From the call:
* Good to merge after fixing tests failures.

bcorrie:
@williamdlees wanted to confirm that this change is correct... 

`AlleleDescription.release_version` is defined as an `integer`
`GermlineSet.release_version` is defined as a `number`

Is it correct for both of these to be `integer`?

bcorrie:
This pull request closed #813 which has a number of outstanding issues, so this needs more work...

williamdlees:
I'm OK with these changes.

javh:
From the call:
* @schristley will review with v1.6 changes and merge if it looks good.

schristley:
Hey @javh , looking at the R test failure after you merged in master, the log is mentioning airr-1.5.0. Do we have another version number somewhere that R is reading?

I think this is separate from the actual test error...

javh:
Oops. I think that's master being out of sync with the v1.5 maintenance branch, which is currently on v1.6. Let me fix that...

PR #821, opened 2025-07-12
Update schema version

schristley:
Programs that use the main branch to test V2 features may get confused with the version in the schema being 1.4, so update it to 2.0 for next version


schristley:
The warning indicates the module will be deprecated in a few months, so time to move forward.

```
/work/lang/python/airr/schema.py:10: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  from pkg_resources import resource_stream
```

Issue #870, opened 2026-04-09
Add v2.0 tagged spec in ADC API spec


bcorrie:
We need to link to the tagged AIRR spec in the ADC API file. For example:

https://github.com/airr-community/airr-standards/blob/cd92d6cb9345edefdcedc9feecee45b6db12b366/specs/adc-api.yaml#L79 

We also need to change the base path to /airr/v2:

https://github.com/airr-community/airr-standards/blob/cd92d6cb9345edefdcedc9feecee45b6db12b366/specs/adc-api.yaml#L17

PR #872, opened 2026-06-05
v2.0.0 Release Prep

javh:
Release prep, starting with release notes.


javh:
Added changes for #870.

scharch (approved):

javh:
I'm going to submit to the R package to CRAN later tonight if no one has any objections to additions to the release changes in this PR.

schristley:
@javh @bcorrie the ADC API change is incomplete. There's an openapi3 version of the spec too. I might have missed some discussion but was the decision to move to `v2` for `baseUrl`? I might be ok with that but I'm planning for CRWG to add some enhancements to V2 of the api so if we don't need to release an V2 of the api, maybe we can hold off?

javh:
@schristley and @bcorrie, I update the OpenAPIv3 spec too.  Let me know if you do/don't want to make the baseUrl change.

javh:
* R package is on CRAN.
* [Release is drafted](https://github.com/airr-community/airr-standards/releases/tag/untagged-48820c599019e6b10786)

I'll like to tag the release and push to PyPI in about 3 hours. Last chance for changes! :grin:

